PR #15, opened 2018-02-10
Fix fan control attributes to enums.

igorbernstein2:
According to ZigBee Cluster Library Specification Revision 6 Draft Version 1.0. The Fan Control cluster (0x0202) has 2 attributes both are enum8.
Currently zigpy treats all enums as unit8_t & unit16_t, which causes the following error when those attributes are written:
`[[<WriteAttributesStatusRecord status=Status.INVALID_DATA_TYPE attrid=0>]]`

This change has been verified to work on the following fan receiver:
https://www.homedepot.com/p/Hampton-Bay-Universal-Wink-Enabled-White-Ceiling-Fan-Premier-Remote-Control-99432/206591100.

This change is based on:
https://github.com/zigpy/bellows/pull/71

### DIFF
--- a/zigpy/types/basic.py
+++ b/zigpy/types/basic.py
@@ -82,11 +82,14 @@ class uint56_t(uint_t):  # noqa: N801
 class uint64_t(uint_t):  # noqa: N801
     _size = 8
 
+
 class enum8(uint8_t):  # noqa: N801
     pass
 
+
 class enum16(uint16_t):  # noqa: N801
     pass
+
 
 class Single(float):
     def serialize(self):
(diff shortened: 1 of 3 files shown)


rcloran:
It'd be really nice if we could have a Python enum inner class and list that a the type in the attributes (a bit like https://github.com/zigpy/zigpy/blob/1b1e6ee1704e319564a6b3fe4bdda5ddbe0edcde/zigpy/zcl/clusters/general.py#L387 ). I played around with that a bit, but I can't find a good way to deserialize it to the right thing with TypeValue. 

Thanks for the contribution!

PR #16, opened 2018-02-10
Bubble expect_reply parameter up to clusters (was:Stop expecting replies for commands.)

igorbernstein2:
Allow for commands to disable waiting for replies. This can be used to workaround devices that improperly implement the zigbee zha spec

Original:
According to ZigBee Cluster Library SpecificationRevision 6 Draft Version 1.0, Section 2.5.12.2 "When Generated", zigbee is not required to issue an explicit reply. If the initiating request is expecting a Default Response then any message sent with the same Transaction sequence number can act as a reply. The implementation to track what counts as a stand-in for a Default Response would be nontrivial.

This change fixes the behavior for the most common case of sending commands by setting expect_reply=False for all commands. This will most likely break other use cases where a command requires a response (ie. the Get Alarm request). In the longer term, the detection of the Default Response stand-in needs to be implemented or the commands in cluster definitions need to be annotated to specify if they require a reply.

This fix allows zigpy to control https://www.homedepot.com/p/Hampton-Bay-Universal-Wink-Enabled-White-Ceiling-Fan-Premier-Remote-Control-99432/206591100.

This fixes https://github.com/zigpy/bellows/issues/92

### DIFF
--- a/zigpy/device.py
+++ b/zigpy/device.py
@@ -77,8 +77,8 @@ class Device(zigpy.util.LocalLogMixin):
         self.endpoints[endpoint_id] = ep
         return ep
 
-    def request(self, profile, cluster, src_ep, dst_ep, sequence, data, **kwargs):
-        return self._application.request(self.nwk, profile, cluster, src_ep, dst_ep, sequence, data, **kwargs)
+    def request(self, profile, cluster, src_ep, dst_ep, sequence, data, expect_reply=True):
+        return self._application.request(self.nwk, profile, cluster, src_ep, dst_ep, sequence, data, expect_reply=expect_reply)
 
     def handle_message(self, is_reply, profile, cluster, src_ep, dst_ep, tsn, command_id, args):
         try:
--- a/zigpy/device.py
+++ b/zigpy/device.py
@@ -77,8 +77,8 @@ class Device(zigpy.util.LocalLogMixin):
         self.endpoints[endpoint_id] = ep
         return ep
 
-    def request(self, profile, cluster, src_ep, dst_ep, sequence, data):
-        return self._application.request(self.nwk, profile, cluster, src_ep, dst_ep, sequence, data)
+    def request(self, profile, cluster, src_ep, dst_ep, sequence, data, **kwargs):
+        return self._application.request(self.nwk, profile, cluster, src_ep, dst_ep, sequence, data, **kwargs)
 
     def handle_message(self, is_reply, profile, cluster, src_ep, dst_ep, tsn, command_id, args):
         try:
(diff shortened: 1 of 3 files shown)


igorbernstein2:
I updated the PR as requested, however it doesn't solve the problem with my device.

Doing a bit more digging, it seems like the device is misbehaving. I originally got confused by frame sequence and tsns. When I saw repeated frame sequences, I assumed that they were the commands that were being sent in response (see below for the log) and could stand as surrogates for the Default Response. I would very much like to get this receiver [working with home assistant](https://github.com/home-assistant/home-assistant/pull/12289) despite it breaking the spec. Since the default responses don't really matter, I was wondering how you'd feel about annotating all of the commands that expect a Default Response and loosening the spec for them? or maybe adding a parameter to the client to make this an opt-in mode that I can plumb through the zha plugin in home assistant?

Adding a bunch of logging the interaction looks like:
```
bellows -v DEBUG zcl 00:22:a3:00:00:16:c5:7d 1 6 command on
debug: Using selector: EpollSelector
....
zigby/device.py:Device#request: sequence=1
debug: Send command sendUnicast
bellows/ezsp.py:EZSP#_command(): self._seq=18
debug: Sending: b'2350219c54d3b5b658924a24ab1593499c4f26a8eccf66e1a77e'
debug: Data frame: b'3350a19c54b11e557e'
debug: Sending: b'8430fc7e'
debug: Application frame 52 (sendUnicast), seq 18 received
debug: Data frame: b'4350b1f1ad8a68774f944a8688556d859dcc653b0e7e'
debug: Sending: b'8520dd7e'
debug: Application frame 89 (incomingRouteRecordHandler), seq 18 received
debug: Data frame: b'5350b19754d3b5b658924a24ab1592499cd526abede50c7e'
debug: Sending: b'8610be7e'
debug: Application frame 63 (messageSentHandler), seq 18 received
  File "/home/pi/bellows-venv/bin/bellows", line 11, in <module>
...
  File "/usr/lib/python3.5/asyncio/tasks.py", line 404, in wait_for
    raise futures.TimeoutError()
```

igorbernstein2:
Sorry, for the confusion, the fan speed control uses attributes, so its not affected. The part that's broken is controlling the light on that fan.
I guess I can update light/zha to pass expect_reply=False. Does that sound alright?

rcloran:
You could ... I'm not sure it's the best course of action for most devices, though. Even if you create a custom device object (something that's not supported today, but I am going to add support for), you'll need this plumbing, which is why I'm OK with you adding it.

igorbernstein:
Ok please merge this. I’ll open separate PR in homeassistant to continue the discussion how the flag can be used

rcloran:
Thanks!

PR #32, opened 2018-04-16
Add Thermostat to zha.DeviceType enum

Adminiuga:
Add Thermostat to zha.DeviceType enum
Fix typo for "oc*c*upancy" attribute in zcl.clusters.hvac.Thermostat attributes.


zigpy/profiles/zha.py
```diff
@@ -71,4 +71,6 @@ class DeviceType(enum.IntEnum):
     DeviceType.ON_OFF_LIGHT_SWITCH: ([0x0007], [0x0004, 0x0005, 0x0006]),
     DeviceType.DIMMER_SWITCH: ([0x0007], [0x0004, 0x0005, 0x0006, 0x0008]),
     DeviceType.COLOR_DIMMER_SWITCH: ([0x0007], [0x0004, 0x0005, 0x0006, 0x0008, 0x0300]),
+    # HVAC
+    DeviceType.THERMOSTAT: ([0x0004, 0x0005, 0x0200, 0x0201, 0x0202, 0x0203, 0x0204], []),
```
rcloran: I'm not sure that this cluster list is correct for thermostat devices. It looks to me like it should be something like

```
DeviceType.THERMOSTAT: ([0x0004, 0x0005, 0x0201], [0x200, 0x202, 0x203]),
```

 (but I may be missing some in the *client* list there).

A `THERMOSTAT` device A `HEATING_COOLING_UNIT` looks more like a device that'd implement pump and fan *server* clusters.

Is this list from an observed real-world device?

Adminiuga: You could be right. At the moment I have only a non representative small sample of "Line Voltage" thermostats and for Sinope one I get the following list of clusters:
```Device:
  NWK: 0x2878
  IEEE: 50:0b:91:40:00:01:09:ae
  Endpoints:
    1: profile=0x104, device_type=DeviceType.THERMOSTAT
      Input Clusters:
        Basic (0)
        Identify (3)
        Groups (4)
        Scenes (5)
        Thermostat (513)
        Thermostat User Interface Configuration (516)
        Temperature Measurement (1026)
        Electrical Measurement (2820)
        Diagnostic (2821)
        Manufacturer Specific (65281)
      Output Clusters:
        Ota (25)
        Manufacturer Specific (65281)
    196: profile=0xc25d, device_type=769
      Input Clusters:
        Power Configuration (1)
Device:```

I could get a list of clusters for StelPro Zigbee thermostat soon. Trying to get one of the "low voltage" thermostats. I have a suspicion that low voltage consumer thermostats may have "fan control" cluster since they are "all in one" units technically. Following the same logic I'm expecting it to have the "Thermostat UI configuration" cluster, while technically it could be a different unit from thermostat.

Comment (Adminiuga) [reply zigpy/profiles/zha.py]: Here's `bellows devices` for StelPro STZB402+ thermostat:
```        Ota (25)
Device:
  NWK: 0xaa7f
  IEEE: f8:f0:05:ff:ff:f7:8d:d7
  Endpoints:
    25: profile=0x104, device_type=DeviceType.THERMOSTAT
      Input Clusters:
        Basic (0)
        Identify (3)
        Groups (4)
        Thermostat (513)
        Thermostat User Interface Configuration (516)
      Output Clusters:
        Temperature Measurement (1026)
Device:
```

Adminiuga: i've got a Zen thermostat for testing. here's `bellows devices` for it
```
Device:
  NWK: 0xac48
  IEEE: 00:24:46:00:00:0e:a1:a3
  Endpoints:
    1: profile=0x104, device_type=DeviceType.THERMOSTAT
      Input Clusters:
        Basic (0)
        Power Configuration (1)
        Identify (3)
        Groups (4)
        Scenes (5)
        Poll Control (32)
        Thermostat (513)
        Fan Control (514)
        Thermostat User Interface Configuration (516)
        Diagnostic (2821)
      Output Clusters:
        Time (10)
        Ota (25)
(venv) lex@debba:~$
```
As I expected the "Fan Control" is there, since it may have just the "fan mode"
so should it be more like:
```
DeviceType.THERMOSTAT: ([0x0004, 0x0005, 0x0201, 0x0202, 0x0204], []),
```

rcloran: It looks like none of these devices conform properly to the standard. It seems that the intent of the standard is that thermostat devices control other ZigBee devices (fans etc), where in real life, most thermostats control their fans etc through a wired interface, and not ZigBee. We should be describing the standard here.

Adminiuga: ah , now I understand. I guess I still was thinking from HASS perspective. But from the standards perspective your list makes most sense.

Adminiuga:
Per ZCL specs, Thermostat device may have use external Temperature, Humidity and Occupancy sensors. Does it mean we need to add those to the "client" clusters as well?

Issue #17, opened 2018-02-11
Ikea tradfri discovery throws exception


andrasbiro:
I'm trying to join an ikea tradfri bulb to home-assistant (v0.63.0), and I traced back the error to this exception:
```
bellows -d/dev/ttyUSB0 permit -D /config/zigbee.db
Joins are permitted for the next 30s...
Device 0x8bb4 (address) joined the network
[0x8bb4] Discovering endpoints
Device 0x796f (address) joined the network
[0x796f] Discovering endpoints
[0x796f] Discovered endpoints: [1]
[0x796f:1] Discovering endpoint information
[0x796f:1] Discovered endpoint information: <SimpleDescriptor endpoint=1 profile=49246 device_type=544 device_version=2 input_clusters=[0, 3, 4, 5, 6, 8, 768, 2821, 4096] output_clusters=[5, 25, 32, 4096]>
Failed ZDO request during device initialization: ('Endpoint request failed: %s', [129, 35764, []])
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/zigpy/device.py", line 54, in _initialize
    raise Exception("Endpoint request failed: %s", epr)
Exception: ('Endpoint request failed: %s', [129, 35764, []])
Done
```

The device itself:
```
bellows -d/dev/ttyUSB0 devices -D /config/zigbee.db
Device:
  NWK: 0x796f
  IEEE: (address removed)
  Endpoints:
    1: profile=0xc05e, device_type=DeviceType.COLOR_TEMPERATURE_LIGHT
      Input Clusters:
        Basic (0)
        Identify (3)
        Groups (4)
        Scenes (5)
        On/Off (6)
        Level control (8)
        Color Control (768)
        Diagnostic (2821)
        LightLink (4096)
      Output Clusters:
        Scenes (5)
        Ota (25)
        Poll Control (32)
        LightLink (4096)
```
The coordinator is a Telegesis EZSP dongle, and I don't yet have a tradfri hub, so the bulb might have old firmware.

rcloran:
The device joins twice (some devices seem to do that ...). The exception is for the first join, but the initialization succeeds for the device with its second NWK.

  Failed ZDO request during device initialization: ('Endpoint request failed: %s', [129, 35764, []])

35764 = 0x8bb4.

Can you provide hass debug logs? Also, since it's in the database (otherwise it wouldn't show in `bellows devices`, hass should try to set it up on start. What's not working?

[...]

PR #33, opened 2018-04-18
fix for zll devices use zll profile for non zll cluster

Yoda-x:
packets to zll devices should use the  zha profile id in frames if the cluster is not 0x1000 
See https://www.silabs.com/documents/public/user-guides/UG103.9.pdf section 3.1

This fixes issue #17 and also other bulbs with same behavior.


AndreasBomholtz:
Can you please also update the tests so that this new code is also tested?

Yoda-x:
@AndreasBomholtz I was afraid to hear to update the test, sigh. Ok, can you provide some guidance what to add? Should I add more test cases? Maybe: zll_profile+cluster 0x1000, zll_profile+cluster 0x0006, zha_profile+cluster 0x0100? and then check for the call counter? 
Many Thanks

Yoda-x:
Can you please provide some explanation what you want to see?  I can implement it, but I'm not really sure what you expect. 
Should I create a new function and move the test cases there ? 
Thanks

AndreasBomholtz:
Just add a new test function called something like test_request_profile_change or something...
And then move your test there.

Yoda-x:
I will redo this request. I messed something up here

PR #38, opened 2018-04-23
Allow attribute to be (optionally) passed to configure reporting as a string to avoid having to manually look up the mapping

SchaumburgM:
There is a lot of other dynamic method/class building to allow using the Cluster/attribute names instead of having to look up the integer index. This change checks the attribute parameter passed to configure_reporting() and if it is a string, it looks up the attribute index and then proceeds as usual.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,7 @@ language: python
 matrix:
   fast_finish: true
   include:
-    - python: "3.4"
-      env: TOXENV=py34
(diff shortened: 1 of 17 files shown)


AndreasBomholtz:
Any chance that you will add the missing test case?

dmulcahey:
#75 implemented this... This PR should be closed to avoid confusion.

PR #45, opened 2018-04-30
Handle devices with quirks

rcloran:


rcloran:
@damarco , I think I'm good to go with this. Thanks for your contributions -- if you could give it a look over, that'd be great. I ended up not adding the SmartThings device quirks (the ones I have all work, except for the battery stuff). I'd love to get support for battery quirks in once this is in.

PR #50, opened 2018-05-10
Move OnOffCluster from in_cluster to out_cluster

damarco:
OnOffCluster of Xiaomi Open/Close Sensor should be an output cluster.


rcloran:
Thanks!

PR #56, opened 2018-06-02
add manufacturer and model to endpoint

dmulcahey:
This adds manufacturer and model to the endpoint object. The values are initially read on device join while the device should be in a responsive state. They are restored from the attribute cache on restarts. This allows HA to avoid "asking" for them when devices are added because they are already set before the device is given to HA.

### DIFF
--- a/zigpy/appdb.py
+++ b/zigpy/appdb.py
@@ -216,6 +216,11 @@ class PersistingListener:
                 if cluster in ep.in_clusters:
                     clus = ep.in_clusters[cluster]
                     clus._attr_cache[attrid] = value
+                    LOGGER.debug("Attribute id: %s value: %s", attrid, value)
+                    if cluster == 0 and attrid == 4:
+                        ep.manufacturer = value.decode('ascii').strip()
+                    if cluster == 0 and attrid == 5:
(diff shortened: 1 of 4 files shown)


zigpy/appdb.py
```diff
@@ -216,6 +216,11 @@ def load(self):
                 if cluster in ep.in_clusters:
                     clus = ep.in_clusters[cluster]
                     clus._attr_cache[attrid] = value
+                    LOGGER.debug("Attribute id: %s value: %s", attrid, value)
+                    if cluster == 0 and attrid == 4:
```
AndreasBomholtz: Can you please use the cluster_id from the cluster class, so it is easier to understand.

zigpy/endpoint.py
```diff
@@ -64,6 +66,9 @@ def __init__(self, device, endpoint_id):
         for cluster in sd.output_clusters:
             self.add_output_cluster(cluster)
 
+        if 0 in self.in_clusters:
```
AndreasBomholtz: Also here....

dmulcahey:
@AndreasBomholtz done

PR #57, opened 2018-06-02
WIP - get attribute state when cluster is added

dmulcahey:
This will allow the correct state to be displayed when devices are added and avoids HA having to poll the cluster or issue a read on add to get the state of the cluster.


dmulcahey:
This isn’t going to do what I want it to... I’ll open another PR when I figure it out.

PR #58, opened 2018-06-03
add a quirk for king of fans zigbee receiver

igorbernstein2:
The King of Fans ceiling fan receiver does not generate default replies. This module overrides all server commands that do not have a mandatory reply to not expect replies at all.

### DIFF
--- a/zigpy/quirks/kof/__init__.py
+++ b/zigpy/quirks/kof/__init__.py
@@ -6,7 +6,7 @@ module overrides all server commands that do not have a mandatory reply to not
  expect replies at all.
 """
 from zigpy.quirks import CustomDevice, CustomCluster
-from zigpy.zcl.clusters.general import Basic, Identify, Groups, Scenes, OnOff, LevelControl
+from zigpy.zcl.clusters.general import Basic, Identify, Groups, Scenes, OnOff, LevelControl, Ota
 from zigpy.zcl.clusters.hvac import Fan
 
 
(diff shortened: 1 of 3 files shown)


zigpy/quirks/kof/__init__.py
```diff
@@ -0,0 +1,88 @@
...
+    A simple mixin that allows a cluster to have configureable list of command
+    ids that do not generate an explicit reply.
+    """
+    void_input_commands = []
+
+    def command(self, command, *args, manufacturer=None, expect_reply=None):
+        """
+        Overrides Cluster#command to configure expect_reply behavior based on
+        void_input_commands. Note that this method changes the default value of
+        expect_reply to None. This allows allows the caller to explicitly force
```
AndreasBomholtz: Double 'allows'

zigpy/quirks/kof/__init__.py
```diff
@@ -0,0 +1,88 @@
...
+        expect_reply to true.
+        """
+        if expect_reply is not None:
```
Adminiuga: can we do `if expect_reply is None` and move else block into then block?

cgarwood:
Any updates on this? I just set up one of these receivers and am getting a bunch of timeout errors in the log, I'm hopeful that this will take care of it.

Adminiuga:
running it in my environment, no complains so far.

igorbernstein2:
Finally found time to finish this. All feedback should be addressed

igorbernstein2:
Friendly ping

PR #59, opened 2018-06-04
support zha window covering devices

felixstorm:
This allows to communicate with the clusters of ZHA window covering devices (like the ubisys shutter control J1) and fixes a few minor typos / issues within these cluster specifications.


felixstorm:
sorry for the tests failing initially - don't have that much python experience yet...

PR #60, opened 2018-06-05
Cache output cluster attributes

dmulcahey:
This adds the ClusterPersistingListener to output clusters. It stores attribute values in the db and also restores  their values when the db is loaded. 

This will allow us to restore the last known state of devices like the Osram lightly dimmer.

### DIFF
--- a/tests/test_appdb.py
+++ b/tests/test_appdb.py
@@ -43,12 +43,13 @@ async def test_database(tmpdir):
     ep.profile_id = 260
     ep.device_type = 0xfffd  # Invalid
     clus = ep.add_input_cluster(0)
-    ep.add_output_cluster(1)
+    out_clus = ep.add_output_cluster(1)
     ep = dev.add_endpoint(3)
     ep.profile_id = 49246
     ep.device_type = profiles.zll.DeviceType.COLOR_LIGHT
     app.device_initialized(dev)
     clus._update_attribute(0, 99)
+    out_clus._update_attribute(0, 99)
     clus.listener_event('cluster_command', 0)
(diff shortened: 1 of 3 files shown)


damarco:
We should use a separate cache (db table) to store out_cluster attributes, because the same cluster_id can be part of the in_cluster and out_cluster list.

dmulcahey:
I don’t believe that is true for a single endpoint. I think the clusters have to be on separate endpoints for that to be the case...

damarco:
I do not think this is a requirement. The Zigbee HA spec lists some devices that have the same cluster as input and output cluster, eg. 'Price' cluster on a 'Meter Interface'.

Some other devices I encountered:

Tradfri Bulb:
`<SimpleDescriptor endpoint=1 profile=49246 device_type=512 device_version=2 input_clusters=[0, 3, 4, 5, 6, 8, 768, 2821, 4096] output_clusters=[5, 25, 32, 4096]>`

Tradfri Motion Sensor:
`<SimpleDescriptor endpoint=1 profile=260 device_type=2128 device_version=2 input_clusters=[0, 1, 3, 9, 2821, 4096] output_clusters=[3, 4, 6, 25, 4096]>`

dmulcahey:
Argh, ok, I’ll make the change. Thanks for following up on this!

dmulcahey:
@damarco done. Thanks for the feedback!

dmulcahey:
After discussing this more with @damarco I am unsure if this is necessary. I am closing this now and will reopen it if further investigation deems it helpful.

PR #65, opened 2018-06-27
make class zdo.Status a sub-class of uint8_t

Adminiuga:
so we could compare it to integers


AndreasBomholtz:
Looks good, thanks.

PR #66, opened 2018-06-28
cleanup zdo.leave() method and remove unneeded extra parameter from cmd request

Adminiuga:
Per definition of [Mgmt_leave_req](https://github.com/zigpy/zigpy/blob/a452c1c61778b116447ab36d1bf7e08b8820ebbf/zigpy/zdo/types.py#L124) of ZDO clusters, Mgmt_leave_req only takes two arguments:
1. device address
2. options
however current implementation also provides an extra parameter which seems to be just a copy'n'paste artifact from the above bind/unbind methods.

Same arguments for Mgmt_leave_req were defined in ZDO specs
image


AndreasBomholtz:
Thanks.

PR #68, opened 2018-07-01
Add aqara body sensor

dmulcahey:
This updates the aqara open close sensor. It was being made to work like a switch instead of a contact sensor. This causes toggles to be created in HA which are not usable. It now identifies as an on off output sensor. I have a corresponding change for HA that will make this a binary sensor instead of a switch. This makes it consistent with how other contact sensors work. 

This also adds support for aqara body sensors which support illuminance and occupancy

### DIFF
--- a/zigpy/quirks/xiaomi/__init__.py
+++ b/zigpy/quirks/xiaomi/__init__.py
@@ -72,7 +72,6 @@ class AqaraOpenCloseSensor(CustomDevice):
     replacement = {
         'endpoints': {
             1: {
-                'device_type': 0x0002,
                 'input_clusters': [0x0000, 0x0003],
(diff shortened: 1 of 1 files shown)


zigpy/quirks/xiaomi/__init__.py
```diff
@@ -1,4 +1,7 @@
-from zigpy.quirks import CustomDevice
+import asyncio
```
AndreasBomholtz: We are not using asyncio anymore.

dmulcahey: What should I use instead?

dmulcahey:
@damarco makes sense. Removed the mapping altogether.

PR #69, opened 2018-07-01
Add 'only_cache' option to read_attributes()

damarco:
Adds 'only_cache' option to read_attributes(). This can be used to restore initial state from cached attributes.


Adminiuga:
@AndreasBomholtz can we get this merged, please? This would help minimizing "Setup of platform zha is taking over 10 seconds" in home assistant.

PR #74, opened 2018-08-15
Create _idx Cluster class attributes even if skipping the registry

Adminiuga:
zigpy.quirks.CustomCluster class does not create _attridx, _server_command_idx, _client_command_idx class attributes and without those reading/writting attributes by their name fails. For example, with the following custom cluster:
```
class SinopeTechnologiesManufacturerCluster(CustomCluster):
    cluster_id = SINOPE_MANUFACTURER_CLUSTER_ID
    name = "Sinopé Technologies Manufacturer specific"
    ep_attribute = 'sinope_manufacturer_specific'
    attributes = {
        0x0010: ('outdoor_temp', t.int16s),
        0x0020: ('secs_since_2k', t.uint32_t),
    }
    client_commands = {}
    server_commands = {}
```
writing attributes using actual attribute names fails with:
```
  File "/usr/lib/python3.5/asyncio/tasks.py", line 239, in _step
    result = coro.send(None)
  File "/home/lex/lex/src/ac-hass/homeassistant/helpers/entity_platform.py", line 324, in _async_add_entity
    await entity.async_added_to_hass()
  File "/home/lex/lex/src/ac-hass/homeassistant/components/climate/zha.py", line 640, in async_added_to_hass
    await self._async_update_time()
  File "/home/lex/lex/src/ac-hass/homeassistant/components/climate/zha.py", line 691, in _async_update_time
    write_attributes(data, manufacturer=self.manufacturer)
  File "/home/lex/lex/src/zigpy/zigpy/zcl/__init__.py", line 232, in write_attributes
    attrid = self._attridx[attrid]
  File "/home/lex/lex/src/zigpy/zigpy/zcl/__init__.py", line 325, in __getattr__
    raise AttributeError("No such command name: %s" % (name, ))
AttributeError: No such command name: _attridx
```

I think we still need to create _idx Cluster class attributes even if the CustomCluster is not being added to the registry so we could address attributes & commands by their name, to keep things more readable


zigpy/zcl/__init__.py
```diff
@@ -35,6 +29,12 @@ def __init__(cls, name, bases, nmspc):  # noqa: N805
                 command_name, schema, is_reply = details
                 cls._client_command_idx[command_name] = command_id
 
+        if not getattr(cls, '_skip_registry', False):
```
AndreasBomholtz: Can you keep the structure as it was before, but just move it down here?

Adminiuga: Done!

PR #82, opened 2018-09-18
Support manufacturer id for configure_reporting().

Adminiuga:
Support `manufacturer` for `Cluster.configure_reporting()` method.

I've got [SmartThings multipurpose sensor](https://www.samsung.com/us/smart-home/smartthings/sensors/samsung-smartthings-multipurpose-sensor-2018-gp-u999sjvlaaa/), mine being specifically PN IM6001-MPP01 and it reject `configure_reporting()` request for the acceleration manufacturer specific cluster, unless configure_reporting() request uses `manufacturer_id` specific to this vendor.

### DIFF
--- a/tests/test_zcl.py
+++ b/tests/test_zcl.py
@@ -282,6 +282,24 @@ def test_configure_reporting_wrong_attrid(cluster):
     assert cluster._endpoint.request.call_count == 0
 
 
+def test_configure_reporting_manuf():
+    ep = mock.MagicMock()
+    cluster = zcl.Cluster.from_id(ep, 6)
+    cluster.request = mock.MagicMock(name='request')
+    cluster.configure_reporting(0, 10, 20, 1)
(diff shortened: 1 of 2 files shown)


Adminiuga:
would it make sense to add a default `Cluster.manufacturer_id = None` attribute and use it in `Cluster.request()` in case default argument `manufacturer=None` and `cluster_id >= 0xfc00` , like in
```
@util.retryable_request
def request(self, general, command_id, schema, *args, manufacturer=None, expect_reply=True):
    if manufacturer is None and self.cluster_id >= 0xfc00:
        manufacturer = self.manufacturer_id
...
```
Then application or quirks just need to set `manufacturer_id` on the cluster and all request to that cluster will use the right `manufacturer`  code while still be able to override manufacturer for specific requests?

rcloran:
Thanks!

PR #83, opened 2018-09-18
Add Smartthings multipurpose sensor quirks.

Adminiuga:
Implement vendor specific `acceleremoter` cluster for [Smartthings multipurpose sensor](https://www.samsung.com/us/smart-home/smartthings/sensors/samsung-smartthings-multipurpose-sensor-2018-gp-u999sjvlaaa/) Per [Smartthings device handler](https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/PROD_2018.09.11/devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy)  there are multiple vendors for the similar sensor, but I was able only to test with sensor from 'samjin' manufacturer.

Configuring reporting for `0xfc02` cluster requires specific manufacturer_id be supplied with all ZCL requests and this functionality is in #82


Adminiuga:
the last coveralls build was unsuccessful. Forced push rebased to latest master to force a rebuild...

PR #91, opened 2018-11-02
Broadcast/unicast 'mgmt_permit_joining_req' implementation

Adminiuga:
This is a blend of the original [bellows zdo_broadacs()](https://github.com/zigpy/bellows/pull/109) implementation and #88


zigpy/device.py
```diff
@@ -131,3 +132,13 @@ def ieee(self):
 
     def __getitem__(self, key):
         return self.endpoints[key]
+
+
+async def broadcast(app, profile, cluster, src_ep, dst_ep, grpid, radius,
```
rcloran: I'm not sure I see the reason for this here. Couldn't things just call `app.broadcast()` directly?

Adminiuga: @rcloran let me know if it is better to call those directly and I'll update the PR. Once it goes through, we could add ZHA config panel and targeted joins to home assistant.

PR #130, opened 2019-02-21
remove ascii restriction

zalke:


zigpy/endpoint.py
```diff
@@ -133,7 +133,7 @@ def add_output_cluster(self, cluster_id, cluster=None):
             if isinstance(value, bytes):
                 try:
                     value = value.split(b'\x00')[0]
-                    attributes[key] = value.decode('ascii').strip()
+                    attributes[key] = value.decode().strip()
                 except UnicodeDecodeError:
                     # Unsure what the best behaviour here is. Unset the key?
                     pass
```
Adminiuga: in case of an UnicodeDecodeError exception, should we set the key to "hexlified" bytes?

dmulcahey: Yes, that’s done in several spots in bellows already. Makes sense to do it here as well.

Adminiuga:
Should be fixed via #147

PR #133, opened 2019-03-23
Add ControllerApplication.shutdown() method.

Adminiuga:
Transition to ControllerApplication shutdown() method. At very least should close the serial port. May perform other "clean up" task depending on radio implementation.


Adminiuga:
@damarco I have a need to add a shutdown() method into bellows to properly clean-up watchdog/reset tasks. I think it is a good idea to have it at zigpy level and require other radios to implement this as well. 
My plan is to replace `radio.close()` in HA with `ApplicationController.shutdown()`

PR #140, opened 2019-04-15
Add "node descriptor" to zigpy.device.Device class.

Adminiuga:
This PR adds ZDO Node Descriptor to Zigpy Device and I'm opening this for further discussion and feedback. 
This is in effort to implement `setExtendedTimeout()` in `bellows` to:

> Tell the stack whether or not the normal interval between retransmissions of a retried unicast message should be increased
by EMBER_INDIRECT_TRANSMISSION_TIMEOUT. The interval needs to be increased when sending to a sleepy node so that the message is not retransmitted until the destination has had time to wake up and poll its parent. The stack will automatically extend the timeout: - For our own sleepy children. - When an address response is received from a parent on behalf of its child. - When an indirect transaction expiry route error is received. - When an end device announcement is received from a sleepy node"

To know whether the destination is an end device or router we pull a mandatory ZDO `Node Descriptor` and store it in the database. Normally we're pulling Node Descriptor as part of device discovery, however for the purpose of migration from older versions, we also pulling it whenever we get traffic from a device which doesn't have a node descriptor set.

### DIFF
--- a/tests/test_appdb.py
+++ b/tests/test_appdb.py
@@ -7,7 +7,7 @@ import zigpy.types as t
 from zigpy.application import ControllerApplication
 from zigpy import profiles
 from zigpy.quirks import CustomDevice
-from zigpy.device import Status
+from zigpy.device import Device, Status
 from zigpy.zdo import types as zdo_t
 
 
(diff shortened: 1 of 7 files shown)


Adminiuga:
@damarco @dmulcahey 
This one is ready for review/merging. I don't want to rush this one into HA 0.92, but would like it pushed to PyPi so we could test it with HA 0.93.dev tree

PR #143, opened 2019-04-23
WIP Implement ZDO commands as enums.

Adminiuga:
Implement ZDO commands as enums.
Allow ZDO command invoking by their name.

Improves code readability and gets rid of "magic" numbers in the code.


Adminiuga:
@damarco @dmulcahey this is to open a discussion. I'll rebase it after/if #140 goes through.

PR #144, opened 2019-04-26
Updated Keen Temp & Humidity sensor's quirk

dbajric:
Now that I had some more time to play with the sensor, I updated the replacement clusters to better match the actual behavior of the sensor.


dbajric:
@dmulcahey - This sensor is advertised and sold as a dual sensor (temperature and humidity), but the data that's emitted from it includes pressure measurement as well. I put it in the replacement cluster in order to prevent Home Assistant's logs to be polluted with warning messages. 

Now, here is something that you may know how to fix. When I put the PressureMeasurement to the input cluster, Home Assistant will recognize it properly, but the value never gets updated from Unknown. If I manually send ask for "measured_value" attribute, I get a value and it is in expected range. So, couple questions for you:

1) Why would the value never change unless I manually ask for the attribute? I have waited for couple of days in hope that it will get updated, but no luck.
2) Are there any tricks that I could do to force the refresh?

Adminiuga:
@dbajric 
> When I put the PressureMeasurement to the input cluster, Home Assistant will recognize it properly, but the value never gets updated from Unknown
> Why would the value never change unless I manually ask for the attribute? I have waited for couple of days in hope that it will get updated, but no luck.

Attribute reporting needs to be configured on that cluster, for the device to send the updates on that cluster.  I'm confused as to why Keen put those clusters as "client clusters" as it doesn't make any sense, unless the device is suppose to receive the measurements, not to send?

With pressure cluster id in the "input clusters" list, remove the device from home assistant, repair it and collect the debug log. Look when HA sends "configure report" command to the device and check what the device replies with. 

PS: can you post a link to the device?

dbajric:
@Adminiuga, I managed to figure out with a sniffer what's going on. Basically, the attribute reporting was happening, but the attribute ID of "measured value" was different from the spec. They put the value on 0x0020 instead of 0x0000. After redirecting the value to appropriate ID, Home Assistant started reporting it correctly.

The sensor is https://keenhome.io/products/temp-sensor

Adminiuga:
Keen cannot be serious charging $40 for a $12 xiaomi sensor, can they?

dbajric:
The value on attr id 0x0020 is usually around 998450.

I created this PR for completeness purposes, but I am totally fine to make the changes in a different lib. You are referring to https://github.com/dmulcahey/zha-device-handlers?

Adminiuga:
yep, https://github.com/dmulcahey/zha-device-handlers

dbajric:
Alright. I have created a PR on that other repo (https://github.com/dmulcahey/zha-device-handlers/pull/88).

I guess it would then make sense to just remove the `keen` folder from here once the PR above is merged and new version is released. I'll open a new PR when ready. For now, closing this one.

PR #150, opened 2019-05-04
Add a `context` parameter to `ListenableMixin.add_listener`

puddly:
This is useful when creating a single listener object that subscribes to events from multiple sources. My use case is a global listener for the OTA cluster for every device that automatically downloads and caches firmware files. It's a little unwieldy to create one-off classes to just pass events along with some context:

```Python

class OTAListener:
    def __init__(self, cluster, global_listener):
        self.cluster = cluster
        self.global_listener = global_listener

    def cluster_command(self, tsn, command_id, args):
        self.global_listener(tsn, command_id, args, self.cluster)

global_listener = MainOTAListener()

for device in devices:
    ota_cluster = device.endpoints[0x001].out_clusters[0x0019]
    ota_cluster.add_listener(OTAListener(ota_cluster, global_listener))
```

Instead, one could do something like this:

```Python

class OTAListener:
    def cluster_command(self, tsn, command_id, args, ota_cluster):
        ...

ota_listener = OTAListener()

for device in devices:
    device.endpoints[0x001].out_clusters[0x0019].add_listener(ota_listener, context=True)
```

It would also be nice to have a catch-all listener (that I currently implement with `__getattr__`) but I will make a second pull request for that.

### DIFF
--- a/tests/test_zigbee_util.py
+++ b/tests/test_zigbee_util.py
@@ -12,20 +12,33 @@ class Listenable(util.ListenableMixin):
 
 def test_listenable():
     listen = Listenable()
-    listener = mock.MagicMock(spec_set=['event'])
-    listen.add_listener(listener)
-    listen.add_listener(listener)
 
+    # Python 3.7 guarantees dict ordering so this will be called first to test error handling
     broken_listener = mock.MagicMock()
     broken_listener.event.side_effect = Exception()
     listen.add_listener(broken_listener)
 
-    listen.listener_event('event')
+    listener = mock.MagicMock(spec_set=['event'])
+    listen.add_listener(listener)
+    listen.add_listener(listener)
+
+    context_listener = mock.MagicMock(spec_set=['event'])
+    listen.add_listener(context_listener, context=True)
+
+    listen.listener_event('event', 'test1')
+    listener.event.assert_has_calls([mock.call('test1'), mock.call('test1')], any_order=True)
+    context_listener.event.assert_has_calls([mock.call('test1', listen)], any_order=True)
+    broken_listener.event.assert_has_calls([mock.call('test1')], any_order=True)
     assert listener.event.call_count == 2
+    assert context_listener.event.call_count == 1
     assert broken_listener.event.call_count == 1
 
-    listen.listener_event('non_existing_event')
+    listen.listener_event('non_existing_event', 'test2')
+    listener.event.assert_has_calls([mock.call('test1'), mock.call('test1')], any_order=True)
+    context_listener.event.assert_has_calls([mock.call('test1', listen)], any_order=True)
+    broken_listener.event.assert_has_calls([mock.call('test1')], any_order=True)
     assert listener.event.call_count == 2
+    assert context_listener.event.call_count == 1
     assert broken_listener.event.call_count == 1
 
 
--- a/tests/test_zigbee_util.py
+++ b/tests/test_zigbee_util.py
@@ -23,11 +23,11 @@ def test_listenable():
     listen.add_listener(listener)
 
     context_listener = mock.MagicMock(spec_set=['event'])
-    listen.add_listener(context_listener, context=True)
+    listen.add_context_listener(context_listener)
 
     listen.listener_event('event', 'test1')
     listener.event.assert_has_calls([mock.call('test1'), mock.call('test1')], any_order=True)
-    context_listener.event.assert_has_calls([mock.call('test1', listen)], any_order=True)
+    context_listener.event.assert_has_calls([mock.call(listen, 'test1')], any_order=True)
     broken_listener.event.assert_has_calls([mock.call('test1')], any_order=True)
     assert listener.event.call_count == 2
     assert context_listener.event.call_count == 1
@@ -35,7 +35,7 @@ def test_listenable():
 
     listen.listener_event('non_existing_event', 'test2')
     listener.event.assert_has_calls([mock.call('test1'), mock.call('test1')], any_order=True)
-    context_listener.event.assert_has_calls([mock.call('test1', listen)], any_order=True)
+    context_listener.event.assert_has_calls([mock.call(listen, 'test1')], any_order=True)
     broken_listener.event.assert_has_calls([mock.call('test1')], any_order=True)
     assert listener.event.call_count == 2
     assert context_listener.event.call_count == 1
--- a/zigpy/util.py
+++ b/zigpy/util.py
@@ -11,18 +11,24 @@ LOGGER = logging.getLogger(__name__)
 
 
 class ListenableMixin:
-    def add_listener(self, listener):
+    def add_listener(self, listener, context=False):
         id_ = id(listener)
         while id_ in self._listeners:
             id_ += 1
-        self._listeners[id_] = listener
+        self._listeners[id_] = (listener, context)
         return id_
 
     def listener_event(self, method_name, *args):
-        for listener in self._listeners.values():
+        for listener, context in self._listeners.values():
+            method = getattr(listener, method_name, None)
+
+            if not method:
+                continue
+
             try:
-                method = getattr(listener, method_name, None)
-                if method:
+                if context:
+                    method(*args, self)
+                else:
                     method(*args)
             except Exception as e:
                 LOGGER.warning("Error calling listener.%s: %s", method_name, e)
--- a/zigpy/util.py
+++ b/zigpy/util.py
@@ -11,23 +11,29 @@ LOGGER = logging.getLogger(__name__)
 
 
 class ListenableMixin:
-    def add_listener(self, listener, context=False):
+    def _add_listener(self, listener, include_context):
         id_ = id(listener)
         while id_ in self._listeners:
             id_ += 1
-        self._listeners[id_] = (listener, context)
+        self._listeners[id_] = (listener, include_context)
         return id_
 
+    def add_listener(self, listener):
+        return self._add_listener(listener, include_context=False)
+
+    def add_context_listener(self, listener):
+        return self._add_listener(listener, include_context=True)
+
     def listener_event(self, method_name, *args):
-        for listener, context in self._listeners.values():
+        for listener, include_context in self._listeners.values():
             method = getattr(listener, method_name, None)
 
             if not method:
                 continue
 
             try:
-                if context:
-                    method(*args, self)
+                if include_context:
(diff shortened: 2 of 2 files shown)


dmulcahey:
@puddly do you actually have OTA working???

puddly:
@dmulcahey I implemented only enough to get my dimmable Trådfri bulbs updated but it does work (and takes 10 minutes per bulb because I can send only 50 bytes per messages). I'll post my listener when I finish cleaning it up.

puddly:
@dmulcahey I don't have any more bulbs to upgrade to actually test this code but it handles messages on the OTA cluster and should upgrade the firmware of any Trådfri device:

```Python
import asyncio
import logging
import contextlib

from collections import defaultdict

import aiohttp
from zigpy.zcl.foundation import Status
from zigpy.zcl.clusters.general import Ota

logging.basicConfig(level=logging.DEBUG)
logger = logging.getLogger(__name__)



class HerdLock:
    '''
    A lock that indicates to the caller if they are the first to acquire it.
    If not, the context manager waits until the lock is released. Used like this:

        lock = HerdLock()
        expensive = None

        async def worker():
            async with lock('LOCK_NAME') as is_first:
                if is_first:
                    print('Actually performing task')
                    await asyncio.sleep(5)
                    expensive = 'got it'

            # Here we can assume `expensive` is set and that only one worker actually fetched it
            print(expensive)

        await asyncio.gather(worker(), worker(), worker(), worker())
    '''

    def __init__(self):
        self.semaphores = defaultdict(asyncio.Semaphore)

    @contextlib.asynccontextmanager
    async def __call__(self, key):
        semaphore = self.semaphores[key]

        try:
            is_first = not semaphore.locked()

            async with semaphore:
                yield is_first
        finally:
            # Delete the lock if we're the last one to release it
            if not semaphore.locked():
                self.semaphores.pop(key, None)


class TrådfriOTAMainListener:
    UPDATE_URL = 'https://fw.ota.homesmart.ikea.net/feed/version_info.json'
    OTA_HEADER = 0x0BEEF11E.to_bytes(4, 'little')
    MAXIMUM_DATA_SIZE = 40

    def __init__(self):
        self._firmware_cache = {}
        self._lock = HerdLock()

    def cluster_command(self, tsn, command_id, args, ota_cluster):
        asyncio.create_task(self._cluster_command(tsn, command_id, args, ota_cluster))

    async def _cluster_command(self, tsn, command_id, args, ota_cluster):
        if not self._firmware_cache:
            logger.info('Downloading firmware update list from IKEA')

            async with self._lock('DOWNLOAD_VERSION_INFO') as was_first:
                if was_first:
                    async with aiohttp.ClientSession() as client:
                        async with client.get(self.UPDATE_URL) as response:
                            firmwares = await response.json(content_type='application/octet-stream')

                        self._firmware_cache.clear()

                        for fw in firmwares:
                            if 'fw_file_version_MSB' not in fw:
                                continue

                            fw['fw_file_version'] = (fw['fw_file_version_MSB'] << 16) | fw['fw_file_version_LSB']
                            self._firmware_cache[(fw['fw_manufacturer_id'], fw['fw_image_type'])] = fw

        if command_id == 0x0001:  # query_next_image
            field_control, manufacturer_code, image_type, current_file_version, hardware_version = args
            key = (manufacturer_code, image_type)

            logger.info('Received an OTA image query from %s', ota_cluster.endpoint.device)

            # We don't know what this device is
            if key not in self._firmware_cache:
                await ota_cluster.query_next_image_response(Status.NO_IMAGE_AVAILABLE, 0x0000, 0x0000, 0x00000000, 0x00000000)
                return

            fw = self._firmware_cache[key]

            # Tell the device we're ready to continue
            await ota_cluster.query_next_image_response(Status.SUCCESS, manufacturer_code, image_type, fw['fw_file_version'], fw['fw_filesize'])
        elif command_id == 0x0003:  # image_block
            field_control, manufacturer_code, image_type, \
            file_version, file_offset, maximum_data_size, \
            request_node_address, block_request_delay = args

            # We assume at this point we won't be getting unsolicited requests for blocks
            key = (manufacturer_code, image_type)
            fw = self._firmware_cache[key]

            # Download the firmware on demand
            if 'fw_data' not in fw:
                async with self._lock(f'DOWNLOAD_FW:{fw["fw_binary_url"]}') as was_first:
                    if was_first:
                        async with aiohttp.ClientSession() as client:
                            async with client.get(fw['fw_binary_url']) as response:
                                data = await response.read()

                        # The IKEA images wrap the Zigbee OTA file in a container
                        offset = data.index(self.OTA_HEADER)
                        fw['fw_data'] = data[offset:offset + fw['fw_filesize']]
                        assert len(fw['fw_data']) == fw['fw_filesize']

            logger.debug('Firmware upgrade progress: %0.2d', 100.0 * file_offset / fw['fw_filesize'])
            data = fw['fw_data'][file_offset:file_offset + min(self.MAXIMUM_DATA_SIZE, maximum_data_size)]

            await ota_cluster.image_block_response(Status.SUCCESS, manufacturer_code, image_type, file_version, file_offset, data)
        elif command_id == 0x0006:  # upgrade_end
            status, manufacturer_code, image_type, file_version = args

            # Upgrade right now
            await ota_cluster.upgrade_end_response(manufacturer_code, image_type, file_version, 0x00000000, 0x00000000)


class TrådfriOTAListener:
    def __init__(self, ota_cluster, main_listener):
        self.ota_cluster = ota_cluster
        self.main_listener = main_listener

    def cluster_command(self, tsn, command_id, args):
        logger.info('Received an OTA cluster command from %s', self.ota_cluster.endpoint.device)

        self.main_listener.cluster_command(tsn, command_id, args, self.ota_cluster)
```

You'd use it like this:

```Python

main_listener = TrådfriOTAMainListener()

ota_cluster = device.endpoints[0x0001].out_clusters[Ota.cluster_id]
ota_cluster.add_listener(TrådfriOTAListener(ota_cluster, main_listener))

# Tell the device we have a new firmware image for it
await ota_cluster.image_notify(0x00, 100, 0, 0, 0)
```

Adminiuga:
OTA updates are interesting and I like the idea of "context" listeners. 
What do you think if we leave the `add_listener()` as is and instead add explicit `add_context_listener()` or something similar?

dmulcahey:
@Adminiuga I prefer the explicit add_context_listener myself

Adminiuga:
And actually make the "context" the 1st parameter for the context listeners?
```
                if context:
                    method(*args, self)
                else:
                    method(*args)
```
?

puddly:
A new method is probably better, I'll change it.

Also, what is the purpose of the `_listeners` dictionary? It seems like a list would work just as well.

Adminiuga:
> what is the purpose of the _listeners dictionary? It seems like a list would work just as well.

Looks like list would work as well. With list we probably don't need check for listeners to have unique IDs

Adminiuga:
Do you mind rebasing it against dev branch and changing the base to dev for this PR?
Other than that it looks good.

puddly:
@Adminiuga Let me know if that works.

Adminiuga:
@dmulcahey we should get this into `zha-device-handlers`

dmulcahey:
@Adminiuga what specifically?

Adminiuga:
ugh, forgot to attach the link https://github.com/zigpy/zigpy/pull/150#issuecomment-489353657
The OTA upgrade for Tradfries. 
Although I'm looking for ideas to move generic code into Zigpy, like clustrer command handler and only have firmware downloader be in quirks. Thus we could instantiate firmware repositories from quirks for each manufacturer. Assuming others will publish their firmware

puddly:
I think quirks will need more than just providers for the OTA files. I believe I followed the spec well enough (see pages 27+ of [the spec](http://www.zigbee.org/wp-content/uploads/2014/11/docs-09-5264-23-00zi-zigbee-ota-upgrade-cluster-specification.pdf)) and while my RGB bulbs check for new firmware a minute after booting and a reply is sent by the listener with a higher version number, they never actually begin requesting data.

A generic solution for manufacturers who hide their OTA updates (i.e. everybody but IKEA) would be to clone an existing device, pair the clone to the manufacturer's hub, and have the hub "upgrade" the clone that saves the upgrade info to a computer.

This is probably the wrong place for this discussion but I'm interested in upgrading the last of my Trådfri bulbs and starting on my Xiaomi devices. If we can get together enough hubs and the right hardware, it shouldn't be hard to add firmware upgrades to bellows or zigpy and finally get rid of these stupid hubs.

Adminiuga:
@puddly  is there a reason you are capping `max_transfer_size` to 40 bytes? I'm seeing Ikea remote requesting transfer size of 63 bytes?

```
 data = fw['fw_data'][file_offset:file_offset + min(self.MAXIMUM_DATA_SIZE, maximum_data_size)]
```

```
OTA image_block handler for 'IKEA of Sweden TRADFRI remote control': field_control=0, manufacturer_code=4476, image_type=4545, file_version=304231795, file_offset=43800, max_data_size=63 ...
```

puddly:
I tried responding with IKEA's preferred size but my adapter refused to send a packet that big (it probably has a 64 byte limit). There's probably a way to increase it.

Adminiuga:
what adapter are you using?
I know there's a limit, but I thought it was something like 96bytes if encryption is used

puddly:
I'm using a [Nortek HUSBZB-1](https://www.nortekcontrol.com/products/2gig/husbzb-1-gocontrol-quickstick-combo/). Does sending 63 bytes work for you?

Adminiuga:
Apparently not. Trying with Ikea remote. Per node descriptor, `maximum_incoming_transfer_size`, `maximum_outgoing_transfer_size` and `maximum_buffer_size` is 82 bytes. 
ZCL portion of `image_block_response` with transfer size of 63 bytes  is totaling to 80 bytes. But that is ZCL layes and node descriptors limits the ASDU sublayer and I don't know how much overhead there's in between.

Adminiuga:
Although, per [Zigbee frame overview](https://www.microchip.com/forums/download.axd?file=0;193947) ASDU is what we're sending in ZCL so it should work

puddly:
@Adminiuga Which repository would be the right place to open an issue regarding OTA development? I have information about how to get access to the Xiaomi (Lumi) Gateway to intercept OTA updates, though I've now discovered that Xiaomi only has them to devices that aren't battery powered.

Adminiuga:
Open an issue at Zigpy something like "RFC OTA Support"

PR #155, opened 2019-05-11
Prevent slicing and int.from_bytes from allowing invalid deserialization

puddly:
Using the pattern `int.from_bytes(s[:n], 'little')` doesn't enforce that `s` is at least `n` bytes long (or that `s` is even non-empty). I've added checks to prevent this from happening for I think all of the affected data types.

It might be better to create some helper functions in the future to deal with this (or switch to using `BytesIO` objects).

### DIFF
--- a/zigpy/types/basic.py
+++ b/zigpy/types/basic.py
@@ -260,7 +260,7 @@ class CharacterString(str):
             raise ValueError('Data is too short')
 
         raw = data[cls._prefix_length:cls._prefix_length + length]
-        raw = bytes.split(b'\x00')[0]
+        raw = raw.split(b'\x00')[0]
 
         r = cls(raw.decode('utf8', errors='replace'))
         r.raw = raw
--- a/zigpy/types/basic.py
+++ b/zigpy/types/basic.py
@@ -9,6 +9,9 @@ class int_t(int):  # noqa: N801
 
     @classmethod
     def deserialize(cls, data):
+        if len(data) < cls._size:
+            raise ValueError('Data is too short to contain %d bytes' % cls._size)
+
         r = cls.from_bytes(data[:cls._size], 'little', signed=cls._signed)
         return r, data[cls._size:]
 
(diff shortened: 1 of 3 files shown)


zigpy/types/basic.py
```diff
@@ -150,9 +153,17 @@ def serialize(self):
 
     @classmethod
     def deserialize(cls, data):
-        bytes = int.from_bytes(data[:cls._prefix_length], 'little')
-        s = data[cls._prefix_length:bytes + 1]
-        return cls(s), data[bytes + 1:]
+        if len(data) < cls._prefix_length:
```
Adminiuga: `if len(data) < cls._prefix_length:`
Nah, this is wrong. `cls._prefix_length` specifies the size in bytes (uint8_t or uint16_t) for the field containing the "length" of data. eg the prefix length of 1 == 1 byte means the length of data should be 0<= len(data) <= 254 (255 is considered invalid length)

Also I could have zero data eg `b''` so `(len(data < cls._prefix_length) is True` while correctly it should deserialize into `b\x00`

Adminiuga: ugh, it is too late. I confused deserialize vs serialize

Adminiuga: I see what you did here. Nice. 
Do you actually see devices sending the wrong data? ie not enough data to deserialize?

puddly: No, not yet. I was converting some OTA code to use the zigpy types module and noticed it might see the same hard-to-debug issue I ran into with similar code using slicing and `int.from_bytes` for parsing.

Adminiuga:
can you rebase against current dev and resolve conflicts?

puddly:
Done.

zigpy/types/basic.py
```diff
@@ -236,11 +251,20 @@ def serialize(self):
...
+
         length = int.from_bytes(data[:cls._prefix_length], 'little')
-        raw = data[cls._prefix_length:length + 1]
-        r = cls(raw.split(b'\x00')[0].decode('utf8', errors='replace'))
+
+        if len(data) < cls._prefix_length + length:
+            raise ValueError('Data is too short')
+
+        raw = data[cls._prefix_length:cls._prefix_length + length]
+        raw = raw.split(b'\x00')[0]
```
Adminiuga: Don't split raw by null byte. Need the entire thing as raw, however input into `decode('utf8')` should be split

Adminiuga: Only because xiaomi are sending bytes in `CharacterString` ZCL type.

PR #156, opened 2019-05-14
RFC Zigpy group management framework.

Adminiuga:
**Group Management Framework**
* ApplicationController contains `groups` property (`Groups` class) which is a dict of `Group`'s instances
* each `Group` instance is a dict containing member devices (keyed on `IEEE` ) and have the following properties
  * `group_id` -- group ID
  * `name` -- group name
 
A Zigpy Device can be added to a group by calling the async `zigpy.device.Device.add_to_group(group_id, group_name)` method.  A Zigpy Device can be removed from a group by calling the async `zigpy.device.Device.remove_from_group(group_id)` method. Each device has a `member_of` dict, keyed by group_id of which device is member of.

My idea is to add radios as a Zigpy device and each radio could implement its own `add_to_group()`/`remove_from_group()` methods to subscribe/unsubscribe coordinator from receiving traffic for specific groups.

### DIFF
--- a/tests/test_appdb.py
+++ b/tests/test_appdb.py
@@ -8,6 +8,8 @@ from zigpy.application import ControllerApplication
 from zigpy import profiles
 from zigpy.quirks import CustomDevice
 from zigpy.device import Device, Status
+import zigpy.zcl
+from zigpy.zcl.foundation import Status as ZCLStatus
 from zigpy.zdo import types as zdo_t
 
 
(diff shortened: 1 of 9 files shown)


Adminiuga:
Can you post signature for this device (`bellows devices` output) and what model/manufacturer is it?

pipiche38:
I don't as I'm not using your lib yet. But I'm developing an other plugin for Zigate, and I can tell you that most of the 2 bang switches are muti-endpoint and you might not want to have the 2 Ep in the same group !

Adminiuga:
Yep, I'm looking into this now. All of my devices have `Groups` cluster on only one endpoint, but I do have 4 button Osram remote and it does have 6 endpoints and each is with its own Groups cluster but it is an output cluster.  
Can you confirm if those  endpoints have the server or client side of groups cluster?

Adminiuga:
I think you have a point. Although I don't have any of those devices, I can see how it makes sense to have individual groups for multi-channel relay controllers. 
Have ordered one from aliexpress to try.

Hedda:
@pipiche38 I'm curious so wondering if you could maybe list or link to any specific manufacturer and model that have Zigbee switches which have multi-endpoints on the same hardware device?

pipiche38:
Here are the one I have identified and quiet commonly used. You get the list of Endpoint and main clusters, with the type of Action

They are listed with Model Name ( got from Cluster 0x0000 Atrribute 0x0005 )

"lumi.ctrl_ln2":
       "01":{"0006":"","0004":"","Type":"Switch"},
       "02":{"0006":"","0004":"","Type":"Switch"},
       "03":{"000c":"","Type":"Power/Meter"}

lumi.ctrl_ln2.aq1":
      "01":{"0006":"","0004":"","Type":"Switch"},
      "02":{"0006":"","0004":"","Type":"Switch"},
      "03":{"000c":"","Type":"Power/Meter"},

"lumi.ctrl_neutral2":
     "01":{"0002":"","0000":"","0019":"","0001":"","000a":"","0003":"","Type":""},
     "02":{"0005":"","0010":"","0006":"","0004":"","Type":"Switch"}
     ,"03":{"0005":"","0010":"","0006":"","0004":"","Type":"Switch"},
     "04":{"0006":"","0012":"","Type":""}
     ,"05":{"0006":"","0012":"","Type":""},
     "06":{"0006":"","0012":"","Type":""}
     "08":{"000c":"","Type":""}}

Hedda:
@pipiche38 I was also wondering which manufacturers (that is, IKEA, Xiaomi/Aquara, Osram, etc.) and exact models as in part-numbers (manufacturer article number for the product model) of remotes/switches and perhaps URL to where you ordered them from to buy the same products?

pipiche38:
Sorry I do not own those devices. I have integrated them and make it compatible with my plugin

PR #158, opened 2019-05-16
If deserialize data is empty, treat it as optional.

Adminiuga:
Relax length checks for data deserialization introduced in #155 
Current schema definition for ZCL clusters does not allow "optional" parameters, like `Hardware Version` field for OTA cluster "next_image_request" command, which is present only if `field_control == 1`
With checks introduced in #155 the deserialization fails, which is not correct. This PR allows "optional" parameters by returning `None` for `deserialize()` call, but only if input is empty

### DIFF
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -5,8 +5,7 @@ import zigpy.types as t
 
 
 def test_int_too_short():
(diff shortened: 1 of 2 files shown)


Adminiuga:
Closing in favor of #160

PR #160, opened 2019-05-20
Allow "Optional" parameters for Commands/Responses schema.

Adminiuga:
Current schema definition for ZCL clusters or ZDO commands does not allow "optional" parameters.
For example, Hardware Version field for OTA cluster "next_image_request" command is present only if field_control == 1
With checks introduced in #155 the deserialization for such commands/responses fails.
This PR allows marking Schema parameters as "optional", returning None if deserialization fails.


presslab-us:
I've been adding some schema entries for Thermostat and DoorLock and was wondering how to handle the optional parameters.  I will wait until this is merged before I do a pull request for those changes.

Adminiuga:
Currently there is no any schemas for `DoorLock` cluster, so I haven't updated or look into that. 
For commands you are originating (regardless whether it is a server or client side command), just omit the optional parameters and zigpy won't add it into the request.
If there are any optional parameters in the responses to requests you originated, then feel free to wrap it into the `Optional()`.
Specific to DoorLock cluster, i'm not expecting you to have originate any "client_commands", since it's unlikely coordinator to have "input" (server side) door lock cluster, unless in near future "radio" zigpy libraries would implement "router/end device" mode.

presslab-us:
I tried using this pull request and sending the command (without the optional parameter), but it gave me the error:

    File "/root/.homeassistant/custom_components/zha/lock.py", line 108, in async_unlock
      success = await self._doorlock_channel.unlock_door()
    File "/root/.homeassistant/custom_components/zha/core/channels/__init__.py", line 54, in wrapper
      result = await command(*args, **kwds)
    ValueError: Wrong number of parameters for request, expected 1 argument(s)


Also I think there needs to be an Optional.serialize() method if those parameters are actually used?

Here is a snip of the schema that I have modified:

    server_commands = {
        0x0000: ('lock_door', (t.Optional(t.CharacterString), ), False),
        0x0001: ('unlock_door', (t.Optional(t.CharacterString), ), False),
        0x0002: ('toggle_door', (t.Optional(t.CharacterString), ), False),

Adminiuga:
Hrm, Initially I was allowing variable number of parameters for requests, but later removed it.  I do see, that for `DoorLock` cluster the Pin is an optional parameter. However ZCL is not telling whether "optional" means "do not send the parameter" or if it means "send an empty octet string". 

What zigbee lock do you have? Can you try for now sending just an empty octetstring `b''` ?

presslab-us:
It is a Yale YRD210.  The Zigbee module in it is labeled HA1.2 V132.  It can accept both no parameter and an empty octetstring for the Lock Door command.

The documentation states `As of HA 1.2, this command includes an optional
code for the lock.` which means that devices that are older (say HA 1.1) are not expecting any parameters.  The HA 1.1 spec says this about the Lock Door command: `It has no payload.`

presslab-us:
I see something strange.  When I have my schema as `0x0000: ('lock_door', (t.Optional(t.CharacterString), ), False),` and execute `await self._doorlock_channel.lock_door(b'')` then I see the literal `b''` in my packet sniffer.

When I use schema `0x0000: ('lock_door', (t.CharacterString, ), False),` and execute `await self._doorlock_channel.lock_door(b'')` I see the exact thing, so it is not just a problem with Optional.

I think there is something wrong with `CharacterString.serialize`.  This means my previous test was not valid.

sniffer

Adminiuga:
CharacterString is an actual string in dev zigpy, so just do  `await
self._doorlock_channel.lock_door('')`

presslab-us:
Yes, thanks, that works as expected.  It seems my lock doesn't care if the parameter is not there, if it is empty, or if it is `b''`

Adminiuga:
IIRC there was an attribute on DoorLock cluster which controls PIN
requirements. Whether vendor implemented it remains to see :)

presslab-us:
It returns `None` when I query the attribute `RequirePINforRFOperation`, and not `True` or `False`, so I would guess it does not support this feature at all.

Adminiuga:
Are you doing `safe_read()` which swallows exceptions? Just out of curiosity try `res, failed = await door_cluster.read_attributes([require_pin_operation_attr_id])` and see if you get anything in `failed` var.

presslab-us:
I was using the web interface in Home Assistant.  But here is the sniffer capture, "Unsupported Attribute".
Capture

Adminiuga:
@presslab-us I've pushed an update to allow optional request params. Cluster schema needs to have optional params wrapped in `t.Optional()`. Can you give it a try? Same thing as in https://github.com/zigpy/zigpy/pull/160#issuecomment-494524277

presslab-us:
Yes, this works, thanks.  I tested using both no parameter, and one parameter and I can confirm that the data using the packet sniffer looks correct.

PR #166, opened 2019-05-23
Update schema entries for DoorLock and Thermostat clusters

presslab-us:
This fills in the missing schema entries for the DoorLock and Thermostat clusters.


presslab-us:
I split it up, located here https://github.com/zigpy/zigpy/pull/168 and here https://github.com/zigpy/zigpy/pull/169

PR #169, opened 2019-05-24
Fill in missing schema entries for Thermostat cluster

presslab-us:


presslab-us:
Would this work?

`0x0001: ('set_weekly_schedule', (t.enum8, t.bitmap8, t.bitmap8, t.List((t.uint16_t, t.int16s, t.Optional(t.int16s)))), False),`

zigpy/zcl/clusters/hvac.py
```diff
@@ -96,15 +96,15 @@ class Thermostat(Cluster):
         0x0047: ('ac_capacity_format', t.enum8),
     }
     server_commands = {
-        0x0000: ('setpoint_raise_lower', (), False),
-        0x0001: ('set_weekly_schedule', (), False),
-        0x0002: ('get_weekly_schedule', (), False),
+        0x0000: ('setpoint_raise_lower', (t.enum8, t.int8s), False),
+        0x0001: ('set_weekly_schedule', (t.enum8, t.bitmap8, t.bitmap8, t.uint16_t, t.int16s, t.int16s), False),
```
Adminiuga: IMO per my interpretation of specs, this should be `(t.enum8, t.bitmap8, t.bitmap8, t.List(t.uint16))`

otherwise, how would you specify for example 3 transitions for Saturday?

presslab-us: Do you think what I mentioned above would work?  Maybe `t.List` and `t.Optional` won't work right together.

`0x0001: ('set_weekly_schedule', (t.enum8, t.bitmap8, t.bitmap8, t.List((t.uint16_t, t.int16s, t.Optional(t.int16s)))), False),`

Adminiuga: we can check it, but IMO it should work. `t.List(t.uint16_t)` means an array as long as you have bytes data. so `b''` should give an empty list and `b'aabbcc'` should give a list of three ints'

Adminiuga:
```
>>> import zigpy.types as t
>>> schema = t.List(t.uint16_t)
>>> 
>>> schema.deserialize(b'')[0]
[]
>>> schema.deserialize(b'aabbcc')[0]
[24929, 25186, 25443]
>>> 
```

presslab-us: Oh I was asking if you think this specifically `(t.enum8, t.bitmap8, t.bitmap8, t.List((t.uint16_t, t.int16s, t.Optional(t.int16s))))` would work.  If not I have no problem doing the `(t.enum8, t.bitmap8, t.bitmap8, t.List(t.uint16))`

Adminiuga: Uh, I missed those are a mix of uint and ints. But `t.List((t.uint16_t, t.int16s, t.Optional(t.int16s))))` won't work, list items are all of the same type. 

So we have two options here:
- make the schedule its own class and put all logic into serialize/deserialize method
- for now just define it as `(t.enum8, t.bitmap8, t.bitmap8, t.List(t.int16s))` 
The `t.uint16_t` is for "transition time" which number of minutes since midnight, so the range withing `0..1440` range so we could use signed int for it. 

For the last option, the application (HA entity or ZHA Channel) would need to parse the array correctly, depending on "number of transition" and "mode" fields.

presslab-us: I had another parenthesis in for a tuple but I realize it needs to be a type that has deserialize method..  What about this?

    class WeeklyScheduleTrans(t.Struct):
        _fields = [
            ('transition_time', t.uint16_t),
            ('heat_setpoint', t.Optional(t.int16s)),
            ('cool_setpoint', t.Optional(t.int16s)),
        ]
    ...
        0x0001: ('set_weekly_schedule', (t.enum8, t.bitmap8, t.bitmap8, t.List(WeeklyScheduleTrans)), False),

Adminiuga: Not really, cause depending on the `mode` it can have either `('heat_setpoint', t.int16s)` or `('cool_setpoint', t.int16s)` or both. And `t.Optional()` is not of help here, because it would screw things for deserialization (and serialization) if you have more than one transition and only heating or cooling. 
For example, a single transition, for sunday for "cooling" only mode at 12:01am (1 minute past midnight, so I don't have to do hex conversion)  for 25C should be something like: `b'\x01\x00\x02\x01\xc4\09'` where:
- `b'\x01'` -- `t.enum8` -- number of transitions -- 1
- `b'\x00'` -- `t.bitmap8` -- day of the week -- Sunday
- `b'\x02'` -- `t.bitmap8` -- mode -- cooling only
- `b'\x01\xc4\x09` -- Transition: `\x01` -- 1 minute past midnight, `b'\xc4\x09'` is 2500 and should be treated as cooling, because "mode" field has bit #1 set to 1.

deserializing it with a t.Struct would result in something like `[1, 0, 2, [<WeeklyScheduleTrans transition_time=1, heat_setpoint=2500, cool_setpoint=None>]]` not quite what you need or expect. If you change conditions to two transitions, but for "heat only" mode, then you would get something like `[0, 0, 1, [<WeeklyScheduleTrans transition_time=1 heat_setpoint=2500 cool_setpoint=2500>, <WeeklyScheuleTrans transition_time=2, heat_setpoint=None, cool_setpoint=None>]]` again not what is needed. 

I say for now just do it as a list of int16s, otherwise take a look at `zigpy.zcl.foundation.TypeValue` class, as it does something similar, deserializing the remainder of data depending on the 1st few fields.

PR #172, opened 2019-05-29
Fix DoorLock attributes schema.

Adminiuga:
Fix DoorLock cluster attributes schema.


Adminiuga:
@presslab-us can you take a quick look at this? Just need a seconds set of eyes to confirm the attributes schema.

presslab-us:
I looked over them all and only saw one issue, `num_of_lock_records_supported` should be `num_of_log_records_supported`

PR #180, opened 2019-06-16
Refactor endpoint initialization.

Adminiuga:
Skip reading model & manufacturer for non ZHA/ZLL endpoint profiles.
ConBee is having issues reading `model` and `manufacturer` from non ZHA/ZLL endpoints: https://github.com/dresden-elektronik/deconz-rest-plugin/issues/158#issuecomment-502411929 and because of it, device discovery fails https://github.com/zigpy/zigpy-deconz/issues/44 

Since ATM `zigpy` library currently supports only ZHA and ZLL profiles we can skip reading model and manufacturer for those endpoints. The endpoints are still going to be there, so quirk matching would not be affected, as long as `model` and `manufacturer` attributes are not matched for those endpoints.

### DIFF
--- a/tests/test_endpoint.py
+++ b/tests/test_endpoint.py
@@ -213,7 +213,7 @@ async def test_init_endpoint_info_unicode(ep):
     await test_initialize_zha(ep)
 
 
-def _init_endpoint_info(ep, test_manuf=None, test_model=None):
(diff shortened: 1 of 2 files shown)


Adminiuga:
Closing in favor of #181

PR #183, opened 2019-06-20
Quirks Device match refactoring.

Adminiuga:
With the quirks number growing, keeping all the quirks in a flat list and iterating over the entire list for matching is not optimal anymore. 

Since 90% of the quirks could be tied to a specific model and manufacturer, this PR refactors quirks device registry into multidimensional dict, indexed by `manufacturer` and `model`. This drastically reduces noise in the debug logs (5488 lines vs 1556 lines) and load time, most noticeable on the slower hardware like Raspberry PIs

This could be improved even more, by defining at least the `manufacturer` in the signature for quirks that still miss it.

### DIFF
--- a/tests/test_quirks.py
+++ b/tests/test_quirks.py
@@ -55,7 +55,7 @@ def _dev_reg(device):
     return registry
 
 
-def test_get_device(real_device):
+def test_get_device_new_sig(real_device):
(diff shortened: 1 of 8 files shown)


dmulcahey:
Nice work. This looks great.

PR #197, opened 2019-08-09
Don't initialize a device that is already initialized

dmulcahey:
We already had a check to not initialize devices that were already initialized. There are cases when a device that has already initialized will get issued a new nwk address. We handled those cases correctly but we were also pumping those devices back through the initialization process. With this change we will only send devices through initialization that aren't already initialized or aren't already in the process of initializing even when the nwk address changes.

### DIFF
--- a/zigpy/device.py
+++ b/zigpy/device.py
@@ -41,11 +41,7 @@ class Device(zigpy.util.LocalLogMixin):
         self._node_handle = None
 
     def schedule_initialize(self):
-        if self.initializing:
-            LOGGER.debug("Canceling old initialize call")
-            self._init_handle.cancel()
-        else:
-            self.initializing = True
+        self.initializing = True
         self._init_handle = asyncio.ensure_future(self._initialize())
 
     async def get_node_descriptor(self):
--- a/tests/test_application.py
+++ b/tests/test_application.py
@@ -54,7 +54,9 @@ async def test_permit(app, ieee):
     ncp_ieee = t.EUI64(map(t.uint8_t, range(8, 16)))
     app._ieee = ncp_ieee
     app.devices[ieee] = mock.MagicMock()
-    app.devices[ieee].zdo.permit = mock.MagicMock(side_effect=asyncio.coroutine(mock.MagicMock()))
+    app.devices[ieee].zdo.permit = mock.MagicMock(
+        side_effect=asyncio.coroutine(mock.MagicMock())
+    )
     app.permit_ncp = mock.MagicMock(side_effect=asyncio.coroutine(mock.MagicMock()))
     await app.permit(node=(1, 1, 1, 1, 1, 1, 1, 1))
     assert app.devices[ieee].zdo.permit.call_count == 0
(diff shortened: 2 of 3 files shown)


dmulcahey:
the 3rd commit:

```
        if self.initializing:
            LOGGER.debug("Canceling old initialize call")
            self._init_handle.cancel()
        else:
            self.initializing = True
```

I don't think this logic is needed any more because we shouldn't be able to get to this based on the change to application.py

dmulcahey:
and whoops... I black'd a test 🤷‍♂

Adminiuga:
>the 3rd commit:
```
        if self.initializing:
            LOGGER.debug("Canceling old initialize call")
            self._init_handle.cancel()
        else:
            self.initializing = True
```

Nah,  we def need it, because we want to go with the latest `initialize()` call, i.e. if we get multiple `handle_join()` in quick succession, usually device leaves and rejoins (or user is still keeping on reset button) so there's no guarantee the device did not reset itself since the 1st initialization call started.

dmulcahey:
So how about we remove the initializing part from the check in application.py

Adminiuga:
Hrm, not following. If we remove the check, do you want to return immediately or go through initialization?

Also, any concerns https://github.com/home-assistant/home-assistant/blob/fafd2284182ed3a973f1916e41923395d4b47625/homeassistant/components/zha/core/gateway.py#L145 not being triggered anymore on NWK update?

dmulcahey:
hrm, not sure it would matter... should act just like the “skipping...” from the other condition there when a device just rejoins normally. The only difference here is the nwk flip. What we need is a way to determine if the nwk just changed or if the device reset. I think undoing this and copying listeners over in quirks has less side effects to think about.

Adminiuga:
with this change your the following events are inhibited:
- `device_joined()`
- `raw_device_initialized()`
- `device_initialized()`

So Zigpy would update NWK and we assume all endpoints were initialized. Does ZHA need to update any IDs of its channels?

Adminiuga:
Like `ZigbeeChannel._log_id`???

zigpy/device.py
```diff
@@ -41,11 +41,7 @@ def __init__(self, application, ieee, nwk):
         self._node_handle = None
 
     def schedule_initialize(self):
-        if self.initializing:
-            LOGGER.debug("Canceling old initialize call")
-            self._init_handle.cancel()
-        else:
-            self.initializing = True
+        self.initializing = True
```
Adminiuga: No, if we do `asyncio.ensure_future()` then there's a chance this would be called twice in succession. We do need to cancel old task. 
And please run this change in your production for a week and report back :)

dmulcahey:
I think that’s the only one... and a better solution is to just assemble that dynamically and then it doesn’t matter, it’s always up to date. What I’m saying above is completely revert this entire change. Then in quirks, copy listeners from the device to the quirked device. That way we don’t lose listeners. 

Actually, can’t we just check if the device is an instance of a custom device? And skip the reapplication of quirks all together in Zigpy?

dmulcahey:
closing in favor of a different solution

PR #198, opened 2019-08-10
RFC - Don't quirk quirked devices

dmulcahey:


Adminiuga:
don't forget to rebase against current dev. It looks like your starting point is a bit old.

PR #213, opened 2019-09-06
Initial implementation of Zigbee OTA updates.

Adminiuga:
Handle OTA cluster requests.
**This is an experimental feature** and is disabled by default. By enabling it **you assume all responsibility** for any unintended consequences.
To  enable file based OTA, create a `zigpy_ota` directory in the same dir as `zigbee.db`. To enable IKEA provider, create an empty `enable_ikea_ota` file in `zipgy_ota/` directory.

This PR only implements OTA cluster command handling and does not distribute any firmware files. It will download firmware for IKEA devices if it is enabled and there're IKEA devices on the network with an outdated firmware. The only other vendor distributing OTA files for their products is [Dresden Elektronik](dresden-elektronik.de)

Fixes (kind of) #154


martin3000:
Is there any documentation how OTA will work? Where do the firmware images come from?

Adminiuga:
Check the issue section, there's RFC for OTA.

PR #222, opened 2019-09-11
fix binding dst endpoint

abmantis:
When binding, the destination endpoint (the endpoint on the coordinator stick) should be 1.
This could be improved in the future to receive the dst endpoint as a parameter.

### DIFF
--- a/zigpy/zdo/__init__.py
+++ b/zigpy/zdo/__init__.py
@@ -90,14 +90,14 @@ class ZDO(zigpy.util.LocalLogMixin, zigpy.util.ListenableMixin):
         dstaddr = types.MultiAddress()
         dstaddr.addrmode = 3
         dstaddr.ieee = self._device.application.ieee
-        dstaddr.endpoint = endpoint
+        dstaddr.endpoint = 1
         return self.Bind_req(self._device.ieee, endpoint, cluster, dstaddr)
 
     def unbind(self, endpoint, cluster):
         dstaddr = types.MultiAddress()
         dstaddr.addrmode = 3
         dstaddr.ieee = self._device.application.ieee
-        dstaddr.endpoint = endpoint
+        dstaddr.endpoint = 1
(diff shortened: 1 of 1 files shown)


dmulcahey:
Interesting... wonder if this affects deconz only? I found this: https://github.com/dresden-elektronik/deconz-rest-plugin/wiki/Pairing-Devices search "bind" and in the paragraph for the first occurrence it specifically states: "For attribute reports, the binding needs to be from a server cluster to endpoint 0x01 of the coordinator (RaspBee or ConBee)"

dmulcahey:
too add more info: https://github.com/zigpy/zigpy/commit/5c635700299e983bcc6b6672a3dc27e5bb088bca

This used to be like this and was changed 2 years ago... wonder what else this could impact?

Adminiuga:
My guess it would affect mostly ConBees, as those are quite specific and filter out all traffic if not matching endpoint ID, profile_id configured on the cluster etc. 

For ConBee we do discovery call and create a zigpy device for coordinator. Perhaps should use that, just in case user decides to change the default endpoint id. 

But I really would like to know what triggered https://github.com/zigpy/zigpy/commit/5c635700299e983bcc6b6672a3dc27e5bb088bca change

abmantis:
According to the deconz docs, the endpoint is always 1 for the coordinator.

Adminiuga:
The default one -- yes. Users can change it
image

abmantis:
Weird. Because the docs say "For attribute reports, the binding needs to be from a server cluster to endpoint 0x01 of the coordinator (RaspBee or ConBee);" :

Adminiuga:
as a test, change your endpoint ID to 3 and see if you get the reports from the plugs

Adminiuga:
Actually it just struck me why my Centralite Motion sensor won't report occupancy. You diagnosis is correct and this is a good find. Give me some time, as I'd like to think about this a little bit more. ATM i'm inclined to use the endpoint ID from coordinator zigpy device instance and failback  to endpoint id 1

PR #228, opened 2019-09-14
Add new Samsung Multipurpose Sensor

Thalagyrt:
Hiya!

I just added a few new Samsung multipurpose sensors to ZHA in my HomeAssistant setup. They have a slightly different input cluster signature and thus don't match the quirks for the older unit. This appears to be fully functional within HomeAssistant with acceleration detection, temperature, open/close, as expected.

If there's a better way to do this (integrate it into the existing `SmartthingsMultiPurposeSensor` class?) let me know and I'll adjust my PR.

I didn't see any tests for specific quirks implementations, so didn't add any here.

Cheers!
James

### DIFF
--- a/zigpy/quirks/smartthings/__init__.py
+++ b/zigpy/quirks/smartthings/__init__.py
@@ -100,3 +100,46 @@ class SmartthingsMultiPurposeSensor(CustomDevice):
             }
         }
     }
+
+
+class SmartthingsMultiPurposeSensor2019(CustomDevice):
+    signature = {
+        "endpoints": {
+            # <SimpleDescriptor endpoint=1 profile=260 device_type=1026
+            # device_version=0 input_clusters=[0, 1, 3, 32, 1026, 1280, 2821, 64514]
+            # output_clusters=[3, 25]>
+            1: {
+                "profile_id": 0x0104,
+                "device_type": 0x0402,
+                "input_clusters": [
+                    0,
+                    1,
+                    3,
+                    32,
+                    1026,
+                    1280,
+                    2821,
+                    SmartThingsAccelCluster.cluster_id,
+                ],
+                "output_clusters": [3, 25],
+            }
+        }
+    }
+
+    replacement = {
+        "endpoints": {
+            1: {
+                "input_clusters": [
+                    0x0000,
+                    0x0001,
+                    0x0003,
+                    0x0020,
+                    0x0402,
+                    0x0500,
+                    0x0B05,
+                    SmartThingsAccelCluster,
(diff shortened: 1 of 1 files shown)


Adminiuga:
@Thalagyrt can you please provide "manufacturer" and "model" attributes for this device and create a PR in https://github.com/dmulcahey/zha-device-handlers ?

Thalagyrt:
Sure. To be clear before I PR anything, are you suggesting that this quirk class should be added solely in that repo under say `zhaquirks/samjin/multi2.py` instead of to zigpy? It's an update to the Samjin Multipurpose sensor sold as the Smartthings Multipurpose sensor - it looks identical to last year's 2018 Samjin, but has a slight hardware change with the `2821` input: https://smile.amazon.com/Samsung-SmartThings-Multipurpose-Sensor-GP-U999SJVLAAA/dp/B07F956F3B/

I have one of the 2018 Samjins that looks otherwise identical and the `SmartthingsMultipurposeSensor` quirk works with it.

Adminiuga:
@dmulcahey Adding David.
My guess either adding `multiv2.py` or adding another class with your signature.  Just wanted to know what do you get for that device for `manufacturer` and `model` attributes as we want to include those into signature.

Adminiuga:
PS: I think I've got mine in 2018 and mine has the following signature. Apparently there're quite a few models of this device
```
Discovered endpoint information: <SimpleDescriptor endpoint=1 profile=260 device_type=1026 device_version=0 input_clusters=[0, 1, 3, 32, 1026, 1280, 64514] output_clusters=[3, 25]>
Manufacturer: Samjin
Model: multi
```

Thalagyrt:
Yeah, that's the previous signature prior to the refresh. I have one of that version as well. That quirk is in the file I modified, it's the quirk right above the one I added in this PR, which is why I made the PR here instead of the quirks repo: https://github.com/zigpy/zigpy/blob/c7777c467777d03ba6bb1b5e59896abb892bbe05/zigpy/quirks/smartthings/__init__.py#L64

Maybe both of these should be moved over to `zha-device-handlers` for consistency? Not sure how you want to keep this organized, as right now these quirks seem to live in different repos.

Adminiuga:
New quirk implementations go to `zha-device-handlers` . Probably should move all quirks to zha-device-handlers for consistency.

Thalagyrt:
OK. I'll close this PR out and add this quirk over there instead, will tag you in the PR shortly. Yeah, consistency would be nice. Let me know if you want some help with that, I'd be happy to whip up some PRs.

PR #236, opened 2019-09-28
Don't use Command enum for cluster specific commands.

Adminiuga:
Don't `repr()` cluster specific commands as Global commands.


amigan:
This fixes my colorloop patch (https://github.com/home-assistant/home-assistant/pull/26549) in homeassistant 0.100.2/zigpy-homeassistant 0.9.0 (as shipped). Otherwise, color_loop_set (0x44) does not exist in the global commands enum and a traceback is generated.

PR #259, opened 2019-11-19
Extract IKEA OTA image offset and size directly from the container

puddly:
The chances of finding the four byte OTA header erroneously are slim but you can avoid searching by just reading the offset straight from the container file.


Adminiuga:
Where did you get NGIS container format?

puddly:
I was only really looking for those two values so the rest of the container is beyond me. I can't find any information on it anywhere. I think I saw something about some OTA generator program by NXP? in some documentation but I can't find it now.

PR #266, opened 2019-12-01
Allow inactive endpoints

dmulcahey:
This PR accounts for endpoints that have the status of inactive. This allows devices such as: lumi.remote.b286opcn01, lumi.remote.b486opcn01, lumi.remote.b686opcn01 to join the network as they report 6 endpoints each but 4 of them report inactive status.

### DIFF
--- a/zigpy/endpoint.py
+++ b/zigpy/endpoint.py
@@ -47,7 +47,11 @@ class Endpoint(zigpy.util.LocalLogMixin, zigpy.util.ListenableMixin):
             sdr = await self._device.zdo.Simple_Desc_req(
                 self._device.nwk, self._endpoint_id, tries=3, delay=2
             )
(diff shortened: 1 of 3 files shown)


zigpy/endpoint.py
```diff
@@ -47,7 +49,11 @@ def __init__(self, device, endpoint_id):
             sdr = await self._device.zdo.Simple_Desc_req(
                 self._device.nwk, self._endpoint_id, tries=3, delay=2
             )
-            if sdr[0] != 0:
+            if sdr[0] == 131:
```
Adminiuga: Can you use the actual Status enum in comparison pls?

Adminiuga: I can remember that 0 is a SUCCESS status, but can't remember what 131 is :)

PR #278, opened 2020-01-15
add ledvance fw provider

billyburly:
Adds provider to support Ledvance zigbee firmware updates. Provides updates for Sylvania, Osram, etc. Tested against Osram Lightify Gardenspot RGB.

### DIFF
--- a/zigpy/ota/provider.py
+++ b/zigpy/ota/provider.py
@@ -15,7 +15,8 @@ LOGGER = logging.getLogger(__name__)
 LOCK_REFRESH = "firmware_list"
 
 ENABLE_IKEA_OTA = "enable_ikea_ota"
-SKIP_OTA_FILES = (ENABLE_IKEA_OTA,)
+ENABLE_LEDVANCE_OTA = "enable_ledvance_ota"
+SKIP_OTA_FILES = (ENABLE_IKEA_OTA,ENABLE_LEDVANCE_OTA,)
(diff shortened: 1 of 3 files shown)


Adminiuga:
Thanks! I've glanced over and looks good. Will take a better look when have more time.

zigpy/ota/provider.py
```diff
@@ -161,6 +162,101 @@ class Trådfri(Basic):
...
+        async with aiohttp.ClientSession() as req:
+            LOGGER.debug("Downloading %s for %s", self.url, self.key)
+            async with req.get(self.url) as rsp:
+                data = await rsp.read()
+
+        assert len(data) > 24
+        offset = data.index(self.OTA_HEADER)
+        assert offset >= 0
+
+        img, _ = OTAImage.deserialize(data[offset:])
```
Adminiuga: Are Ledvance OTA images not starting with the OTA header? the Ikea images were not standard, wrapped in another container and needed the checks above. If they are standard OTA images you can skip lines 203--205 and just do `img, _ = OTAImage.deserialize(data)`

billyburly: The Ledvance images do start with the header. Like you inferred, I followed the IKEA provider. Will remove the check

Adminiuga: Don't worry about it for now.

PR #299, opened 2020-02-27
Send content-type header on firmware get requests

walthowd:
Explicitly send a content type header on firmware get requests to allow aiohttp json decoder to work more reliably.

### DIFF
--- a/zigpy/ota/provider.py
+++ b/zigpy/ota/provider.py
@@ -131,6 +131,7 @@ class Trådfri(Basic):
 
     UPDATE_URL = "https://fw.ota.homesmart.ikea.net/feed/version_info.json"
     MANUFACTURER_ID = 4476
+    HEADERS = {'content-type': 'application/json'}
 
     async def initialize_provider(self, ota_dir: str) -> None:
         if ota_dir is None:
@@ -147,7 +148,7 @@ class Trådfri(Basic):
 
         LOGGER.debug("Downloading IKEA firmware update list")
         async with self._locks[LOCK_REFRESH]:
-            async with aiohttp.ClientSession() as req:
+            async with aiohttp.ClientSession(headers=self.HEADERS) as req:
                 async with req.get(self.UPDATE_URL) as rsp:
                     fw_lst = await rsp.json(content_type=None)
         self._cache.clear()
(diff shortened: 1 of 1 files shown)


dmulcahey:
Shouldn’t we update the content type in the gets now too?

Adminiuga:
Well, that's a good question.  supposedly as it is now it shouldn't care and parse everything as json?

@walthowd what happens without this patch and #265 reverted? i.e with 
`fw_lst = await rsp.json(content_type="application/octet-stream")`  if you don't set the header in the sesssion.

And second questsion: do you get any response at all if you don't send headers?

walthowd:
@Adminiuga For me in the US, it worked fine without this patch and rolling back #265. That left me with:

`fw_lst = await rsp.json(content_type="application/octet-stream")`

For the PR I was assuming we only need to fix the finicky JSON decoder in aiohttp -- The other gets of the binary firmware files all worked fine for me without any changes.

dmulcahey:
Best practice (esp if the server supports it) is to tell the server what you’re expecting back. If “none” should just work then this isn’t necessary I guess. 🤷‍♂

Adminiuga:
I guess I need to catch up to discord 1st. I wish discord could do message threads. With all that traffic it is hard to keep up

Adminiuga:
@walthowd add another log lines for each provider after it finishes downloading the list.

walthowd:
Tried late last night on 105.5 and 106.1 but could not reproduce original failure. Wondering now if it was a transient network issue.

Adminiuga:
And I'm on the fence whether to enabled the headers. Not that I think it's wrong, just have no way of knowing  what those servers really need. 
TBH, i think we should have the content-type header set, just don't want to unintentionally break things, since it was working without it before.
So I'm open to adding the headers, but lets also move the https://github.com/zigpy/zigpy/blob/f671be43a24945be0c50b543c788e5d7927337bd/zigpy/ota/provider.py#L149  and log the message after we finished downloading the list. We already know when each provider is started and if we move the log line, then we'll know for sure each list was downloaded too. 

And if we'll break things, then we'll remove it in the next version.

puddly:
I don't think adding a `Content-Type` header to a GET request will have any impact on `aiohttp` because the `json` method only ever checks the `Content-Type` of the *response* (as it should):

https://github.com/aio-libs/aiohttp/blob/91963c0e96bc5ea4d884cb413d2f65c637f8c4c6/aiohttp/client_reqrep.py#L962-L982

Sending a `Accept: application/json` would be the "proper" way to ask IKEA to only send back a JSON response, though it currently has no effect and might even trigger a [HTTP 406 error](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406) error in the future. Maybe `Accept: application/json,*/*` is better?

PR #305, opened 2020-02-29
Use Accept instead of Content-Type in firmware provider GET requests

puddly:
The `Content-Type` header introduced in #299 will be ignored by any compliant HTTP server when sent with a GET request. `Accept` is the appropriate header to use here, since we're expecting only JSON responses.

I've also documented the purpose of the `response.json(content_type=None)` line and removed that relaxation for the Ledvance provider, since their responses already have the appropriate `Content-Type`.

### DIFF
--- a/zigpy/ota/provider.py
+++ b/zigpy/ota/provider.py
@@ -137,7 +137,7 @@ class Trådfri(Basic):
 
     UPDATE_URL = "https://fw.ota.homesmart.ikea.net/feed/version_info.json"
     MANUFACTURER_ID = 4476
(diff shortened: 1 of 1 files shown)


dmulcahey:
This was the original intent... we just didn’t do it correctly. Thanks for correcting this.

Adminiuga:
Do you think we should accept `application/octet-stream` as well?

Adminiuga:
Nevermind, you are already doing that. Thank you!

puddly:
Not explicitly but I don't think it really matters, it doesn't seem like the `Accept` header is being looked at.

PR #339, opened 2020-03-27
Add ability to set constant values for attributes on CustomCluster

dmulcahey:


zigpy/quirks/__init__.py
```diff
@@ -95,3 +97,41 @@ def set_device_attr(attr):
 
 class CustomCluster(zigpy.zcl.Cluster):
     _skip_registry = True
+    _CONSTANT_ATTRIBUTES: Optional[Dict[int, Any]] = None
```
Adminiuga: Do we need this as private? 🤷‍♂

Issue #326, opened 2020-03-21
Alot of warnings "FOREIGN KEY constraint failed" in log


MaximumSU:
I setup new zha network with some XBee modules. I adding, and configuring 8 modules, disable unneeded switches and sensors in device configs. All looking good except of one node. That node send the IO sampling of enabled pins every 1 minute. And every minute i see in log 6 messages
"FOREIGN KEY constraint failed":
```
2020-03-21 18:05:50 DEBUG (MainThread) [zigpy_xbee.uart] Frame received: b'\x91\x00\x13\xa2\x00AUJ\x1a\x03F\xe8\xe8\x00\x92\xc1\x05\x01\x01\x0c\xb8\x00\x00('
2020-03-21 18:05:50 DEBUG (MainThread) [zigpy_xbee.api] Frame received: explicit_rx_indicator
2020-03-21 18:05:50 DEBUG (MainThread) [zigpy_xbee.api] _handle_explicit_rx: (00:13:a2:00:41:55:4a:1a, 0x0346, 232, 146, 1, b'010cb8000028')
2020-03-21 18:05:50 DEBUG (MainThread) [zigpy.zcl] [0x0346:232:0x0092] ZCL deserialize: <ZCLHeader frame_control=<FrameControl frame_type=CLUSTER_COMMAND manufacturer_specific=False is_reply=False disable_default_response=False> manufacturer=None tsn=5 command_id=0>
2020-03-21 18:05:50 DEBUG (MainThread) [zigpy.zcl] [0x0346:232:0x0092] ZCL request 0x0000: [{'digital_pins': [0, 0, 0, 1, 1, 1, 0, 1, 0, 0, 1, 1, 0], 'analog_pins': [0, 0, 0, 0, 0, 0, 0, 0], 'digital_samples': [0, 0, 0, 1, 0, 1, 0, 0, 0, 0, 0, 0, 0], 'analog_samples': [0, 0, 0, 0, 0, 0, 0, 0]}]
2020-03-21 18:05:50 WARNING (MainThread) [zigpy.util] Error calling listener.attribute_updated: FOREIGN KEY constraint failed
2020-03-21 18:05:50 WARNING (MainThread) [zigpy.util] Error calling listener.attribute_updated: FOREIGN KEY constraint failed
2020-03-21 18:05:50 WARNING (MainThread) [zigpy.util] Error calling listener.attribute_updated: FOREIGN KEY constraint failed
[...]

PR #343, opened 2020-03-31
Relax attributes table schema.

Adminiuga:
Quirks often update attributes on unexisting clusters and endpoints.

### DIFF
--- a/zigpy/appdb.py
+++ b/zigpy/appdb.py
@@ -11,7 +11,7 @@ from zigpy.zdo import types as zdo_t
 
 LOGGER = logging.getLogger(__name__)
 
-DB_VERSION = 0x0002
+DB_VERSION = 0x0003
 
 
 def _sqlite_adapters():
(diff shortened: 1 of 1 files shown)


pdecat:
I was facing the issue described in #326 so I looked at this change to figure how to get rid of it.

How will this change apply to existing `attributes` tables?
Is there an automated migration path planned?

As sqlite does not seem to support altering a table's foreign constraints (see https://www.techonthenet.com/sqlite/foreign_keys/drop.php), I went ahead with the suggested solution and executed the following DDL after having stopped the HA instance and backing up the zigbee.db file:

```
PRAGMA foreign_keys=off;

BEGIN TRANSACTION;

ALTER TABLE attributes RENAME TO _attributes_old;

CREATE TABLE attributes  (
  ieee ieee,
  endpoint_id,
  cluster,
  attrid,
  value,
  FOREIGN KEY(ieee) REFERENCES devices(ieee) ON DELETE CASCADE
);

INSERT INTO attributes SELECT * FROM _attributes_old;

COMMIT;

PRAGMA foreign_keys=on;
```

PR #352, opened 2020-04-02
Refactor enums.

Adminiuga:
Allow instantiation using string values, eg `t.enum8("42")` or `t.enum8("0x28")` to maintain backward compatibility from the times when `enum`s where just `int`s.


elupus:
Old change, but i just ran into that this can't instantiate from a string name. When used from HA UI, you get the string name printed in a field on read, but trying to write the same value fails since it will try to int() cast the value.

It should likely check that it's numeric before trying to intify.

PR #366, opened 2020-04-12
Requires specific methods to be implemented by all libs.

Adminiuga:
Require the library libs to implement the following methods:
- `shutdown()`
- `startup()`
- `request()`
- `permit_ncp()`

### DIFF
--- a/tests/test_appdb.py
+++ b/tests/test_appdb.py
@@ -15,8 +15,32 @@ from zigpy.zdo import types as zdo_t
 
 
 def make_app(database_file):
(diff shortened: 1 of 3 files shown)


Hedda:
@Adminiuga Do you know if "reset" (for restart or reboot) is also a common method as well or not?

Adminiuga:
No, that is left to the radio lib to implement. Some radios like EZSP need full blown reset to start operations, some like ConBee are pretty much good to go once they are plugged in

PR #396, opened 2020-04-27
Fix URLs for zigpy, zigpy-cc, zigpy-zigate, zha-device-handlers, zha,…

Gamester17:
… plus nitpick

Fix URLs for zigpy, zigpy-cc, zigpy-zigate, zha-device-handlers, zha, plus nitpick

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,11 +7,13 @@
 
 Zigbee Home Automation integration with zigpy allows you to connect one of many off-the-shelf Zigbee adapters using one of the available Zigbee radio library modules compatible with zigpy to control Zigbee based devices. There is currently support for controlling Zigbee device types such as binary sensors (e.g., motion and door sensors), sensors (e.g., temperature sensors), lightbulbs, switches, and fans.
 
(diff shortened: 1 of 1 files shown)


Adminiuga:
> zigpy works with separate radio libraries which can each interface with multiple USB and GPIO radio hardware adapters/modules over different native UART serial protocols

Zigpy doesn't work/use separate radio libraries. Zigpy is just a library which contains common code between the radios, pertaining to Zigbee ZCL, ZDO and application state management. Radio libs are extending zigpy ControllerApplication class, not the other way around.
I think it is better

>  zigpy contains common code implementing Zigbee ZCL, ZDO and application state management which is being used by various radio libraries implementing actual interface with the radio modules.

Gamester17:
How about this?

> zigpy contains common code implementing Zigbee ZCL, ZDO and application state management which is being used by various radio libraries implementing the actual interface with the radio modules from different manufacturers. The separate radio libraries interface with radio hardware adapters/modules over USB and GPIO using different native UART serial protocols. 
> 
> Radio libraries for zigpy includes ...

Gamester17:
Like so?

PR #403, opened 2020-05-04
Update README.md with related projects

Hedda:
Update README.md with related projects as thought be good to let new zigpy developers (and users) know of these in case they would like to help out, reuse the code, or utilize them for related projects.


Hedda:
@

PR #404, opened 2020-05-04
Update README.md adding a troubleshooting section about debug logs from HA

Hedda:
Update README.md adding a troubleshooting section about debug logs from Home Assistant


puddly:
It might also be useful to duplicate the HA debugging config into the project issue template. Many submitted issues require a few interactions with the submitter before there's enough context to do anything.

PR #416, opened 2020-05-15
Fix `isinstance` check for optional `Struct` subclasses

puddly:
`Struct.__init__` uses `isinstance(other, self.__class__)` for its "copy constructor", which silently fails if a type is declared optional by using `t.Optional(SomeStruct)`. This fix feels a little hacky because of the MRO list traversal and relying on the `Optional` class's name but I can't really see an alternative that doesn't involve redoing a bunch of type stuff elsewhere in the codebase.

### DIFF
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -215,6 +215,17 @@ def test_struct_init():
     assert ts.c == "TestStruct"
 
 
+def test_struct_optional_init():
+    class TestStruct(t.Struct):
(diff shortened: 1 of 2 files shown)


Adminiuga:
In reality `Optional`  itself was a hack, but I didn't have a better idea at the time.
Is there a better way to implement `Optional` type?

puddly:
You can probably replace the entire `Struct` class with the `attrib` module by putting an `'optional'` key in the attributes' metadata: https://www.attrs.org/en/stable/extending.html#metadata

PR #444, opened 2020-07-22
Make fixed-size integers and Enum subclasses' types behave properly

puddly:
The major improvement here is integer `Enum`s' and their members' types now make sense. This simplifies a bit of code.

Before:

```Python
import zigpy.types as t

class TestEnum(t.enum8):
    Member = 0x00

assert isinstance(TestEnum.Member, t.uint8_t) is False  # this is annoying
assert type(TestEnum.Member.value) is int  # this breaks a lot of stuff
TestEnum.Member.serialize()  # works only because TestEnum itself has a serialize method
```

After:

```Python
assert isinstance(TestEnum.Member, t.uint8_t) is True
assert type(TestEnum.Member.value) is t.uint8_t
TestEnum.Member.serialize()  # no special handling needed
```

Other improvements:

 - `t.uint8_t(-1)` and `t.uint8_t(256)` throw a `ValueError`. The former's `OverflowError` is re-raised as a `ValueError`.

Problems:

 - `t.LVList` not being an actual type breaks a few things relying on `isinstance` and `issubclass` to work. While fixing my introduced bug I found a new one with `pytype_to_datatype_id` whose root cause was also `t.LVList`. I fixed this in zigpy-znp [by making `LVList` a real type](https://github.com/zha-ng/zigpy-znp/blob/660423511ac548f45e6a88117937ac92c6ecdafe/zigpy_znp/types/basic.py#L176-L228) but this has the downside (upside?) of requiring every subclass of `t.LVList` to be explicitly named and not constructed anonymously. It would be cool to have `t.LVList[t.uint8_t, ValueType]` be compatible with `class SubType(t.LVList, length_type=t.uint8_t, item_type=ValueType):  pass`

### DIFF
--- a/tests/test_struct.py
+++ b/tests/test_struct.py
@@ -119,7 +119,7 @@ def test_struct_attribute_assignment():
     s1 = TestStruct(a=1)
     s1.a = -1
 
-    with pytest.raises(OverflowError):
+    with pytest.raises(ValueError):
         s1.serialize()
 
     s1.a = 1
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -4,6 +4,30 @@ import pytest
 import zigpy.types as t
 
 
+def test_abstract_ints():
+    assert issubclass(t.uint8_t, t.uint_t)
+    assert not issubclass(t.uint8_t, t.int_t)
+    assert t.int_t._signed is True
+    assert t.uint_t._signed is False
+
+    with pytest.raises(TypeError):
+        t.int_t(0)
+
+    with pytest.raises(TypeError):
+        t.FixedIntType(0)
+
+
+def test_int_out_of_bounds():
+    t.uint8_t(0)
+
+    with pytest.raises(ValueError):
+        # Normally this would throw an OverflowError. We re-raise it as a ValueError.
+        t.uint8_t(-1)
+
+    with pytest.raises(ValueError):
+        t.uint8_t(0xFF + 1)
+
+
 def test_int_too_short():
     with pytest.raises(ValueError):
         t.uint8_t.deserialize(b"")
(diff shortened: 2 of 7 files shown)


puddly:
I'm also now debating whether or not to revert `HexRepr` back to a mixin or to keep it on the metaclass.

Adminiuga:
hrm, somehow I've missed this one. 

would `isinstance(TestEnum.Member, t.enum8) is True` still be the case? I don't remember the exact reason, but I think I wanted `TestEnum.Member` not being an instance of `t.uint8_t` Something with type handling/conversion, as some devices were picky about correct attribute type id being used for operaitons

puddly:
Both would be true because `issubclass(t.enum8, t.uint8_t)` is True:

```Python
In [1]: import zigpy.types as t

In [2]: class TestEnum(t.enum8):
  ...     Member = 0x01
  ...

In [3]: isinstance(TestEnum.Member, t.enum8)
Out[3]: True

In [4]: isinstance(TestEnum.Member, t.uint8_t)
Out[4]: True
```

Adminiuga:
IIRC need to make sure that `Cluster.write_attributes({0x0000: TestEnum.Member})` would use `enum8` data type ID in `WriteAttributeRecord` which I think still is the case, because it would not depend on `TestEnum.Member` but on that attr definition

puddly:
Hmm, I'll have to try to make that a unit test later.

Adminiuga:
i think I already have it in https://github.com/zigpy/zigpy/blob/670564995a56fad243bc79bc0bbb2ebd64678e7e/tests/test_zcl.py#L396 so since it wasn't triggered i guess this change is fine.

Just test it in prod and if everything is fine, then we can merge it.

puddly:
It works with new device joining and discovery but I really don't have much else to test at the moment.

PR #446, opened 2020-07-25
Concrete List/LVList/FixedList types

puddly:
The main change is overloading subclass and instance checks in the list metaclass so that the classes can be used as types:

```Python
class LVListSubclass(t.LVList, length_type=t.uint8_t, item_type=t.uint16_t):
    pass

assert issubclass(LVListSubclass, t.LVList[t.uint8_t, t.uint16_t])
assert isinstance(LVListSubclass(), LVListSubclass)
assert isinstance(LVListSubclass(), t.LVList[t.uint8_t, t.uint16_t])
```

This still allows for the creation of anonymous `t.LVList`s that work as you'd expect. With #444, most of the weirdness in the type system will be fixed and hacks relying on the names of classes can be removed (e.g. https://github.com/zigpy/zigpy/blob/41986610e84a51cd89ff09701ce1a055d7c9a6a6/zigpy/zcl/foundation.py#L136-L152).

### DIFF
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -263,7 +263,20 @@ def test_list():
     assert isinstance(d, t.List[t.uint8_t])
 
 
-def test_list_types():
+def test_fixedlist():
+    with pytest.raises(ValueError):
+        t.FixedList[t.uint8_t, 2]([]).serialize()
+
+    with pytest.raises(ValueError):
+        t.FixedList[t.uint8_t, 2]([1]).serialize()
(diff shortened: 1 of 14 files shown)


puddly:
I'm not 100% happy with the complexity of the list metaclass and the duplicated kwarg dictionary to get `t.LVList[length_type, item_type]` working but I'm hoping to maybe use it for other classes in the future that have a similar structure.

PR #460, opened 2020-08-20
Update setup.py and tox.ini to fully setup the testing env with pip

puddly:
Allows you to do `pip install zigpy[testing]` and then just run `pytest`. 

Also, `setup.py test` is deprecated (https://github.com/pypa/setuptools/issues/1684) and `tests_require` aren't used by `tox`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,25 +1,21 @@
 """Setup module for zigpy"""
 
-from os import path
+import pathlib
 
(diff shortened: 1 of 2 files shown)


puddly:
An alternative would be to rename `testing` to `dev` and add `tox` and `pre-commit` as dependencies.

PR #470, opened 2020-09-02
Update CONTRIBUTING.md

pipiche38:
Hereby I'm trying to document what could be the MainListener template. 
I think that each call back needs to be introduced here with their correct API

In such it will ease the new comers to develop a first app.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -125,6 +125,69 @@ Run `pytest -lv`, which will show you a stack trace and all the local variables 
 
 This section is meant to describe the zigpy API (Application Programming Interface) and how-to to use it.
 
+### MainListener template
+
+Here should be describe what are the skeleton of a MainListner, what are the methods to be define there in order to use event/commands from below layer
+
+```
+class MainListener:
+    """
+    Contains callbacks that zigpy will call whenever something happens.
+    Look for `listener_event` in the Zigpy source or just look at the logged warnings.
+    """
+    def __init__(self, application):
+        pass
+
+    def device_joined(self, device):
+        # Call everytime a device join the network
+        pass
+
+
+    def device_announce(self, zigpy_device):
+        # Call at device annoucement
+        pass
+
+
+    def device_initialized(self, device, *, new=True):
+        # Called at runtime after a device's information has been queried
+        # It is also a good advice to call it at startup in order to load the existing device to the DB.
+        # new will allow to make the differences between startup and ongoing
+
+        LOGGER.info("Device is ready: new=%s, device=%s NwkId: %s IEEE: %s signature=%s", new, device, device.nwk, device._ieee, device.get_signature()
+        if new and device.nwk != 0x0000 and len( device.get_signature()) > 0:     
+            domoCreateDevice( self, device._ieee, device.get_signature() )
+        for ep_id, endpoint in device.endpoints.items():
+            if ep_id == 0: # Ignore ZDO
+                continue
+            # You need to attach a listener to every cluster to receive events
+            for cluster in endpoint.in_clusters.values():
+                # The context listener passes its own object as the first argument
+                # to the callback
+                cluster.add_context_listener(self)
+
+
+    def device_left(self, device):
+        # When a device left the network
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -125,6 +125,69 @@ Run `pytest -lv`, which will show you a stack trace and all the local variables
...
+
+    def cluster_command(self, cluster, command_id, *args):
+        # ???? not sure on the parameters
+        pass
+
+    def attribute_updated(self, cluster, attribute_id, value):
+        # Call everytime and event is received. 
+        # Each object is linked to its parent (i.e. app > device > endpoint > cluster)
+        device = cluster.endpoint.device
+        pass
```
Adminiuga, line 187: these two are for "Cluster Event Listneres" --> events from the actual device/cluster. Everything above it is "ApplicationController" Listeners --> events happening on the network, like device leaving, joining etc.

CONTRIBUTING.md
```diff
@@ -125,6 +125,69 @@ Run `pytest -lv`, which will show you a stack trace and all the local variables
...
+
+
+    def device_removed(self, device):
+        # When a device is removed from the controler
+        pass
+
+
+    def cluster_command(self, cluster, command_id, *args):
+        # ???? not sure on the parameters
+        pass
```
Adminiuga, line 181: the signature is `def cluster_command(self, command_id, *args)` when added via `add_listener()` method. Signature `def cluster_command(self, cluster, command_id, *args)` is applicable when added via `add_context_listenere()` method. see https://github.com/zigpy/zigpy/blob/64e4f731d34e3ed0aa75cb1146cb79ea1314c91a/zigpy/util.py#L28-L49

Adminiuga: `command_id` is an `int` representing a "cluster specific command", for example `command_id == 0x0000` is the "Off" command if received from the OnOff cluster. The type of cluster: in_cluster (server cluster) or out_cluster (client cluster) may have different commands, so application needs to be aware if the command is coming from an input or output clusters. 
`*args` are going to be command specific. Look into https://github.com/zigpy/zigpy/blob/dev/zigpy/zcl/clusters definitions.

CONTRIBUTING.md
```diff
@@ -125,6 +125,69 @@ Run `pytest -lv`, which will show you a stack trace and all the local variables
...
+
+    def attribute_updated(self, cluster, attribute_id, value):
```
Adminiuga, line 183: Same as for `cluster_command` this signature is correct when added through `add_context_listener()` method. When added through `add_listener()` the `cluster` instance parameter is omitted.

CONTRIBUTING.md
pipiche38: I'm not sure what you expect from me. I'm trying to understand how to use the all zigpy library, so feel free to update the PR accordingly. I'm sure that I'm wrong in a lot of places.

CONTRIBUTING.md
pipiche38: I'm not sure what you expect from me. I'm trying to understand how to use the all zigpy library, so feel free to update the PR accordingly. I'm sure that I'm wrong in a lot of places.

pipiche38:
My purpose here is to create some User Guide on how to develop an app. 

So from my end, I imagine a kind of guide where we describe the few steps on how to create an apps.

I think that we should have a kind of workflow which describe from a UI command to the Zigbee radio what is the flow and vice-versa from a Zigbee radio event to the UI app what is the data flow.

For you @Adminiuga it is obvious, but for new comers this is not. and pointing out some codes, others apps and ZHA as such , I'm sorry this is quiet hard .

Gamester17:
@puddly & @dmulcahey Do you maybe have some suggestions or feedback on a guide for how-to get started with the zigpy library?

Gamester17:
I like to suggest also looking at the README.md from zsmartsystems/com.zsmartsystems.zigbee for inspiration:

https://github.com/zsmartsystems/com.zsmartsystems.zigbee/blob/master/README.md

PR #479, opened 2020-09-10
Enhance get_signature

pipiche38:
Makes get_signature method returning more device informations such as 
Model Id, 
Manufacturer Name/Code, 
ProfileID  
DeviceId

This will ease getting essential information on the application side.

PS/ I'm thinking to return battery level when the device has a battery component.

### DIFF
--- a/zigpy/device.py
+++ b/zigpy/device.py
@@ -338,7 +338,7 @@ class Device(zigpy.util.LocalLogMixin, zigpy.util.ListenableMixin):
         return self.endpoints[key]
 
     def get_signature(self):
-        #return the device signature by providing essential device information
+        # return the device signature by providing essential device information
         #    - Model Identifier ( Attribut 0x0005 of Basic Cluster 0x0000 )
         #    - Manufacturer Name ( Attribut 0x0004 of Basic Cluster 0x0000 )
         #    - Endpoint list
--- a/zigpy/device.py
+++ b/zigpy/device.py
@@ -346,25 +346,10 @@ class Device(zigpy.util.LocalLogMixin, zigpy.util.ListenableMixin):
         signature = {}
         if self._manufacturer is not None:
             signature["manufacturer_name"] = self.manufacturer
-        if self.manufacturer_id is not None:
-            signature["manufacture_code"] = self.manufacturer_id
         if self._model is not None:
             signature["model"] = self._model
         if self.node_desc.is_valid:
-            signature["mac_capability_flags"] = self.node_desc.mac_capability_flags
-            signature[
-                "descriptor_capability_field"
-            ] = self.node_desc.descriptor_capability_field
-            signature["server_mask"] = self.node_desc.server_mask
-            signature["byte1"] = self.node_desc.byte1
-            signature["byte2"] = self.node_desc.byte2
-            signature["maximum_buffer_size"] = self.node_desc.maximum_buffer_size
-            signature[
-                "maximum_incoming_transfer_size"
-            ] = self.node_desc.maximum_incoming_transfer_size
-            signature[
-                "maximum_outgoing_transfer_size"
-            ] = self.node_desc.maximum_outgoing_transfer_size
+            signature["node_descriptor"] = self._node_desc.as_dict()
 
         for endpoint_id, endpoint in self.endpoints.items():
             if endpoint_id == 0:  # ZDO
--- a/zigpy/device.py
+++ b/zigpy/device.py
@@ -348,7 +348,7 @@ class Device(zigpy.util.LocalLogMixin, zigpy.util.ListenableMixin):
                 - Cluster In
                 - Cluster Out
         """
-        signature = { "manufacturer_name":'', "manufacture_code":'', "model":'' }
+        signature = {"manufacturer_name": '', "manufacture_code": '', "model": ''}
 
         if self._manufacturer is not None:
             signature["manufacturer_name"] = self.manufacturer
--- a/zigpy/device.py
+++ b/zigpy/device.py
@@ -343,13 +343,18 @@ class Device(zigpy.util.LocalLogMixin, zigpy.util.ListenableMixin):
         #    - Manufacturer Name ( Attribut 0x0004 of Basic Cluster 0x0000 )
         #    - Endpoint list
         #        - Profile Id, Device Id, Cluster Out, Cluster In
-        signature = {"manufacturer_name": '', "manufacture_code": '', "model": ''}
+        signature = { }
         if self._manufacturer is not None:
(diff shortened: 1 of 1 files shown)


pipiche38:
Do not understand the last issue of travis !! Please let me know what is expected !

zigpy/device.py
```diff
@@ -338,13 +338,26 @@ def __getitem__(self, key):
         return self.endpoints[key]
 
     def get_signature(self):
-        signature = {}
+        # return the device signature by providing essential device information
+        #    - Model Identifier ( Attribut 0x0005 of Basic Cluster 0x0000 )
+        #    - Manufacturer Name ( Attribut 0x0004 of Basic Cluster 0x0000 )
+        #    - Endpoint list
+        #        - Profile Id, Device Id, Cluster Out, Cluster In
+        signature = {"manufacturer_name": '', "manufacture_code": '', "model": ''}
```
Adminiuga, line 346: Just keep it empty.

Adminiuga:
It is expecting tests cases covering both scenarios when when manuf/model is present and when not.

zigpy/device.py
```diff
@@ -338,13 +338,26 @@ def __getitem__(self, key):
...
+        if self._manufacturer is not None:
+            signature["manufacturer_name"] = self.manufacturer
+        if self.manufacturer_id is not None:
+            signature["manufacture_code"] = self.manufacturer_id
```
Adminiuga, line 350: Instead of manufacturer_id is could be better to store the node descriptor, as it gives you more information about the device

Adminiuga (commented):
Address the comments and it is good to go

puddly:
Make sure to follow the contributing guide's development environment installation instructions closely: https://github.com/zigpy/zigpy/blob/dev/CONTRIBUTING.md#setup

Right now the CI is failing because `black` would reformat your edits a little bit, which would be done automatically with the `pre-commit` hook. You can run those same checks locally and receive the same output by running `tox`.

zigpy/device.py
pipiche38: Good point, I'll see how to put those information too

zigpy/device.py
pipiche38: what do you mean ? I should keep signature empty, and then initialize each element ?

Adminiuga: yes. do `signature = {}` if model, manufacturer are `None` then they should be missing from the sig

zigpy/device.py
```diff
@@ -338,13 +338,31 @@ def __getitem__(self, key):
         return self.endpoints[key]
 
     def get_signature(self):
-        signature = {}
+        # return the device signature by providing essential device information
+        #    - Model Identifier ( Attribut 0x0005 of Basic Cluster 0x0000 )
+        #    - Manufacturer Name ( Attribut 0x0004 of Basic Cluster 0x0000 )
+        #    - Endpoint list
+        #        - Profile Id, Device Id, Cluster Out, Cluster In
+        signature = { }
```
Adminiuga, line 346:
```suggestion
        signature = {}
```

pipiche38:
> Why not just `signature['node_descriptor'] = str(self._node_desc)` ? That would fully allow you to reconstruct NodeDescriptor to the last detail if you have to

Agree, but doesn't it make sense to have to reconstruct later when we can get directly the information!

Adminiuga:
> > Why not just `signature['node_descriptor'] = str(self._node_desc)` ? That would fully allow you to reconstruct NodeDescriptor to the last detail if you have to
> 
> Agree, but doesn't it make sense to have to reconstruct later when we can get directly the information!

then you could do `signature['node_descriptor'] = self._node_desc.as_dict()`

pipiche38:
> > This will ease getting essential information on the application side.
> 
> can you expand a little bit on this? What essential information are you looking for? and what's the "application side? -- I assume it's not zigpy application, is it?
> What's the use case?

By application, i mean the application using the zigpy library. In my case the plugin for Domoticz. In that context, I need to create widgets in order to have the User interacting with the device.
List of clusters are going to give me what are the capabilities like Switch, Level Control, Window Covering ....
The Profile and Device Id can give me a bit more information on the device type
Finally the Model Name, will help me to handle specific devices and their features. I could have a cluster 0x0006 but get multi events and there is no real way to automatic discover it.

Then in normal time, when having to update a widget based on a attribute update I need also to correlate to the device information.

So to summarize, if the application (the plugin for instance) get all information in a dictionary like get_signature does this is good and there is no need to manipulate other means.

Adminiuga:
Hrm, Ok.
I mean you also have the list of clusters in `device.endpoints[endpoint_id].in_clusters`. Not much different then accessing the signature dict. 

It seems to me internally that signature is used only in quirks testing. Not sure if ZHA UI leverages the same method or not

zigpy/device.py
```diff
@@ -338,13 +338,27 @@ def __getitem__(self, key):
...
+        #    - Manufacturer Name ( Attribut 0x0004 of Basic Cluster 0x0000 )
+        #    - Endpoint list
+        #        - Profile Id, Device Id, Cluster Out, Cluster In
         signature = {}
+        if self._manufacturer is not None:
+            signature["manufacturer_name"] = self.manufacturer
+        if self._model is not None:
+            signature["model"] = self._model
+        if self.node_desc.is_valid:
+            signature["node_descriptor"] = self._node_desc.as_dict()
```
Adminiuga, line 352: My bad. it is supposed to be 
```suggestion
            signature["node_descriptor"] = self.node_desc.as_dict()
```

pipiche38:
> Hrm, Ok.
> I mean you also have the list of clusters in `device.endpoints[endpoint_id].in_clusters`. Not much different then accessing the signature dict.
> 
> It seems to me internally that signature is used only in quirks testing. Not sure if ZHA UI leverages the same method or not

This is may be where I come from and how I structured the access to ZiGate in the ZiGate plugin but I really appreciate to have only one call to get all infos.

PS/ Unfortunately for now, it is hard to test, as the zigpy-zigate master/dev doesn't work with zigpy/dev

Adminiuga:
> but I really appreciate to have only one call to get all infos.

Well, this is duplicating. To me, the `zigpy` is not really that high. and `get_signature()` feels more like a representation of the device, which is already there. if the actual app needs that data structured in one place, `zigpy.device.Device` instance was supposed to be that representation and  if the app needs another "presentation", then the app should do so upon  starting zigpy and iterating over all devices. 
And since we apparently already using it for ZHA, we'll keep it.

> fortunately for now, it is hard to test, as the zigpy-zigate master/dev doesn't work with zigpy/dev

What's broken?

pipiche38:
Thanks for approving the PR

PR #482, opened 2020-09-12
Refactor t.Struct fields

Adminiuga:
An unassigned annotated attribute is a field.
An assigned attribute of `StructField` class is a field.
An assigned and annotated attribute of `StructField` class is a field.
Anything else is not a field.


zigpy/types/struct.py
```diff
@@ -114,26 +114,9 @@ def fields(cls) -> typing.List["StructField"]:
         for name in list(annotations) + [v for v in variables if v not in annotations]:
             # It's a lot easier to debug when things break immediately instead of
```
puddly, line 115: This comment is now unnecessary but otherwise it looks good to me.

puddly (approved):

PR #483, opened 2020-09-13
Remove unused `Optional` annotations from `Struct` and cache `fields`

puddly:
This is now not supported (and was never actually used in the codebase):

```Python
class TestStruct(t.Struct):
    foo: t.Optional[t.uint8_t]
```

The following is also now invalid due to Python providing no way to get the order of annotated and unannotated attributes:

```Python
class TestStruct(t.Struct):
    foo: t.uint8_t
    bar = t.StructField(type=t.uint8_t)
    baz: t.uint8_t
```

Instead, you have to use `None` as a type annotation (maybe it'd be better to use `t.StructField`??):

```Python
class TestStruct(t.Struct):
    foo: t.uint8_t
    bar: None = t.StructField(type=t.uint8_t)
    baz: t.uint8_t
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,3 +71,5 @@ ENV/
 
 # Visual Studio Code
 .vscode
+
+.DS_Store
--- a/tests/test_struct.py
+++ b/tests/test_struct.py
@@ -1,5 +1,4 @@
 import enum
-import typing
 from unittest import mock
 
 import pytest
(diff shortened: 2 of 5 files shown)


puddly:
One feature that I'm unusure of is this:

```Python
class TestStruct(t.Struct):
    foo: t.uint8_t

s = TestStruct()
s.foo = -123

# Should this raise an error by converting the fields to their appropriate types?
s.as_dict() == {"foo": -123}
```

Adminiuga (approved):

PR #485, opened 2020-09-14
Keep list of neighbors for Coordinator and Routers

Adminiuga:
Scan Coordinator and Routers for neighbors on a regular interval and keep that information persistent


dmulcahey:
Nice!

PR #498, opened 2020-09-23
Prevent defective devices from going into OTA upgrade loops

puddly:
Mitigates the IKEA OTA issue in https://github.com/zigpy/zigpy/issues/497 by disallowing zigpy from responding to an image request from a single device with the exact same image file more than once.

### DIFF
--- a/tests/test_ota.py
+++ b/tests/test_ota.py
@@ -111,6 +111,29 @@ async def test_get_image_new(ota, image, key, image_with_version, monkeypatch):
     assert ota.async_event.call_count == 1
 
 
+async def test_has_performed_upgrade(ota, image_with_version):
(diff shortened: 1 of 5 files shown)


puddly:
This is still a WIP as I'd like to improve the tests a little and add a few more corner cases.

pipiche38:
I suggest that we have a way to disable such mitigation and for exemple:

1/ when doing a development based on the zigpy library, we might be interested to have such behaviour from a testing purposes.
2/ I don't think Zigbee suggest to prevent such behaviour, while I agree this is not a good one.

Adminiuga:
@puddly when it is ready, convert it to "ready for review"

PR #500, opened 2020-09-25
Reject invalid SiLabs upgrade images

puddly:
Another fix for #497. This prevents zigpy from even considering the corrupted IKEA OTA files. Only known OTA containers from Silicon Labs with an invalid structure are rejected so this will not affect other firmwares.

I'm hesitant to include a ton of OTA files into the zigpy `tests/` directory so I've run it locally on every single IKEA OTA file downloaded from the current stable and test feeds, along with every OTA file from an older stable feed, and then every file in https://github.com/Koenkk/zigbee-OTA. All OTA images are marked either as "UNKNOWN" or "VALID", with eight images from the currently buggy IKEA feed being marked as "INVALID" (confirmed by testing with Silicon Labs' own firmware tool).

If you have other images you can test every file in a folder recursively:

```Python
import pathlib

from zigpy.ota import OTAImage
from zigpy.ota.validators import validate_ota_image

for f in pathlib.Path("~/Projects/zigpy/ikea-firmwares").expanduser().glob('**/*'):
    if not f.is_file():
        continue

    if f.suffix == '.signed':
        ikea_container = f.read_bytes()

        ota_offset = int.from_bytes(ikea_container[16:20], "little")
        ota_size = int.from_bytes(ikea_container[20:24], "little")

        data = ikea_container[ota_offset:ota_offset + ota_size]
    else:
        data = f.read_bytes()

    try:
        image, _ = OTAImage.deserialize(data)
    except Exception:
        print(f'Not an OTA image: {f}')
        continue

    result = validate_ota_image(image)
    print(f'{result}: {f}')
```

### DIFF
--- a/tests/test_ota.py
+++ b/tests/test_ota.py
@@ -6,6 +6,7 @@ import zigpy.application
 import zigpy.ota
 import zigpy.ota.image
 import zigpy.ota.provider
+import zigpy.ota.validators
 
 from .async_mock import AsyncMock, MagicMock, patch, sentinel
 
@@ -42,8 +43,14 @@ def key():
 def ota():
     app = MagicMock(spec_set=zigpy.application.ControllerApplication)
     tradfri = MagicMock(spec_set=zigpy.ota.provider.Trådfri)
+    validate_ota_image = MagicMock(
+        spec_set=zigpy.ota.validators.validate_ota_image,
+        return_value=zigpy.ota.validators.ValidationResult.VALID,
+    )
+
     with patch("zigpy.ota.provider.Trådfri", tradfri):
-        return zigpy.ota.OTA(app)
+        with patch("zigpy.ota.validate_ota_image", validate_ota_image):
+            yield zigpy.ota.OTA(app)
 
 
 async def test_ota_initialize(ota):
@@ -109,6 +116,42 @@ async def test_get_image_new(ota, image, key, image_with_version, monkeypatch):
     assert res.header == newer.header
     assert res.subelements == newer.subelements
     assert ota.async_event.call_count == 1
+
+
+async def test_get_image_invalid(ota, image, image_with_version):
+    corrupted = image_with_version(image.version)
+
+    zigpy.ota.validate_ota_image.return_value = (
+        zigpy.ota.validators.ValidationResult.INVALID
+    )
+    ota.async_event = AsyncMock(return_value=[None, corrupted])
+
+    assert len(ota._image_cache) == 0
+    res = await ota.get_ota_image(MANUFACTURER_ID, IMAGE_TYPE)
+    assert len(ota._image_cache) == 0
+
+    assert res is None
+
+
+@pytest.mark.parametrize("v1", [0, 1])
+@pytest.mark.parametrize("v2", [0, 1])
+async def test_get_image_invalid_then_valid_versions(v1, v2, ota, image_with_version):
+    image = image_with_version(100 + v1)
+    image.header.header_string = b"\x12" * 32
+
+    corrupted = image_with_version(100 + v2)
+    corrupted.header.header_string = b"\x11" * 32
+
+    ota.async_event = AsyncMock(return_value=[corrupted, image])
+    zigpy.ota.validate_ota_image.side_effect = [
+        zigpy.ota.validators.ValidationResult.INVALID,
+        zigpy.ota.validators.ValidationResult.VALID,
+    ]
+
+    res = await ota.get_ota_image(MANUFACTURER_ID, IMAGE_TYPE)
+
+    # The valid image is always picked, even if the versions match
+    assert res.header.header_string == image.header.header_string
 
 
 def test_cached_image_expiration(image, monkeypatch):
--- a/tests/test_ota_validators.py
+++ b/tests/test_ota_validators.py
@@ -1,0 +1,163 @@
+import pytest
+
+from zigpy.ota import validators
+from zigpy.ota.image import ElementTagId, OTAImage, SubElement
+from zigpy.ota.validators import ValidationResult
+
+VALID_EBL_IMAGE = (
+    b"\x00\x00\x00\x8C" + b"\xAB" * 140 + b"\xFC\x04\x00\x00" + b"\xFF" * 44
+)
+VALID_GBL_IMAGE = b"\xEB\x17\xA6\x03\x00\x00\x00\x00"
+
+
+def create_subelement(tag_id, value):
+    return SubElement.deserialize(
+        tag_id.serialize() + len(value).to_bytes(4, "little") + value
+    )[0]
+
+
+def test_parse_silabs_ebl():
+    img = b"AA" + b"\x00\x04" + b"test" + b"\xFC\x04\x00\x00" + b"\xFF" * 52
+
+    assert list(validators.parse_silabs_ebl(img)) == [
+        (b"AA", b"test"),
+        (b"\xFC\x04", b""),
+    ]
+
+    with pytest.raises(AssertionError):
+        list(validators.parse_silabs_ebl(b""))
+
+    # Needs to be a multiple of 64 bytes
+    with pytest.raises(AssertionError):
+        list(validators.parse_silabs_ebl(img[:-1]))
+
+    with pytest.raises(AssertionError):
+        list(validators.parse_silabs_ebl(img + b"\xFF"))
+
+    # Bad length
+    with pytest.raises(AssertionError):
+        list(validators.parse_silabs_ebl(b"AA\xFF\xFF" + b"\xFF" * 60))
+
+    img2 = (
+        b"AA"
+        + b"\x00\x05"
+        + b"test1"
+        + b"BB"
+        + b"\x00\x04"
+        + b"test"
+        + b"\xFC\x04\x00\x00"
+        + b"\xFF" * 43
+    )
+
+    assert list(validators.parse_silabs_ebl(img2)) == [
+        (b"AA", b"test1"),
+        (b"BB", b"test"),
+        (b"\xFC\x04", b""),
+    ]
+
+
+def test_parse_silabs_gbl():
+    img = b"AAAA" + b"\x04\x00\x00\x00" + b"test"
+
+    assert list(validators.parse_silabs_gbl(img)) == [(b"AAAA", b"test")]
+
+    with pytest.raises(AssertionError):
+        list(validators.parse_silabs_gbl(b""))
+
+    with pytest.raises(AssertionError):
+        list(validators.parse_silabs_gbl(img[:-1]))
+
+    with pytest.raises(AssertionError):
+        list(validators.parse_silabs_gbl(img + b"\xFF"))
+
+    # Bad length
+    with pytest.raises(AssertionError):
+        list(validators.parse_silabs_gbl(b"AAAA\xFF\xFF\xFF\xFF"))
+
+    img2 = (
+        b"AAAA"
+        + b"\x05\x00\x00\x00"
+        + b"test1"
+        + b"BBBB"
+        + b"\x04\x00\x00\x00"
+        + b"test"
+    )
+
+    assert list(validators.parse_silabs_gbl(img2)) == [
+        (b"AAAA", b"test1"),
+        (b"BBBB", b"test"),
+    ]
+
+
+def test_validate_firmware():
+    assert validators.validate_firmware(VALID_EBL_IMAGE) == ValidationResult.VALID
+    assert (
+        validators.validate_firmware(VALID_EBL_IMAGE[:-1]) == ValidationResult.INVALID
+    )
+    assert (
+        validators.validate_firmware(VALID_EBL_IMAGE + b"\xFF")
+        == ValidationResult.INVALID
+    )
+
+    assert validators.validate_firmware(VALID_GBL_IMAGE) == ValidationResult.VALID
+    assert (
+        validators.validate_firmware(VALID_GBL_IMAGE[:-1]) == ValidationResult.INVALID
+    )
+
+    assert validators.validate_firmware(b"UNKNOWN") == ValidationResult.UNKNOWN
+
+
+def test_validate_ota_image_simple_valid():
+    image = OTAImage()
+    image.subelements = [
+        create_subelement(ElementTagId.UPGRADE_IMAGE, VALID_EBL_IMAGE),
+    ]
+
+    assert validators.validate_ota_image(image) == ValidationResult.VALID
+
+
+def test_validate_ota_image_complex_valid():
+    image = OTAImage()
+    image.subelements = [
+        create_subelement(ElementTagId.ECDSA_SIGNATURE, b"asd"),
+        create_subelement(ElementTagId.UPGRADE_IMAGE, VALID_EBL_IMAGE),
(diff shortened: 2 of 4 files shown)


puddly:
This will only detect corruption that affects the structure of the EBL/GBL firmware images. It is still possible for an image to be corrupted while still being parsable, which will require calculating the embedded CRC-32 checksum as well. PR #498 will also help mitigate the problem as validating all possible firmware image types is a little out-of-scope for zigpy.

Adminiuga:
do we know where the checksum is so we could calculate it? 
The bad ikea images, do those have incorrect structure and/or crc?

puddly:
There are multiple checksums if I remember correctly, one for the metadata and one for the other reset (or maybe it's for the whole file?). My original plan was to compute the checksum but the broken IKEA images can't even be parsed (i.e. Simplicity Commander throws the same error) so the checksumming won't tell us anything new.

puddly:
There are eight broken images, not just the two that we've seen so far, so it'd be useful to confirm that one of these other devices is indeed affected: https://github.com/zigpy/zigpy/issues/497#issuecomment-699016025

puddly:
Still no false negatives with CRC-32 checksum verification for both EBL and GBL images.

I don't have any encrypted images to test with but I believe they just use a few new tags to encapsulate the regular erase-and-write tags while retaining the plaintext header and checksum footer.

MattWestb:
@puddly From the release [1.14.0](https://ww8.ikea.com/ikeahomesmart/releasenotes/releasenotes.html) its 8 updated firmware released and i think its all the device you have finding (plus the GW firmware is on the list).
The crazy thing its that the versions is not updated only one new OTA file that is broken with "Improvements During FW Over The Air Upgrade" and IKEA is always updating the versions in the metadata then releasing new firmware.
If not the GW firmware was updated im my GW (its saying its running on the last 1.11.47) i wold saying all its hacked both the release note and the "new"  firmware but not so good made then its not working for the signed files is being rejected by the NCP security check.
The relesnotes was not possible to access for one week ago with "access denied but the cert is updated 2020.09.22 and looks walide for Inter IKEA Systems B.V (can being reworked after one hacking ?!?!?).
Very likely the GWs NCP is refusing false FW then the CRC and the signing is being falce and i hope its the same with the "host firmware" in the GW or they have getting one large problem in their ecosystem being real hacked.
Was downloading the NCP and the "Host" firmware if some is interesting taking on lock on it (the HTTPS was down so its from the HTPP json). [TF_GW_FW.zip](https://github.com/zigpy/zigpy/files/5286572/TF_GW_FW.zip)


Perhaps i should sending one e-mail to paul.arsson@ikea.com (technical contact in the FCC certificate) and asking in swedish if the 8 firmware is genuine or something else.

In the end great work done for solving the problem !!!!

puddly:
@MattWestb thanks for the gateway info, I found a few embedded 4096-bit RSA public keys that are interesting. [A thesis by Mikael Szreder](https://liu.diva-portal.org/smash/get/diva2:1362068/FULLTEXT01.pdf) that was published last year and outlined the reverse engineered format of the ikea `.signed` container. Below is a complete parser for the format:

```Python
# Algorithm details from: https://liu.diva-portal.org/smash/get/diva2:1362068/FULLTEXT01.pdf

import sys
import pathlib
import hashlib

from cryptography.exceptions import InvalidSignature
from cryptography.hazmat.backends import default_backend
from cryptography.hazmat.primitives import serialization, hashes
from cryptography.hazmat.primitives.asymmetric.padding import PKCS1v15

PUBLIC_KEYS = {
    "key 1": serialization.load_pem_public_key(b"""
-----BEGIN PUBLIC KEY-----
MIICIjANBgkqhkiG9w0BAQEFAAOCAg8AMIICCgKCAgEAwH8PFOYPFH8Rx/2KHnoi
1liUoYUQH68nFc1zLwaQ+LA7J8bQgZP9VNkTiV1wfU3K9l0jTgzi1lffjEeB7OVt
X7E3dMh0NrAa12YfGCP1M7ziaxw/L+jBIUkHKpSgI/1KwArBmAz/LXAKESE1mC/I
hiuJ0Pd418PlxF4wJJogFhczYsvlcRJ6VNriQa3Ycf0TehMgJf7tpWi7ixaOk2Aa
A9jwFfPPcBX+RG6+ecGrO805aJ7KSGs06qxVygDCstAff1PaUP03OlX6Y7ZsKGXi
39XDc8rGx8nNDcUU2fEfijwT1YxPVaC+F/H9dZPDfJgYsPx8ien67Dqw4+MIFR91
8vRaD3EfHq2gdEJyQRX9sqCaBGXe4mhu2bhrUQCWyijc4oXO3Po7g1BXE6Oc95tm
jlZYYfGdfXADpZJQ7V16wNlLP/5abkTsahvVbXyUl67VakxBmhzO3ApqjQ4AiTPU
8S0KCM32U7+nDsaKPMocE7FijjEQhrvVGEANH8gwqBCOlEgtbLBOmVkwMvRhyIIq
L11UOlTTx4nEu1OucD3MxQmG9v45wpBUGhK0L+HHrLpOvG+ZRuCi73Mw9kFLbEaB
gW5tvykL7iNhNfWuc0CMMO/IdLVudyDri1HiQdp3f1BIF/sCcLHnax8Z+4qTxXPp
LZsuj6t0G1EHavLHL7GacQkCAwEAAQ==
-----END PUBLIC KEY-----
""", backend=default_backend()),
    "key 2": serialization.load_pem_public_key(b"""
-----BEGIN PUBLIC KEY-----
MIICIjANBgkqhkiG9w0BAQEFAAOCAg8AMIICCgKCAgEAxjnwYE2XJHDce9XYY/A6
l98xcQ247RCe/j/of0CHujCRp6wZNYOJMSIVFe8xwhMXhm3pJ07H6sEKi0G5xfE0
gsIptD2aQWo+ZVvHT1DdXUS2XStuweOUUDbLSCEpPIZ0T5dMU/sq+nhCcOPLZ/Kj
PCk2TCPtuWsAlOAVkHEJdbUrSoANbvb6teZIT+mr09yuJyKsQSbDU4HgXl6csqAt
bBbxzZ/e/zq3GGn4vJxQ0hOcVwjvEYXCAafMDFmRce/i9Kc6zju+aUoS8z03n57F
ZyRBKa4oY3wTENgxPEBycZhugyDlwYOv5oiqxRf6chDKbCzT521jkhP8SRXfCH/g
BxaJQlGXpGpOMwiXQFmoU4tH4id43dpNItWp0ra0J4wDxiwxm60QPsS7Zvgpm32g
lRoN14Bpaqjki4xpM054MBZHWSG4JH27iGfDU2nA33OgA6FYmLv30rn4vzV5cvJ0
Cz048+NanQRaFIfXSjcYLtXJZTJpU6dSNzI5EBGpkp1pruLIcrRCTXaYg/yGqKgB
4SFe7Wg+x7+zIr2Y6r8NMmnX2VRSPgFD7IyPw5xxYpLZd9Ozd7ELQ4rUQkHqigne
tZltzf5f7FMEBUo/Qjoo0efTnz7BMcFLEMLZ0c/fPcNpcj5R4mWDTjX+hmXTHskp
1y+jpv9XO7KfPkKD33GDIjsCAwEAAQ==
-----END PUBLIC KEY-----
""", backend=default_backend())
}

f = pathlib.Path(sys.argv[1])

data = f.read_bytes()

magic = data[0:4]
unknown1 = int.from_bytes(data[4:8], "little")
unknown2 = int.from_bytes(data[8:12], "little")
relative_offset_until_ota = int.from_bytes(data[12:16], "little")
ota_offset = int.from_bytes(data[16:20], "little")
ota_size = int.from_bytes(data[20:24], "little")
footer_offset = int.from_bytes(data[24:28], "little")

unknown3 = int.from_bytes(data[28:32], "little")
block_size = int.from_bytes(data[32:36], "little")
num_block_hashes = int.from_bytes(data[36:40], "little")

assert magic == b"NGIS"
assert unknown1 == 12
assert unknown2 == 2010
assert unknown3 == 11
assert relative_offset_until_ota + 12 == ota_offset
assert 40 + 32 * num_block_hashes == ota_offset
assert block_size * num_block_hashes >= ota_size

for block_num in range(num_block_hashes):
    offset = ota_offset + block_size * block_num
    size = block_size - max(0, offset + block_size - (ota_offset + ota_size))

    block = data[offset:offset + size]
    checksum = hashlib.sha256(block).digest()

    assert data[40 + 32 * block_num:40 + 32 * (block_num + 1)] == checksum


assert len(data) - (ota_offset + ota_size) == 512
header = data[0:ota_offset]
header_signature = data[-512:]

for name, pubkey in PUBLIC_KEYS.items():
    try:
        pubkey.verify(
            signature=header_signature,
            data=header,
            padding=PKCS1v15(),
            algorithm=hashes.SHA256(),
        )

        print(f"Image was signed with {name}")
        break
    except InvalidSignature:
        pass
else:
    print(f"Image has an invalid signature!")
```

Every single IKEA image (including the broken ones) have the correct block checksums and were signed by the correct private key so the images must have been corrupted before being signed, making this script useless for integrity verification😄. Interestingly, there are *two* signing keys: "key 1" was used to sign all of the firmwares before the last update, and the current "bad" ones; "key 2" is being used for the new testing firmwares. I wonder why?

MattWestb:
Glad to helping you in some way !
If IKEA is updating there signing keys in the deveces thru OTA / GBL files they must having one container with patch for the new token in the hardware config / lockbits region but you have only seeing the "normal" firmware container in the OTA (new) files so its not being possible for the device using the "new" OTA files then the signing it wrong.
Can IKEA rolling out "new" devices firmware with new signing keys ? Sounds stupin doing that with the same mfg and device ID. Soul being better changing the device ID for the new ones and leaving the old one untouched / with old keys.
One thing can being that the corruption have being made before the its being signed, Silabs is writing warnings for that and recomeding making S37 / GBL files with extra CRC for protecting the firmware and having the signing PC off line then not being used for signing (so not possible stealing the master key).

Still not making sense for my !!!!

puddly:
The gateway "unwraps" this container though and sends the normal Zigbee OTA files to the devices, leaving the CRC-32 within the GBL/EBL files as the only thing they can verify, correct? Or is there another signature within the written firmware itself that's verified by the bootloader after it written to flash?

Adminiuga:
Are there subelements with crc and signature?
Btw, we are not checking ota crc in zigpy, right? I wonder if those are correct

MattWestb:
Was i have reading from the Gecko OTA (applikation) bootloader the devices is storing the OTA file on the external flash and if the CRC is OK its rebooting to the bootloader that is doing the flashing of the internal flash and the erasing the external flash and rebooting the NCP. (that sis for serie one with external flash for OTA storages). Then signing / encrypting the OTA files its still being used in the OTA module and bootloader and its its not OK the device is not flashing the firmware and erasing the OTA storage. The transport to the flash storage is transparent and the CRC / Signing / encrypting is first doing then the file is saved. The GW is only one transport and the device is making all security check and if its Ok its flashing.
If you "unwarping" the OTA file the device is not flashing it !!
But You can flashing it with SWD at 0x4000 (if a normal OTA file) and its its updated.
Look on the duped file for the internal flash and comparing it with the OUA file i was attaching before and you see that is the same as the device have flashing it self.

MattWestb:
The GW can being the flashing is made thru serial download from the host then the NCP have no external flash but still must being one GLB file and signed if the NCP have it installed or its not flashing it (Sonoff Zigbee Bridge is doing that and the factory have giving us signed firmware for manually updating the NCP firmware to normal EZSP).

puddly:
@Adminiuga I've not encountered a single image image with more than just the `UPGRADE_IMAGE` subelement, unfortunately. Including a ECDSA signature and the associated certificate (how would we verify the cert?) within the Zigbee OTA image would be a really useful way to check image integrity without relying on parsing vendor-specific formats.

The CRC-32 checksum is in the very last tag of the EBL/GBL image so we can't even get them when the images themselves are structurally broken. But CRC-32 verification is good to have nonetheless and very little extra code.

puddly:
By the way, do you think we can replace `OTAImage.MAXIMUM_DATA_SIZE` with the coordinator's `NodeDescriptor.maximum_outgoing_transfer_size` or something? I'll have to test it.

MattWestb:
@puddly The "IKEA container" is one raw file that is 1 x 1 that the flashed at 0x4000 and then the OTA file is not encrypted. Only CRC and signing its outside the "IKEA container.
If trying doing one CRC check and its being false or cant not being done (the end of the file is missing) = Bad OTA file.

MattWestb:
The upper is the dumped internal flash of one LED1836G9 at 0x4000 is the "app image" starting. (at 0x0 is the first bootloader and the second stage bootloader is at 0x800).
The lower is the signed OTA for LED1836G9 with the same version and after the signing and firmware image is starting and its inplain (raw) as you can see until the end of the image.

OTACRC1

MattWestb:
If Zigpy implanting CRC check then its only downloading the file and aborting the OTA if its not OK and not need sending it to the device that doing CRC check and throwing it one hour later (router lesser and end device much longer and eating the battery). = win win (minus coding).

puddly:
Definitely. Two of the eight affected devices are battery-powered ("onoff-shortcut-control" and "dimmer").

I think the current OTA code will actually perform an invalid upgrade if one of these slow devices is performing a doomed-to-fail OTA upgrade when zigpy and HA is restarted, as we are obtaining the OTA image with `(manufacturer_id, image_type)` as the key, not `(manufacturer_id, image_type, file_version)`. It will hopefully fail to re-fetch the "bad" image and then fail with `Status.ABORT` but there is still a tiny chance that a new, valid OTA upgrade will take its place and zigpy will continue sending blocks from the wrong image.

MattWestb:
This is the [official list](https://ww8.ikea.com/ikeahomesmart/releasenotes/releasenotes.html) of the "updated" firmwares:
````
Release Version: 1.14.0
10th September 2020
New features and changes in Gateway: 1.11.47
General Bug Fix and Performance Improvement
New features and changes in Accessories:
FLOALT NA (V-2.0.029).
TRÅDFRI Control Outlet (V-2.0.024).
TRÅDFRI WS2.0 Opal(E27,E14,GU10) (v-2.0.029).
GUNNARP (IP44) EU (V-2.3.050).
SYMFONISK Sound Remote (V-2.1.024).
LEPTITER,TRÅDFRI E14 Chandelier, WS2.0 Clear E27/E26, GUNNARP NA (v-2.3.050).
TRÅDFRI WW2.0 2700K E27, 2700K GU10, 2200K Deco Bulb (V-2.3.50).
TRÅDFRI Open/Close Remote,TRÅDFRI On/Off Button (v-2.2.010).
Improvements During FW Over The Air Upgrade.
````
I have not checking all but that i have done some that i have and the version not updated = the same as the "old" firmware plus broken OTA file.
I think its all firmwares you have in your defect list.

MattWestb:
If the device is receiving the complete OTA file and storing it in the external flash its first checking the CRC and if its OK its doing the flash of the OTA and if its not its throwing the OTA file and aborting the update.
The OTA file in GLB format have 3 CRC checks [UG103.6: Bootloader Fundamentals](https://www.silabs.com/documents/public/user-guides/ug103-06-fundamentals-bootloading.pdf) 5.1.2 Data Verification. 
Tak one fast look on the paper its describing more or less all things easily.

PR #504, opened 2020-09-28
Limit the length of logged `image` objects during firmware validation

puddly:
https://github.com/zigpy/zigpy/pull/500 logs `OTAImage` validation statuses with the `DEBUG` log level and invalid images with the `ERROR` log level. The inclusion of `OTAImage` objects in these log message will cause them to be logged in their entirety, which will include over 200kb of image data into logs during normal use when invalid images are encountered.

This change limits the length of the logged objects to at most 500 bytes, which is more than enough to capture the full OTA header and the header of the encapsulated firmware.

### DIFF
--- a/zigpy/ota/__init__.py
+++ b/zigpy/ota/__init__.py
@@ -53,10 +53,10 @@ class OTA(zigpy.util.ListenableMixin):
         self._not_initialized = True
         self._listeners = {}
         ota_config = app.config[CONF_OTA]
-        if ota_config[CONF_OTA_IKEA]:
-            self.add_listener(zigpy.ota.provider.Trådfri())
         if ota_config[CONF_OTA_DIR]:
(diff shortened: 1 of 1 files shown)


Adminiuga:
I'm not positive we should print the image repr either in debug or error. 
Imo we should just print the image header or some unique Id of the image.

puddly:
Printing just the header works for me. We could also keep track of the image's source and print that instead. I would just prefer not to have "image error" messages without any context whatsoever.

Adminiuga:
I agree with that, should include some context, like couldn't deserialize image or failed crc check etc

puddly:
I'll add more verbose errors then to the validation module.

Adminiuga:
DM me when it's ready for final review, so I don't forget about it

PR #508, opened 2020-09-30
Allow configuration of IKEA update URL

Adminiuga:
Allow configuration of IKEA update URL via `ikea_update_url` configuration parameter.
For example in Home Assistant core `configuration.yaml` add:

```yaml
zha:
  zigpy_config:
    ota:
      ikea_provider: true
      ikea_update_url: http://fw.test.ota.homesmart.ikea.net/feed/version_info.json
```

### DIFF
--- a/zigpy/config/__init__.py
+++ b/zigpy/config/__init__.py
@@ -36,6 +36,7 @@ CONF_NWK_UPDATE_ID = "update_id"
 CONF_OTA = "ota"
 CONF_OTA_DIR = "otau_directory"
 CONF_OTA_IKEA = "ikea_provider"
+CONF_OTA_IKEA_URL = "ikea_update_url"
 CONF_OTA_LEDVANCE = "ledvance_provider"
 CONF_TOPO_SCAN_PERIOD = "topology_scan_period"
 
@@ -72,6 +73,7 @@ SCHEMA_NETWORK = vol.Schema(
 SCHEMA_OTA = {
     vol.Optional(CONF_OTA_DIR, default=CONF_OTA_OTAU_DIR_DEFAULT): vol.Any(None, str),
     vol.Optional(CONF_OTA_IKEA, default=CONF_OTA_IKEA_DEFAULT): cv_boolean,
+    vol.Optional(CONF_OTA_IKEA_URL): vol.Url(),
     vol.Optional(CONF_OTA_LEDVANCE, default=CONF_OTA_LEDVANCE_DEFAULT): cv_boolean,
 }
 
--- a/zigpy/ota/provider.py
+++ b/zigpy/ota/provider.py
@@ -11,7 +11,7 @@ from typing import Dict, Optional
 import aiohttp
 import attr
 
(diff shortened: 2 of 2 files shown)


MattWestb:
Thanks cod master !!!

I was downgrading oncontroll outlet in deCONZ to 10005777-4.1-TRADFRI-control-outlet-2.0.024.ota.ota and then deleting it in deCONZ and pairing it in ZHA.

First I was putting the same OTA file in the /config/zigpy_ota and restarting ZHA and it was not updating the plug.
Then adding the normal  "ikea_provider: true" and its downloading but throwing the invalid OTA file and not sending it to the device.
Then putting one 2.0.024 OTA file from deCONZ in the ots folder and restating ZHA and repower the plug. Its starts updating and I disconnecting it for more testing.
Deleting the 2.0.024 nd adding    ikea_update_url: http://fw.test.ota.homesmart.ikea.net/feed/version_info.json

Restarting ZHA and repower the plug and its starting the update OK.


Great work done !!!

Only missing possibility forcing loafing of one older version of OTA files theru folder load and config parameter like:

````
zha:
  zigpy_config:
    ota:
      override_with_local_version: true
````

puddly:
"`override_with_local_version: true`" should be the default behavior as of https://github.com/zigpy/zigpy/pull/505.

Are you 100% sure you have the correct OTA file for the device? The broken image from the IKEA website will not even be considered so if the local one isn't being picked up it may be a problem with the file (or the file-based loader is having problems unwrapping the IKEA container).

MattWestb:
I have old OTA files that was downloaded in deCONZ and is working (and the test json is also working).
For the same version (server / local) its working, i have trying with the same version and the ZHA is starting sending the local file to the device.

My proposal is that having one older (lower version) local OTA file and sending it to the device and doing one "downgrade" as its possible in deCONZ. 

Then having the 2.0.022 in the local OTA folder the device is getting the 2.0.024 from the test server = good for "normal" users but not for advance testing.
Also if the device is on 2.0.024 and the local is 2.0.022 its not getting the "downgrade" OTA.

Perhaps i should caling the config parameter "force_version_downgrade" or "override_device_version".

Its only one proposal and if i really need doing it I can doing it in deCONZ (that have losing all devices lats week and Bellows have getting all of them).

MattWestb:
@walthowd  I was putting the `ikea_update_url:` in the wiki but was not changing the text so much.
Can you taking one look so i did not destroying you work !!!

walthowd:
Yup, looks great! What we really need now is someone to host a IKEA update URL somewhere that only has *working* firmware in it!

MattWestb:
For the moment is it not necessary then only IKEA GW and the new scene shortcut button (no update only new device) have being added from the last good ones in the test URL and the all other is at production standard. I finding it interesting wath is happening little later if IKEA is releasing one new GW and locking all updates to their end devices or is fixing bad data setting in the devices (with patch commands for the user data) or if its only corrupted / damaged files on the server.

Great working done !!!

PR #515, opened 2020-10-05
Fix attributes being lost when an existing device is re-added

abmantis:
INSERT OR REPLACE deletes and inserts when the record already exists.
Since table constraints have ON DELETE CASCADE, the attributes (and probably endpoints and so on) were being deleted when an existing device is saved. 
After rebooting HA, no quirk would match since the device would have not manufacturer attr.

Should we also change the other INSERT OR REPLACE commands?

### DIFF
--- a/tests/test_appdb.py
+++ b/tests/test_appdb.py
@@ -54,7 +54,9 @@ def make_ieee(init=0):
 
 
 class FakeCustomDevice(CustomDevice):
-    pass
+    replacement = {
+        "endpoints": {1: {"input_clusters": [0, 1, 3], "output_clusters": [6]}}
+    }
 
 
 def mock_dev_init(status: Status):
@@ -70,7 +72,7 @@ def mock_dev_init(status: Status):
 
 def fake_get_device(device):
     if device.endpoints.get(1) is not None and device[1].profile_id == 65535:
-        return FakeCustomDevice(device.application, make_ieee(1), 199, device)
+        return FakeCustomDevice(device.application, device.ieee, device.nwk, device)
     return device
 
 
@@ -483,3 +485,48 @@ async def test_neighbors(tmpdir):
     assert dev_2.neighbors[0].device is dev_1
     assert dev_2.neighbors[1].device is None
     assert dev_2.neighbors[1].neighbor.ieee == ieee_3
+
+
+@patch(
+    "zigpy.device.Device.schedule_initialize", new=mock_dev_init(Status.ENDPOINTS_INIT)
+)
+async def test_device_rejoin(tmpdir):
+    db = os.path.join(str(tmpdir), "test.db")
+    app = await make_app(db)
+    ieee = make_ieee()
+    nwk = 199
+    app.handle_join(nwk, ieee, 0)
+
+    dev = app.get_device(ieee)
+    ep = dev.add_endpoint(1)
+    ep.profile_id = 65535
+    ep.device_type = profiles.zha.DeviceType.PUMP
+    clus = ep.add_input_cluster(0)
+    ep.add_output_cluster(1)
+    app.device_initialized(dev)
+    clus._update_attribute(4, "Custom")
+    clus._update_attribute(5, "Model")
+
+    # Everything should've been saved - check that it re-loads
+    with patch("zigpy.quirks.get_device", fake_get_device):
+        app2 = await make_app(db)
+    dev = app2.get_device(ieee)
+    assert dev.nwk == nwk
+    assert dev.endpoints[1].device_type == profiles.zha.DeviceType.PUMP
+    assert dev.endpoints[1].in_clusters[0]._attr_cache[4] == "Custom"
+    assert dev.endpoints[1].in_clusters[0]._attr_cache[5] == "Model"
+    assert dev.endpoints[1].manufacturer == "Custom"
+    assert dev.endpoints[1].model == "Model"
+
+    # device rejoins
+    dev.nwk = nwk + 1
+    with patch("zigpy.quirks.get_device", fake_get_device):
+        app2.device_initialized(dev)
+
+    app3 = await make_app(db)
+    dev = app3.get_device(ieee)
+    assert dev.nwk == nwk + 1
+    assert dev.endpoints[1].device_type == profiles.zha.DeviceType.PUMP
+    assert 0 in dev.endpoints[1].in_clusters
+    assert dev.endpoints[1].manufacturer == "Custom"
+    assert dev.endpoints[1].model == "Model"
--- a/zigpy/appdb.py
+++ b/zigpy/appdb.py
@@ -263,8 +263,14 @@ class PersistingListener:
                 device.status,
             )
             return
-        q = "INSERT OR REPLACE INTO devices (ieee, nwk, status) VALUES (?, ?, ?)"
-        self.execute(q, (device.ieee, device.nwk, device.status))
+
+        try:
+            q = "INSERT INTO devices (ieee, nwk, status) VALUES (?, ?, ?)"
+            self.execute(q, (device.ieee, device.nwk, device.status))
+        except sqlite3.IntegrityError:
+            LOGGER.debug("Device %s already exists. Updating it.", device.ieee)
+            q = "UPDATE devices SET nwk=?, status=? WHERE ieee=?"
+            self.execute(q, (device.nwk, device.status, device.ieee))
         self._save_node_descriptor(device)
         if isinstance(device, zigpy.quirks.CustomDevice):
             self._db.commit()
--- a/zigpy/appdb.py
+++ b/zigpy/appdb.py
@@ -263,7 +263,6 @@ class PersistingListener:
                 device.status,
             )
             return
-
         try:
             q = "INSERT INTO devices (ieee, nwk, status) VALUES (?, ?, ?)"
             self.execute(q, (device.ieee, device.nwk, device.status))
@@ -271,6 +270,7 @@ class PersistingListener:
             LOGGER.debug("Device %s already exists. Updating it.", device.ieee)
             q = "UPDATE devices SET nwk=?, status=? WHERE ieee=?"
             self.execute(q, (device.nwk, device.status, device.ieee))
+
         self._save_node_descriptor(device)
         if isinstance(device, zigpy.quirks.CustomDevice):
             self._db.commit()


Adminiuga:
Ha, that would explain the behavior. Since you were able to reproduce it, does it fix the issue?

abmantis:
@Adminiuga yep. At least with the IKEA motion sensor, with which I was able to reproduce it 100%, it is now fixed.
Tested both in my dev (conbee) and production (elelabs) envs.

puddly:
Tests pass for me locally with this change so I'm not sure what Travis CI is complaining about.

I think I see the specific circumstances that are required for this bug to occur:

 1. The first `Device._initialize` call has triggers multiple calls to `Endpoint.initialize()`. One of these must fail, which causes the following codepath to be followed:
 ```Python
 if any(ep_failed_init):
     self.initializing = False
     self.application.listener_event("device_init_failure", self)
     await self.application.remove(self.ieee)
     return
 ```

 2. Because `device.initializing` is set to `False` and `device.application.remove` is a coroutine, this specific function can "block" for many seconds, permitting the device to rejoin the network with a *new* NWK and for `device.schedule_initialize` to happen again.

3. Depending on the completion order, one of two things can happen:

 a. `device.application.remove` eventually completes before the second `Device.initialize` call completes and then calls `application.listener_event("device_removed", dev)`, which may remove the device from the database.
 b. The second `Device.initialize` call runs first (which I think is happening here) and the `INSERT OR REPLACE` bug happens.

As another potential fix, try reordering the snippet I copy/pasted from `device.py` so that `self.initializing = False` happens after `await self.application.remove(self.ieee)`.

Adminiuga:
I finally could reproduce this long eluding issue, following the @abmantis' steps to reproduce.

> Because device.initializing is set to False and device.application.remove is a coroutine, this specific function can take many seconds, permitting the device to rejoin the network with a new NWK and for device.schedule_initialize to happen again (because device.initializing is False).

@puddly while the "remove" could take a while and potentially device can rejoin with the new NWK, per se it is not the root cause, as I did the following:
1. new instance, no devices
2. joined IKEA button
3. reset the IKEA button, without removing it, which generated the "device_left" event
4. restarted HA instance
5. after restart, zigbee.db still had correct information about button and there were no any pending coroutines, because HA was just restarted
6. re-join IKEA button successfully and now it is a time bomb, because it all endpoints and clusters were removed from DB, but in the `ControllerApplication.devices` the device instance is still correct. It is not until the next restart the problem manifests itself

abmantis:
@puddly I don't think that's how it happened, as I haven't seen any endpoint failure.
appdb._save_device() is always called at the end of device init (whether it is a new device or existing one). When that is called with an existing device, it will always REPLACE the record, which deletes records on other tables.
I think some devices will update the manuf + model attribute after pairing, that's why you don't always see the bug. Also, it only happens for quirky devices.

My steps that always work on the Ikea motion:
1 - no devices paired
2 - start pairing on HA
3 - pair motion sensor
4 - wait for sensor to show on the UI
5 - start pairing on HA again (just refresh the page)
6 - reset the sensor so it is re-paired

Adminiuga:
but why tox is failing? is it because of bionic?

puddly:
Works on 20.04 for me. Can you re-run the job with Travis?

Adminiuga:
It works for you on travis? I've tried re-running, same result. right now travis is on bionic

puddly:
My bad, I can't keep track of all these codename and version numbers. I tested it locally.

I wonder why this is showing up:

```
WARNING  zigpy.util:util.py:49 Error calling listener.raw_device_initialized: near "ON": syntax error
```

Is Travis's Python using an older version of SQLite without this feature? Having version-dependent SQLite code may be a problem, especially when 18.04 is an LTS release and isn't going away any time soon.

puddly:
Yeah, that's the issue:

> UPSERT is a special syntax addition to INSERT that causes the INSERT to behave as an UPDATE or a no-op if the INSERT would violate a uniqueness constraint. UPSERT is not standard SQL. UPSERT in SQLite follows the syntax established by PostgreSQL. UPSERT syntax was added to SQLite with version 3.24.0 (2018-06-04).
> 
> An UPSERT is an ordinary INSERT statement that is followed by the special ON CONFLICT clause shown above. 

Ubuntu 18.04:
```console
$ python3 -c "import sqlite3; print(sqlite3.sqlite_version)"
3.22.0
```

Ubuntu 20.04:
```console
$ python3 -c "import sqlite3; print(sqlite3.sqlite_version)"
3.31.1
```

abmantis:
Ubuntu 20.04 is also an LTS. How is HA policy on this kind of thing?

puddly:
Is there any way to recreate this feature by using a transaction containing a `SELECT` followed by an `UPDATE` or something?

Adminiuga:
```
# docker exec -ti homeassistant /bin/bash
bash-5.0# python3 -c "import sqlite3; print(sqlite3.sqlite_version)"
3.32.1
bash-5.0# 
```

Hass.os 0.116.b5

Adminiuga:
this probably gonna blow a few setups not on hass os. But even debian buster has 
```
adminiuga@debba:~$ python3 -c "import sqlite3; print(sqlite3.sqlite_version)"
3.27.2
```

Adminiuga:
so it is not the sqlite module python version, but the actual sqlite version what matters, doesn't it?

puddly:
Yes. `sqlite3.version` is the version of the Python module and `sqlite3.sqlite_version` is the SQLite version: https://docs.python.org/3/library/sqlite3.html#sqlite3.sqlite_version

Adminiuga:
i'm on the fence whether to use try/except UNIQUE constraint and do update in the exception handling?
or make it a breaking change???

puddly:
I'm trying to write a unit test to test out some fixes but I still can't get the database into the inconsistent state:

```Python

async def test_join_leave(tmpdir):
    db = os.path.join(str(tmpdir), "test.db")
    app1 = await make_app(db)

    ieee = make_ieee()

    device = app1.add_device(ieee, 0x1234)
    device.status = Status.ENDPOINTS_INIT
    device.initializing = False
    device.node_desc = zdo_t.NodeDescriptor(0, 1, 2, 3, 4, 5, 6, 7, 8)

    ep = device.add_endpoint(1)
    ep.profile_id = 260
    ep.device_type = profiles.zha.DeviceType.PUMP

    cluster = ep.add_input_cluster(0)
    cluster._update_attribute(0, 99)
    cluster._update_attribute(4, b"Custom")
    cluster._update_attribute(5, b"Model")

    async def mockleave(*args, **kwargs):
        return [0]

    device.zdo.leave = mockleave

    app1.handle_join(device.nwk, device.ieee, 0x000)
    app1.device_initialized(device)
    app1.handle_leave(device.nwk, device.ieee)

    await app1.shutdown()
    del app1

    app2 = await make_app(db)
    app2.handle_join(device.nwk, device.ieee, 0x000)
    app2.device_initialized(device)
    device = app2.get_device(ieee)

    assert device.endpoints[1].basic._attr_cache[4] == b"Custom"
    assert device.endpoints[1].basic._attr_cache[5] == b"Model"
```

Adminiuga:
it only would happen for "quirked" devices, cause quirked devices would quit saving after updating the `devices` table with new nwk and in the proccess deleting endpoints and clusters on cascade.

this gets DB into inconsistent state:
```
diff --git a/tests/test_appdb.py b/tests/test_appdb.py
index c198b1e..e3e5d34 100644
--- a/tests/test_appdb.py
+++ b/tests/test_appdb.py
@@ -54,7 +54,9 @@ def make_ieee(init=0):
 
 
 class FakeCustomDevice(CustomDevice):
-    pass
+    replacement = {
+        "endpoints": {1: {"input_clusters": [0, 1, 3], "output_clusters": [6]}}
+    }
 
 
 def mock_dev_init(status: Status):
@@ -70,7 +72,7 @@ def mock_dev_init(status: Status):
 
 def fake_get_device(device):
     if device.endpoints.get(1) is not None and device[1].profile_id == 65535:
-        return FakeCustomDevice(device.application, make_ieee(1), 199, device)
+        return FakeCustomDevice(device.application, device.ieee, device.nwk, device)
     return device
 
 
@@ -483,3 +485,46 @@ async def test_neighbors(tmpdir):
     assert dev_2.neighbors[0].device is dev_1
     assert dev_2.neighbors[1].device is None
     assert dev_2.neighbors[1].neighbor.ieee == ieee_3
+
+
+@patch(
+    "zigpy.device.Device.schedule_initialize", new=mock_dev_init(Status.ENDPOINTS_INIT)
+)
+async def test_device_rejoin(tmpdir):
+    db = os.path.join(str(tmpdir), "test.db")
+    app = await make_app(db)
+    ieee = make_ieee()
+    nwk = 199
+    app.handle_join(nwk, ieee, 0)
+
+    dev = app.get_device(ieee)
+    ep = dev.add_endpoint(1)
+    ep.profile_id = 65535
+    ep.device_type = profiles.zha.DeviceType.PUMP
+    clus = ep.add_input_cluster(0)
+    ep.add_output_cluster(1)
+    app.device_initialized(dev)
+    clus._update_attribute(4, "Custom")
+    clus._update_attribute(5, "Model")
+
+    # Everything should've been saved - check that it re-loads
+    with patch("zigpy.quirks.get_device", fake_get_device):
+        app2 = await make_app(db)
+    dev = app2.get_device(ieee)
+    assert dev.nwk == nwk
+    assert dev.endpoints[1].device_type == profiles.zha.DeviceType.PUMP
+    assert dev.endpoints[1].in_clusters[0]._attr_cache[4] == "Custom"
+    assert dev.endpoints[1].in_clusters[0]._attr_cache[5] == "Model"
+    assert dev.endpoints[1].manufacturer == "Custom"
+    assert dev.endpoints[1].model == "Model"
+
+    # device rejoins
+    with patch("zigpy.quirks.get_device", fake_get_device):
+        app2.device_initialized(dev)
+    dev = app2.get_device(ieee)
+    assert dev.nwk == nwk
+    assert dev.endpoints[1].device_type == profiles.zha.DeviceType.PUMP
+    assert dev.endpoints[1].in_clusters[0]._attr_cache[4] == "Custom"
+    assert dev.endpoints[1].in_clusters[0]._attr_cache[5] == "Model"
+    assert dev.endpoints[1].manufacturer == "Custom"
+    assert dev.endpoints[1].model == "Model"
```

Adminiuga:
the test still works incorrectly though with the fix applied it still blows

Adminiuga:
or it quite possible there's another bug, when device rejoins with an already joined device then it blows the attribute cache???

puddly:
I think you have to re-populate the cache when the new `CustomEndpoint`s are created.

Adminiuga:
yeah, i think it is the cache blowing, till next attr update or reload

PR #521, opened 2020-10-20
Move quirks code out of `appdb.py` and into `application.py` and `device.py`

puddly:
- `Device`s are now replaced with `CustomDevice` objects within the `application` module instead of the `appdb` module, inside of `device_initialized` and `_load_db`. This simplifies the creation of other other database providers (e.g. JSON) and removes unrelated functionality from the `appdb` module.
 - `appdb` no longer sets the `manufacturer` and `model` attributes when creating a device. This is done dynamically within the `Device` class by finding the first `Basic` cluster with these cached attributes.
 - listeners on endpoints, clusters, and the device itself are preserved when creating a `CustomDevice`.

### DIFF
--- a/zigpy/device.py
+++ b/zigpy/device.py
@@ -10,6 +10,7 @@ import zigpy.exceptions
 import zigpy.neighbor
 from zigpy.types import NWK, BroadcastAddress, Relays
 import zigpy.util
+from zigpy.zcl.clusters.general import Basic
 import zigpy.zcl.foundation as foundation
 import zigpy.zdo as zdo
 
@@ -292,7 +293,28 @@ class Device(zigpy.util.LocalLogMixin, zigpy.util.ListenableMixin):
 
     @property
     def manufacturer(self):
-        return self._manufacturer
+        if self._manufacturer is not None:
+            return self._manufacturer
+
+        for endpoint_id, endpoint in self.endpoints.items():
+            if endpoint_id == 0:
+                continue
+
+            if Basic.cluster_id not in endpoint.in_clusters:
+                continue
+
+            cluster = endpoint.in_clusters[Basic.cluster_id]
+
+            if 0x0004 not in cluster._attr_cache:
+                continue
+
+            value = cluster._attr_cache[0x0004]
+
+            if isinstance(value, bytes):
+                value = value.split(b"\x00")[0]
+                return value.decode().strip()
+            else:
+                return value
 
     @manufacturer.setter
     def manufacturer(self, value):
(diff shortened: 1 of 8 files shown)


puddly:
Travis has been queued/running for more than half an hour now but everything passes locally for me with Tox.

zigpy/device.py
```diff
@@ -292,7 +293,28 @@ def ieee(self):
...
+            if 0x0004 not in cluster._attr_cache:
+                continue
+
+            value = cluster._attr_cache[0x0004]
+
+            if isinstance(value, bytes):
+                value = value.split(b"\x00")[0]
+                return value.decode().strip()
+            else:
+                return value
```
Adminiuga, line 317: I'd rather have done it once than every time you access the property

zigpy/device.py
```diff
@@ -306,7 +328,28 @@ def manufacturer_id(self) -> Optional[int]:
...
+            if 0x0005 not in cluster._attr_cache:
+                continue
+
+            value = cluster._attr_cache[0x0005]
+
+            if isinstance(value, bytes):
+                value = value.split(b"\x00")[0]
+                return value.decode().strip()
+            else:
+                return value
```
Adminiuga, line 352: same. as we're not expecting this to change

Adminiuga:
Is applying quirks this way looses attr cache on restoration?

puddly:
No, the attribute cache is still populated by `PerstingListener._load_attributes` and copied into the quirked clusters when a device is replaced. I believe this is explicitly tested by the main db unit test?

zigpy/device.py
puddly: I've added a `Device._update_model_manufacturer` method that is called when a new device is initialized or loaded from the DB.

Adminiuga: maybe just make this a private method, pass it the attr you are looking for and then set the `self._manufacturer` or `self.model` to the returned value.
So there's no code duplication and we don't run this every time the property is checked

puddly: Like `_update_model_manufacturer` is now doing (https://github.com/zigpy/zigpy/pull/521/files#diff-62448beb11fb75efec436d695b9dda880fb688b152a4969aa32e501124072072R393-R402)

It'll be run only once (at most twice I think?) per device. You can see this in the unit test: https://github.com/zigpy/zigpy/pull/521/files#diff-72107d6f4a703abaeb6a7cf709e0c99f7ed0b7c74f0b14b0cbc4f35c7c650e26R309-R337

Adminiuga:
are you running this in production?

puddly:
Not yet but I'm planning on testing it out later.

PR #528, opened 2020-10-27
Refactor persisting listener to use aiosqlite

Adminiuga:
Refactor `appdb.PersistingListener` to use [`aiosqlite`](https://github.com/omnilib/aiosqlite) so there's no disk io in the loop.
Fixes https://github.com/home-assistant/core/issues/42345
Adds new method `ControllerApplication.pre_shutdown()` which should be called instead of the `shutdown()` method implemented by the radio libraries.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,7 @@ setup(
     author="Russell Cloran",
     author_email="rcloran@gmail.com",
     license="GPL-3.0",
-    packages=find_packages(exclude=["*.tests"]),
-    install_requires=["aiohttp", "crccheck", "pycryptodome", "voluptuous"],
+    packages=find_packages(exclude=["tests"]),
+    install_requires=["aiohttp", "aiosqlite", "crccheck", "pycryptodome", "voluptuous"],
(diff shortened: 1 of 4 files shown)


puddly:
In reference to your comment on #659:

> the problem you might be running into, could be related to `aiosqlite`, it a test fails and you don't shutdown the persistent listener, then the test never exits, because aios1qlite thread never finishes, see [omnilib/aiosqlite#82](https://github.com/omnilib/aiosqlite/pull/82)

I actually ran into this exact issue just now 😆.

Can we rename the ControllerApplication `shutdown` method overridden by applications to `_shutdown` or `post_shutdown` or something? I would have never thought to call `pre_shutdown` instead of `shutdown` to shut down the application if I hadn't stumbled upon it in the zigpy source. It would be a lot more natural for `shutdown` to call `pre_shutdown`, not the other way around.

Adminiuga:
I agree, but that means all the radio libs would need a PR to update so I took the easy way. Feel free to submit PR for all the libs if you decide to take this approach. 
Maybe zigpy could have the `shutdown` method and the libraries could have `shutdown_ncp()` which has to be overridden. similar to permit_ncp 🤷

PR #531, opened 2020-10-29
Handle special identification for some sensors

abmantis:
This allows sensors like Xiaomi and Terncy ones to be initialized without exchanging interview messages.
Requires the following PR in quirks: https://github.com/zigpy/zha-device-handlers/pull/550

Example of how quick a Xiaomi Motion sensor gets paired (I've pressed the button only once): https://photos.app.goo.gl/GhyLpZEAHQqCgSsm9

### DIFF
--- a/zigpy/endpoint.py
+++ b/zigpy/endpoint.py
@@ -80,6 +80,26 @@ class Endpoint(zigpy.util.LocalLogMixin, zigpy.util.ListenableMixin):
 
         self.status = Status.ZDO_INIT
 
+    def initialize_from_quirk(self, endpoint):
+        if self.status == Status.ZDO_INIT:
+            return
+        self.profile_id = endpoint["profile_id"]
+        self.device_type = endpoint["device_type"]
+        if self.profile_id == 260:
+            self.device_type = zigpy.profiles.zha.DeviceType(self.device_type)
+        elif self.profile_id == 49246:
+            self.device_type = zigpy.profiles.zll.DeviceType(self.device_type)
+
+        for cluster in endpoint["input_clusters"]:
+            clus = self.add_input_cluster(cluster)
+            if cluster == 0:
+                clus._attr_cache[4] = self.manufacturer
+                clus._attr_cache[5] = self.model
+        for cluster in endpoint["output_clusters"]:
+            self.add_output_cluster(cluster)
+
+        self.status = Status.ZDO_INIT
+
     def add_input_cluster(self, cluster_id, cluster=None):
         """Adds an endpoint's input cluster
 
--- a/zigpy/quirks/registry.py
+++ b/zigpy/quirks/registry.py
@@ -15,6 +15,7 @@ class DeviceRegistry:
     def __init__(self, *args, **kwargs):
         dd = collections.defaultdict
         self._registry = dd(lambda: dd(list))
+        self._model_registry = dd(list)
 
     def add_to_registry(self, custom_device: CustomDeviceType) -> None:
         """Add a device to the registry"""
@@ -22,20 +23,24 @@ class DeviceRegistry:
         if models_info:
             for manuf, model in models_info:
                 self.registry[manuf][model].append(custom_device)
+                self._model_registry[model].append(custom_device)
         else:
             manufacturer = custom_device.signature.get("manufacturer")
             model = custom_device.signature.get("model")
             self.registry[manufacturer][model].append(custom_device)
+            self._model_registry[model].append(custom_device)
 
     def remove(self, custom_device: CustomDeviceType) -> None:
         models_info = custom_device.signature.get(SIG_MODELS_INFO)
         if models_info:
             for manuf, model in models_info:
                 self.registry[manuf][model].remove(custom_device)
+                self._model_registry[model].remove(custom_device)
         else:
             manufacturer = custom_device.signature.get("manufacturer")
             model = custom_device.signature.get("model")
             self.registry[manufacturer][model].remove(custom_device)
+            self._model_registry[model].remove(custom_device)
 
     def get_device(self, device: DeviceType) -> Union[CustomDeviceType, DeviceType]:
         """Get a CustomDevice object, if one is available"""
@@ -135,6 +140,11 @@ class DeviceRegistry:
 
         return device
 
+    def get_device_metadata(self, model, manufacturer=None):
+        if manufacturer:
+            return self.registry[manufacturer][model]
+        return self.model_registry[model]
+
     @staticmethod
     def _match(a, b):
         return set(a) == set(b)
@@ -142,6 +152,10 @@ class DeviceRegistry:
     @property
     def registry(self):
         return self._registry
+
+    @property
+    def model_registry(self):
+        return self._model_registry
 
     def __contains__(self, device: CustomDeviceType) -> bool:
         manufacturer, model = device.signature.get(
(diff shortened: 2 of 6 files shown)


puddly:
Nice! I think this also can be done entirely within quirks by intercepting and instantly responding to basic discovery requests sent by zigpy?

Since quirks already contains the full endpoint and cluster info for a lot of devices, it might be possible to quickly query the model / manufacturer on endpoint 1 (if that uniquely identifies the device) and continue with the full process only if that fails. That should keep Xiaomi-specific code out of core zigpy.

abmantis:
> Nice! I think this also can be done entirely within quirks by intercepting and instantly responding to basic discovery requests sent by zigpy?
I don't think we can do it all on quirks side, because zigpy replaces the device with the quirk after the init only. So, at the point of init, the device is not a quirk device.

> Since quirks already contains the full endpoint and cluster info for a lot of devices, it might be possible to quickly query the model / manufacturer on endpoint 1 (if that uniquely identifies the device) and continue with the full process only if that fails. That should keep Xiaomi-specific code out of core zigpy.
Are you suggesting sending a network request on cluster 0 to query the model/manuf? I've done that (check the commits before the last one), but @Adminiuga didn't think that is a good idea. Also, we would be adding an extra request for all devices that is not really needed.
And in the case of Xiaomi's they would not reply to that anyway, so we still need to handle the special message before discovery.

abmantis:
Anything else needed to get this reviewed?

Adminiuga (changes requested):
That's a lot of effort for little gain. Yeah, xiaomi a pain to pair, but you pair them just once. After that all rejoins are painless even without any cludges. I'd rather see that effort spent on
- multiple entity id support per cluster
- zgp support
- touchlink reset support

I've tried decoupling ZCL general command deserialization to be independent of the device instance and it is doable, **BUT** Xiaomi 🤦 some xiaomi devices send mallformated general commands and requires special treating again.

I don't want to merge it in the current form. will ask to change how it works and split into a number of smaller PRs against a separate branch.
All these PRs must pass existing tests with no or little modifications to the existing tests and all new code must have tests

abmantis:
> That's a lot of effort for little gain. Yeah, xiaomi a pain to pair, but you pair them just once. After that all rejoins are painless even without any cludges. I'd rather see that effort spent on
> 
>     * multiple entity id support per cluster
> 
>     * zgp support
> 
>     * touchlink reset support

IMO this is a big UX improvement. I had the need to migrate a network with a few Xiaomi sensors, and it sucks. With this, it would take just a second for each sensor.

Also, from a new user POV, ZHA will suck compared to Z2M or Deconz. A lot of new users have Xiaomi sensors. They will install ZHA, try to pair a sensor and it will fail (there's no indication on the UI that you need to press it a lot). Then they'll go to Deconz and it'll just work with a single reset on the sensor. I've heard this story first hand from a friend that tried to use ZHA and ended up in Deconz because he had a lot of Xiaomi sensors and was having a lot of trouble to pair them.

> I've tried decoupling ZCL general command deserialization to be independent of the device instance and it is doable, **BUT** Xiaomi facepalm some xiaomi devices send mallformated general commands and requires special treating again.

I understand that Xiaomi sensors suck spec-wise, but a lot of users have them, so...

> I don't want to merge it in the current form. will ask to change how it works and split into a number of smaller PRs against a separate branch.
> All these PRs must pass existing tests with no or little modifications to the existing tests and all new code must have tests

I don't understand what you want me to change. Just split and add tests?

zigpy/quirks/registry.py
```diff
@@ -15,27 +15,32 @@ class DeviceRegistry:
...
         self._registry = dd(lambda: dd(list))
+        self._model_registry = dd(list)
 
     def add_to_registry(self, custom_device: CustomDeviceType) -> None:
         """Add a device to the registry"""
         models_info = custom_device.signature.get(SIG_MODELS_INFO)
         if models_info:
             for manuf, model in models_info:
                 self.registry[manuf][model].append(custom_device)
+                self._model_registry[model].append(custom_device)
```
Adminiuga, line 26: I see no point duplicating model lists in `self._registry` dict and in `self._models` 
Can you refactor this into a separate PR where `self._registy` is a dict of manufactures with value of an entry in `self._model` dict?

e.g. something like
```
    def __init__(self, *args, **kwargs):
        dd = collections.defaultdict
        self._model_registry = dd(list)
        self._registry = dd(lambda: self._model_registry)
```

then you don't even need to expose the model registry and no need for 
```
        if manufacturer:
            return self.registry[manufacturer][model]
        return self.model_registry[model]
```

a simple `return self._registry[manufacturer][model] would do it`

puddly:
What about adding more events that receive objects from registered listeners, like in the OTA code (https://github.com/zigpy/zigpy/blob/dev/zigpy/ota/__init__.py#L72)? This will allow zigpy to notify interested listeners of the device initialization status and allow for device replacement at any point in the process. For example:

 - `device_joined`
 - `node_descriptor`
 - `unsolicited_basic_message` (maybe)
 - `model_manufacturer_info`

Centralizing the `CustomDevice` replacement and keeping it in `application.py` (https://github.com/zigpy/zigpy/pull/521) should help with this. These changes would move all of this special code into zha-device-handlers and keep it out of zigpy. and would also allow for some rudimentary progress to be added to ZHA's web UI.

zigpy/endpoint.py
```diff
@@ -80,6 +80,26 @@ async def initialize(self):
...
+        for cluster in endpoint["input_clusters"]:
+            clus = self.add_input_cluster(cluster)
+            if cluster == 0:
+                clus._attr_cache[4] = self.manufacturer
+                clus._attr_cache[5] = self.model
+        for cluster in endpoint["output_clusters"]:
+            self.add_output_cluster(cluster)
+
+        self.status = Status.ZDO_INIT
+
```
Adminiuga, line 102: Nope. Make a `module` method or `CustomDevice` class method to create a device instance from a quirk.
You give it a zigpy device instance and a model and it returns a  Quirk instance. Similar to https://github.com/zigpy/zha-device-handlers/blob/347b5084bf1cbc767e71f916087b6437b3b2d1c8/tests/conftest.py#L77

make this in a second PR

Adminiuga:
>Also, from a new user POV, ZHA will suck compared to Z2M or Deconz. A lot of new users have Xiaomi sensors. They will install ZHA, try to pair a sensor and it will fail (there's no indication on the UI that you need to press it a lot). 

I've explained it and really we could be more selective and provide more feedback during joining. The problem is not with zigpy, problem that in certain case newly joined devices they complete initialization relatively quickly in zigpy, but the entity in UI is going to show 90s later. That's the problem, not that xiaomi are slowly initializing.

Adminiuga:
> I personally struggle with joining Xiaomi switches to my network and end up manually editing them into the database so this would be a really useful addition.

The only struggle with xiaomi is to get them to scan the network and do association request. But once you've got the association request in, just keep it awake a few times and you are done

abmantis:
> > Also, from a new user POV, ZHA will suck compared to Z2M or Deconz. A lot of new users have Xiaomi sensors. They will install ZHA, try to pair a sensor and it will fail (there's no indication on the UI that you need to press it a lot).
> 
> I've explained it and really we could be more selective and provide more feedback during joining. The problem is not with zigpy, problem that in certain case newly joined devices they complete initialization relatively quickly in zigpy, but the entity in UI is going to show 90s later. That's the problem, not that xiaomi are slowly initializing.

I was just referring to xiaomi. I've already changed zha to skip init for xiaomi sensors and it still takes a bit in zigpy (and multiple presses). Not as much as before, ofc.

Adminiuga:
Closing this in favor of https://github.com/zigpy/zigpy/pull/555 Allows implementing all this in zha-device-handlers without big modifications to zigpy

PR #555, opened 2020-11-20
Framework implementing message handlers from uninitialized devices

Adminiuga:
1. Allow registration of `handle_message` handlers for incoming messages from uninitialized devices
2. Add `handle_message` event to remove ZHA hack

This would allow implementation of fast xiaomi initialization in quirks

### DIFF
--- a/zigpy/quirks/registry.py
+++ b/zigpy/quirks/registry.py
@@ -1,7 +1,7 @@
 import collections
 import itertools
 import logging
-from typing import Union
+from typing import Dict, List, Union
 
 import zigpy.quirks
 from zigpy.typing import CustomDeviceType, DeviceType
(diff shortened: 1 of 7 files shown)


zigpy/quirks/registry.py
```diff
@@ -19,11 +19,16 @@
...
+TYPE_MANUF_QUIRKS_DICT = Dict[str, TYPE_MODEL_QUIRKS_LIST]
+
 
 class DeviceRegistry:
     def __init__(self, *args, **kwargs):
-        dd = collections.defaultdict
-        self._registry = dd(lambda: dd(list))
+        self._model_quirks: TYPE_MODEL_QUIRKS_LIST = collections.defaultdict(list)
+        self._registry: TYPE_MANUF_QUIRKS_DICT = collections.defaultdict(
+            lambda: self._model_quirks
```
michalmo, line 30: This means the same instance (models list) will be shared for all the manufacturer ids. And is the cause of some issues people are having with device matching, eg: https://github.com/zigpy/zha-device-handlers/pull/538#issuecomment-738039431

Adminiuga: I need to test but you are probably right. If you reverse it, does it match correctly then?

michalmo: Yes, it matches correctly when i change it to `lambda: collections.defaultdict(list)`

Adminiuga: ok, that would require a change to zha-device-handlers too. Should do it before next release is shipped

Adminiuga: It is now in 1.0.0b3 and should be in the next release. Give it a try and open an issue if still is not fixed.

michalmo: confirmed working, thanks.

Issue #563, opened 2020-11-30
errors with ServiceCall zha.remove


uav2608:
I have several ZigBee devices installed in my Homeassistant with ZHA. I wanted to remove several of them with service 'ZHA.remove' and got error with the below log:
Logger: homeassistant.core
Source: components/zha/api.py:889
First occurred: 4:50:02 PM (2 occurrences)
Last logged: 4:50:04 PM

Error executing service: <ServiceCall zha.remove (c:1bfdc945a315675ac97491271717c9ce): ieee=00:15:8d:00:02:b0:fa:ae>
Error executing service: <ServiceCall zha.remove (c:ea476d66a59affd227817aaeb10043f2): ieee=00:15:8d:00:02:b0:fa:ae>
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/core.py", line 1467, in catch_exceptions
    await coro_or_task
  File "/usr/src/homeassistant/homeassistant/core.py", line 1486, in _execute_service
    await handler.job.target(service_call)
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 569, in admin_handler
    await result
  File "/usr/src/homeassistant/homeassistant/components/zha/api.py", line 889, in remove
    await application_controller.remove(ieee)
  File "/usr/local/lib/python3.8/site-packages/zigpy/application.py", line 150, in remove
    await self.force_remove(dev)
  File "/usr/local/lib/python3.8/site-packages/zigpy_znp/zigbee/application.py", line 660, in force_remove
    assert leave_rsp.Status == t.ZDOStatus.SUCCESS
AssertionError

puddly:
This issue has been fixed in the development branch of https://github.com/zigpy/zigpy-znp, which I'm going to release in a few days. Here are instructions for how to install it if you want to try it out: https://github.com/zigpy/zigpy-znp#testing-dev-with-home-assistant-core

I've also submitted a PR to zigpy to prevent this issue from causing `remove` to fail if the radio library's `force_remove` fails: https://github.com/zigpy/zigpy/pull/564

uav2608:
> This issue has been fixed in the development branch of https://github.com/zigpy/zigpy-znp, which I'm going to release in a few days. Here are instructions for how to install it if you want to try it out: https://github.com/zigpy/zigpy-znp#testing-dev-with-home-assistant-core
> 
[...]

PR #564, opened 2020-11-30
Ensure `app.force_remove` throwing an exception does not break removal

puddly:
Makes sure that a radio library's `force_remove` throwing an exception does not break removal from the DB (https://github.com/zigpy/zigpy/issues/563).

---

Imagine this scenario:

```
[Coordinator]  <--->  [Router]  <--->  [Router to be removed]
```

If the "router to be removed" is joined through "router" and you shut "router to be removed" off, it will retain the network information and have everything it needs to re-join the network. Calling `app.remove(router_to_be_removed)` will fail, as will the force remove. I believe powering this router back on will cause it to rejoin the network (since it never believes it left) and it will re-announce itself, causing zigpy to think it's a newly-joined device and start reconfiguring it. Is this expected behavior?

### DIFF
--- a/tests/test_application.py
+++ b/tests/test_application.py
@@ -298,6 +298,19 @@ async def test_remove_with_reply_timeout(app, ieee):
     assert app.force_remove.call_count == 1
 
 
+async def test_force_remove_failure(app, ieee):
+    app.force_remove = AsyncMock(side_effect=RuntimeError("Failure"))
+
+    # The application fails to remove the device
+    with pytest.raises(RuntimeError):
+        await _remove(app, ieee, [1])
+
+    assert app.force_remove.call_count == 1
+
+    # But the device was still removed
+    assert ieee not in app.devices
+
+
 def test_add_device(app, ieee):
     app.add_device(ieee, 8)
     app.add_device(ieee, 9)
--- a/zigpy/application.py
+++ b/zigpy/application.py
@@ -146,11 +146,13 @@ class ControllerApplication(zigpy.util.ListenableMixin, abc.ABC):
         except (zigpy.exceptions.DeliveryError, asyncio.TimeoutError) as ex:
             LOGGER.debug("Sending 'zdo_leave_req' failed: %s", ex)
 
-        if not zdo_worked:
-            await self.force_remove(dev)
-        self.devices.pop(ieee, None)
-
-        self.listener_event("device_removed", dev)
+        try:
+            if not zdo_worked:
+                await self.force_remove(dev)
+        finally:
+            # Remove the device from the database even if `force_remove` fails
+            self.devices.pop(ieee, None)
+            self.listener_event("device_removed", dev)
 
     async def force_remove(self, dev):
         raise NotImplementedError


Adminiuga:
Hrm,
I think there's a miss understanding of the `force_remove()` Really the `force_remove()` method stems from EZSP specific command `removeDevice()` which never fails, but technically we ignore the status of that EZSP command response. 
For other libraries, this was just a "plug" method notifying that device will be removed from zigpy and if radio has to do any cleanup, then it's the time to do it. But frankly speaking, even in EZSP this seems to be a redundant method, to what zigpy already does. 

>will the force remove. I believe powering this router back on will cause it to rejoin the network (since it never believes it left) and it will re-announce itself, causing zigpy to think it's a newly-joined device and start reconfiguring it. Is this expected behavior?

it is kind of expected behavior is at the moment. Technically the device is still on the network and we do use ZDO device annce as a trigger for a new device join. 

So, my expectation for `force_remove()` was that it is not doing any network traffic and if it does, then it handles the exceptions too. 
in `zigpy_znp`, what do you do for the `force_remove()`

On unrelated note, I think zigpy can be a bit smarter and do the following:
1. for the router, just sent the leave request and optionally log a warning if status != Success or it times out
2. for the end device:
  a. send zdo leave to the end device directly
  b. send zdo leave requests with child EUI to all parents which contain the child EUI in the neighbor table

Adminiuga:
> So, my expectation for force_remove() was that it is not doing any network traffic and if it does, then it handles the exceptions too. 

Or is there a compelling reason to have `force_remove()` at all? in other words, what potentially could work better after a failed ZDO leave request???

puddly:
> in `zigpy_znp`, what do you do for the `force_remove()`

In the current state of the `dev` branch, more or less what you suggested zigpy should do: https://github.com/zigpy/zigpy-znp/blob/32df3a8bbbba1935acbfa2fba0b3c2065a9335b0/zigpy_znp/zigbee/application.py#L668-L713 . I agree that this change should probably be moved into zigpy and `force_remove` made no-op in zigpy-znp, since Z-Stack doesn't expose anything lower-level like EZSP's APS leave (I think that's what force remove does?).

Adminiuga:
>On unrelated note, I think zigpy can be a bit smarter and do the following:
for the router, just sent the leave request and optionally log a warning if status != Success or it times out
for the end device:
a. send zdo leave to the end device directly
b. send zdo leave requests with child EUI to all parents which contain the child EUI in the neighbor table

Ha, this is actually what you are doing :) https://github.com/zigpy/zigpy-znp/blob/32df3a8bbbba1935acbfa2fba0b3c2065a9335b0/zigpy_znp/zigbee/application.py#L668

puddly:
This version contains fewer bugs:

```Python
parents = []

for dev in self.devices.values():
    for neighbor in dev.neighbors:
        # Scanned relationships are not always both present so we need to find
        # potential parents through both the parent and the child
        if (
            dev is device
            and neighbor.neighbor.relationship == Neighbor.RelationShip.Parent
            and neighbor.device not in parents
        ):
            parents.append(neighbor.device)
        elif (
            neighbor.device is device
            and neighbor.neighbor.relationship == Neighbor.RelationShip.Child
            and dev not in parents
        ):
            parents.append(dev)
```

Adminiuga:
```
        if (
            dev is device
            and neighbor.neighbor.relationship == Neighbor.RelationShip.Parent
            and neighbor.device not in parents
        ):
            parents.append(neighbor.device)
```
is this branch when the `device` itself is a router? is your intention sending the ZDO leave request not only to the router, but to one of its neigbour router too?

puddly:
I'm thinking (and I may be mistaken) that if the device doesn't receive the leave request normally, maybe it changed parents and zigpy never was notified, so we should try sending it to all devices that claim to be `device`'s parent, and also to all parents reported by `device`.

Adminiuga:
Can End Devices contain the neighbor table? The "neighbor is parent" relationship does not make sense to me, cause I'm under impression the end devices do not maintain the neighbors table. If some do, would be interesting to see the neighbour table of its parent.
image

I was just thinking just find all neighbours which contain that device and the device type is the end_device, although we know upfront the device type of the device we are removing, so I think we can just filter through all devices which contain that ieee in the neighbours.

puddly:
I think I may just be confused about the definitions. On my live network, I'm observing the following topology scan results (400 total entries):

 - The majority of relationships (381/400) are "Sibling".
 - Only RFDs, not routers, appear as "Child" (16/400). I was mistakenly thinking that routers kept track of other routers joined through them.
 - I am seeing a few (3/400) records identifying other routers (and the coordinator) as a "Parent". Only IKEA plug-in switches. No bulbs. Is this some firmware quirk?

Adminiuga:
> I am seeing a few (3/400) records identifying other routers (and the coordinator) as a "Parent". Only IKEA plug-in switches. No bulbs. Is this some firmware quirk

Are you saying the devices reporting a "Parent" in their neighbors table are only IKEA plug-in switches? They do report themselves as a Zigbee router in the node descriptor, don't they?

We do know the device_type of the devices we are removing, so I think it is safe to send ZDO leave request to any other device if both are true:
1. device we are removing is an end device
2. device we are removing is in the neighbor table of the other device

puddly:
> Are you saying the devices reporting a "Parent" in their neighbors table are only IKEA plug-in switches?

Here's the exact response:

```Python
ZDO.MgmtLqiRsp.Callback(
    Src=0x43FD,  # TRÅDFRI Outlet
    Status=<Status.SUCCESS: 0>,
    Neighbors=Neighbors(
        entries=17,
        start_index=15,
        neighbor_table_list=[
            Neighbor(
                extended_pan_id=a8:c0:3b:db:53:ca:60:a8,
                ieee=90:fd:9f:ff:fe:32:9a:0b,
                nwk=0xFBA4,  # TRÅDFRI Outlet
                packed=5,  # Router + RxOnWhenIdle.On + Parent
                permit_joining=<PermitJoins.Unknown: 2>,
                depth=15,
                lqi=255
            ),
            Neighbor(
                extended_pan_id=a8:c0:3b:db:53:ca:60:a8,
                ieee=00:15:8d:00:01:e5:16:b7,
                nwk=0x7542,  # Aqara Motion
                packed=18,  # EndDevice + RxOnWhenIdle.Off + Child
                permit_joining=<PermitJoins.NotAccepting: 0>,
                depth=2,
                lqi=162
            )
        ]
    )
)
```

> They do report themselves as a Zigbee router in the node descriptor, don't they?

They do. They're both running the latest available firmwares and have the same node descriptor.

MattWestb:
All my IKEA Zigbee 3 and LL router devices is parents to the coordinator and all other ZB3 routers is siblings (2 lines between the routers ala 2 ways communicating) and My Philips HUE bulb is sibling to the coordinator and only one line to all other routers. The same was with the chinese HOMA LED drives only one line and coordinator as sibling (is in deCONZ for the moment).
I think its have with the zigbee stack if its working with ZigbeePro (not need ZB3) or some older standard.

ZigZag02

Philips HUE Router. All router connections with one line and as siblings also to ZNCP.

ZigZag03

ZNCP view. All IKEA routers have the ZNCP as parent and have all other routers as siblings with 2 lines (only one to the HUE). 

In one real Zigbee 3 network old not ZB3 routers is degraded as end devices I have reading but i think its more like one "multi connected" end devices.

In deCONZ the HOMAs is only talking  to the coordinator and not to other routers that  is very near (no line at all between them only router - ZNCP).

MattWestb:
Neighbour reported from ZHA-MAP from the Philips HUE router.
The` "relation": "None_of_the_above",` its interesting !!

[neighbours_0017880102c68659.txt](https://github.com/zigpy/zigpy/files/5638903/neighbours_0017880102c68659.txt)

Alos to the ZNCP.

Adminiuga:
I feel IKEA and some other vendors may use "parent" relationship to designate a device it's using for routing to coordinator

MattWestb:
My 2 new LIDL 3 x plugs with USB is one having the other as parent and not the NCP (not in the old ZigZac) after some days in the mesh so your feelings i can share in that technical case !

Then the strange is the Philips HUE bulb that only having siblings and no parents (relationship 0x03 from the table and in my attached neighbor file a bow). Is my HUE bulb working in Zigbee 3 compatibility mode as one no routing ZLL device ? 

Not important then its working but INTERESTING !!

MattWestb:
My old Philips HUE bulb is reporting ZCL ver =1 and the old IKEA Motion sensor = both is ZLL with or without Zigbee PRO( R22).
Normally IKEA ZB3 is reporting 2
All LIDL / tuya Zigbee 3 is reporting 3 (Zigbee 3 ++)

I cant finding it but its written in more species that ZB3 devices shall reported 2 and older Zigbee / old PRO shall reporting 1.
Attribute 3 is reserved for some of the power profile in some paper (I think its the Zigbee Smart Energy (ZSE)) but i cant finding it for the moment.

I still think that no Zigbee PRO (R22) routes is degraded in ZB3 and is working more or less as no sleeping end devices with multi connections ala my old HUE bulb that dont have any parents.

I was buying one LIDL E14 RGBW yesterday for putting inside one christmas star and its have also taking the 3 x outlet with USB as its parent so can being it have good antenna (I was trying opening it but its 3 wings screws very deep inside so not possible without the right tools) or its prefuring routers with ZCL = 3 (not Bellows).

PR #577, opened 2020-12-13
Support arbitrary callables as quirk matchers

puddly:
The goal of this change is to support complex quirks that accept a `Device` instance and possibly return a replacement class (or instance). At the moment it's not very useful but there has been interest in supporting quirks kept independently of the `zhaquirks` package, maybe as a config option:

```yaml
zigpy:
  custom_quirks_path: /config/custom_zha_quirks
```

One use case is supporting modified Xiaomi motion sensors, which have a timeout of 5s instead of the usual 120s. They are otherwise indistinguishable from other Xiaomi motion sensors and cannot be matched with a normal quirk:

```Python
# /config/custom_zha_quirks/fast_xiaomi.py
import zigpy.quirks
from zhaquirks.xiaomi.aqara.motion_aq2 import MotionAQ2

FAST_XIAOMI_SENSORS = {"AA:BB:CC:DD:EE:FF", "00:11:22:33:44:55:66"}


@zigpy.quirks.register_quirk_matcher
def fast_xiaomi_motion(device):
    if str(device.ieee).upper() not in FAST_XIAOMI_SENSORS:
        return

    replacement = MotionAQ2(device._application, device.ieee, device.nwk, device)
    replacement.endpoints[1].occupancy.reset_s = 5

    return replacement
```

This approach can also be used to simplify the existing quirks matching code within zigpy and possibly move some of it into zha-device-handlers.

### DIFF
--- a/tests/test_quirks.py
+++ b/tests/test_quirks.py
@@ -740,3 +740,91 @@ def test_different_manuf_same_model(real_device, real_device_2):
     )
     assert len(manuf2_list) == 1
     assert manuf2_list[0] is TestDevice_2
+
+
+@pytest.mark.parametrize("should_match", [True, False])
+def test_quirk_matchers(should_match, real_device, real_device_2):
+    class TestDevice_1(zigpy.quirks.CustomDevice):
+        signature = {
+            SIG_MODELS_INFO: (("manufacturer", "model"),),
+            SIG_ENDPOINTS: {
+                1: {
+                    SIG_EP_PROFILE: 255,
+                    SIG_EP_TYPE: 255,
+                    SIG_EP_INPUT: [3],
+                    SIG_EP_OUTPUT: [6],
+                }
+            },
+        }
+
+        def get_signature(self):
+            pass
+
+    class TestDevice_2(zigpy.quirks.CustomDevice):
+        signature = {
+            SIG_MODELS_INFO: (("A different manufacturer", "model"),),
+            SIG_ENDPOINTS: {
+                1: {
+                    SIG_EP_PROFILE: 255,
+                    SIG_EP_TYPE: 255,
+                    SIG_EP_INPUT: [3],
+                    SIG_EP_OUTPUT: [6],
+                }
+            },
+        }
+
+        def get_signature(self):
+            pass
+
+    registry = DeviceRegistry()
+    registry.add_to_registry(TestDevice_1)
+    registry.add_to_registry(TestDevice_2)
+
+    @registry.add_matcher_to_registry
+    def matcher(device):
+        if should_match and device.ieee == real_device.ieee:
+            return TestDevice_2
+
+    if should_match:
+        assert isinstance(registry.get_device(real_device), TestDevice_2)
+    else:
+        assert isinstance(registry.get_device(real_device), TestDevice_1)
+
+    assert isinstance(registry.get_device(real_device_2), TestDevice_2)
+
+
+def test_quirk_matcher_instance(real_device, real_device_2):
+    class TestDevice_2(zigpy.quirks.CustomDevice):
+        signature = {
+            SIG_MODELS_INFO: (("A different manufacturer", "model"),),
+            SIG_ENDPOINTS: {
+                1: {
+                    SIG_EP_PROFILE: 255,
+                    SIG_EP_TYPE: 255,
+                    SIG_EP_INPUT: [3],
+                    SIG_EP_OUTPUT: [6],
+                }
+            },
+        }
+
+        def get_signature(self):
+            pass
+
+    registry = DeviceRegistry()
+
+    @registry.add_matcher_to_registry
+    def matcher(device):
+        if device.ieee != real_device.ieee:
+            return
+
+        new_device = TestDevice_2(device._application, device.ieee, device.nwk, device)
+        assert device.nwk != 3456
+        new_device.nwk = 3456
+
+        return new_device
+
+    new_real_device = registry.get_device(real_device)
+    assert isinstance(new_real_device, TestDevice_2)
+    assert new_real_device.nwk == 3456
+
+    assert registry.get_device(real_device_2) is real_device_2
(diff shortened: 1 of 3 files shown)


Adminiuga:
specifically for fast xiaomi sensors, I'd prefer another approach: virtual attribute writes. In other words leverage zigpy cache.
A successful `write_attributes()` updates zigpy cache and it is persistent across reboots. We already have "plugged attribute reads" we could do the same for attribute writing. Specifically for xiaomi, cause I don't know if you can write anything to their clusters, so it is trivial to plug. Then just define a "plugged writtable attribute" and use it as your reset timeout. And we already have UI for attribute manipulations, so should be more user friendlier than modifying the configs or quirks.

dmulcahey:
@Adminiuga what about documenting vars folks can set... and using them via the existing device config? the place where we can override things like the matched component / device type? Then we pass that config to Quirks and if the well known key is present -> use the value otherwise use the default (hard coded value) To me this would be an extensible way to handle this. Could be used for things like alarm code for the control panels when I finish IAS ACE stuff and other similar things...

dmulcahey:
I also think this could be a cool approach to allow folks to override built in quirks... looks interesting

puddly:
Right now quirks are stateless and match from scratch every startup which I think is quite beneficial, as you will be able to split apart one quirk into two or modify broken quirk signatures without requiring database changes (i.e. if a firmware upgrade changes functionality or an old quirk accidentally matches two devices). Using the attribute cache would keep quirk state in `zigbee.db`, which I think would be a step backwards.

An alternative, as @dmulcahey said, would be "configurable" quirks from within the zigpy config. This would also keep unnecessary state out of the database and allow for easy tweaking.

Adminiuga:
> Right now quirks are stateless and match from scratch every startup which I think is quite beneficial, as you will be able to split apart one quirk into two or modify broken quirk signatures without requiring database changes (i.e. if a firmware upgrade changes functionality or an old quirk accidentally matches two devices). Using the attribute cache would keep quirk state in zigbee.db, which I think would be a step backwards.

Hrm? yeah, but I'm not saying to change quirk matching. My comment was a tad unrelated. It wasn't about the quirk matching, but about having the users to create/upload a custom quirk for their modified xiaomis. My point, specifically for xiaomi or any other IAS custom clusters which have to be reset in quirk, a configurable "self reset" timeout is something users have been asking for a while now. Yeah you keep timeout in the zigpy cache, one could argue: we keep on-off state of the on-off cluster is zigpy cache too. Or divisor/multipliers for the measearument clusters. Same idea

Adminiuga:
> The goal of this change is to support complex quirks that accept a Device instance and possibly return a replacement class (or instance). At the moment it's not very useful but there has been interest in supporting quirks kept independently of the zhaquirks package, maybe as a config option:

I'm really torn about this one. This is one more way for users to shoot themselves in the foot. Just one misconfigured matcher or quirk and you are flooded with issue reports. 
On the other hand, could be useful to let users to test a quirk that is being developed and provide feedback.

Adminiuga:
Modify ZHA gateway to also try loading quirks from `/config/custom_components/zhaquirks` ?

puddly:
This type of functionality may also be useful for Tuya using tons of different manufacturer names: https://github.com/zigpy/zha-device-handlers/issues/679, https://github.com/zigpy/zha-device-handlers/issues/677, https://github.com/zigpy/zha-device-handlers/issues/668, https://github.com/zigpy/zha-device-handlers/issues/663, ...

Adminiuga:
Yeah, but the signatures are still slightly different for those Tuyas. 
This goes deeper. Do you just match by model and manufacturer, but that what do you do for endpoints if they differ between vendors? replace clusters selectively?

Adminiuga:
> This type of functionality may also be useful for Tuya using tons of different manufacturer names with the same prefix but a randomly generated suffix

For Tuyas, maybe it is better to change the signature to include only the "model", thus in would match any "manufacturer" and we won't have to update signature for each vendor.

MattWestb:
I think only model is not working for tuya TRV.
Look in the signature for model and manufacturer in https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/tuya/valve.py

The` ("_TYST11_kfvq6avy", "fvq6avy"),` is the same as ` ("_TZE200_kfvq6avy", "TS0601"),` but with different zigbee module and also the cluster is different.

Only for being sure pleas taking on look before making some large change for not getting problems.

Adminiuga:
It still would match based on clusters too, so going to try two quirks before finding a match

rednus:
Hi This sounds interesting. 

One of the use case I can think of is Tuya. As @Adminiuga  said Tuya is becoming a big PITA. Manufacturers implementing the API on devices and not following the Tuya protocol stricktly. Its not just on the device signature, but also the Cluster Commands being used and the ZHA device implementation.  One example I can give which I have been working on 744(https://github.com/zigpy/zha-device-handlers/issues/744#issuecomment-797720318) , TS0601 window coverings. Apart from having different device signatures despite using same hardware, some devices take 0x01 for open and some take 0x02; Some use 0x02 command when sending status report, some use 0x02; oh and the DP (data point value) can be different for each manufacturer which should be part of the command being sent to the device to perform a specific action.  But this way users can implement their own quirks not having to worry about breaking someone else's device functionality. 

In my little experience; if I can suggest the below: 
* Use the  /config/custom_components/zhaquirks as @Adminiuga  suggested, this way users can install the quirks for the devices they own from HACS if available or copy their own quirk there. 
* It would be better if this would be a full quirk implementation than just changing small bits
* Look at quirks here before trying to match from packaged zigpy/zha-device-handler quirks (i.e. custom quirks take priority)

Finally if possible: 
* Have a file in custom_components/quirks say fixtures.py where users can specify device ieee and quirk or custom_quirk to force use; useful if their device does not follow the signature but a quirk is sufficient to support it
*  Not sure about this, but a possibility to reload a quirk (after modification) without restarting HA

Off topic, would it be possible to implement friendly name replacements (for manufacturer and model)?

puddly:
Now that https://github.com/zigpy/zha-device-handlers/pull/802 been merged, custom quirks can be written. The folder structure is a little different but they still take precedence over built-in quirks. Just change the relative imports to absolute imports.

PR #589, opened 2020-12-19
ControllerApplication State class and counters

Adminiuga:
A container class to keep minimal network and node information in one place and framework for "Counters".
Instead of keeping attributes like `pan_id`, `extended_pan_id` as the attributes of the `ControllerApplication()` instance, keep those in one class. 
Add a Counter() framework, to keep request/success failure statistics per device. Originally the counters were implemented to deal with EZSP counter overflow, but I've tested an implementation of software counters for request success/failure stats for `bellows` radio module and it worked very well, with stats being very close to the hardware counters. The plan is to keep counter stats on per device level, which should bring a bit more visibility into health of the network.

### DIFF
--- a/tests/test_app_state.py
+++ b/tests/test_app_state.py
@@ -1,0 +1,194 @@
+"""Test unit for app status and counters."""
+
+import pytest
+
+import zigpy.application.state as app_state
+
+COUNTER_NAMES = ["counter_1", "counter_2", "some random name"]
+
+
+@pytest.fixture
+def counters():
+    """Counters fixture."""
+    return app_state.Counters("ezsp_counters", COUNTER_NAMES)
+
+
+def test_counter():
+    """Test basic counter."""
+
+    counter = app_state.Counter("mock_counter")
+    assert counter.value == 0
+
+    counter = app_state.Counter("mock_counter", 5)
+    assert counter.value == 5
+    assert counter.reset_count == 0
+
+    counter.update(5)
+    assert counter.value == 5
+    assert counter.reset_count == 0
+
+    counter.update(8)
+    assert counter.value == 8
+    assert counter.reset_count == 0
+
+    counter.update(9)
+    assert counter.value == 9
+    assert counter.reset_count == 0
+
+    counter.reset()
+    assert counter.value == 9
+    assert counter._raw_value == 0
+    assert counter.reset_count == 1
+
+    # new value after a counter was reset/clear
+    counter.update(12)
+    assert counter.value == 21
+    assert counter.reset_count == 1
+
+    counter.update(15)
+    assert counter.value == 24
+    assert counter.reset_count == 1
+
+    # new counter value is less than previously reported.
+    # assume counter was reset
+    counter.update(14)
+    assert counter.value == 24 + 14
+    assert counter.reset_count == 2
+
+    counter.reset_and_update(14)
+    assert counter.value == 38 + 14
+    assert counter.reset_count == 3
+
+
+def test_counter_str():
+    """Test counter str representation."""
+
+    counter = app_state.Counter("some_counter", 8)
+    assert str(counter) == "some_counter = 8"
+
+
+def test_counters_init(counters):
+    """Test counters initialization."""
+
+    assert counters.name == "ezsp_counters"
+    assert counters.list
+    assert len(counters.list) == 3
+
+    cnt_1, cnt_2, cnt_3 = counters.list
+    assert cnt_1.name == "counter_1"
+    assert cnt_2.name == "counter_2"
+    assert cnt_3.name == "some random name"
+
+    assert cnt_1.value == 0
+    assert cnt_2.value == 0
+    assert cnt_3.value == 0
+
+    with pytest.raises(KeyError):
+        counters["no such counter"]
+
+    counters["some random name"] = 2
+    assert cnt_3.value == 2
+    assert counters["some random name"].value == 2
+    assert counters["some random name"] == 2
+    assert counters["some random name"] == cnt_3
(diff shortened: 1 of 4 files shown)


tests/test_app_state.py
```diff
@@ -0,0 +1,194 @@
...
+    assert cnt_2.value == 0
+    assert cnt_3.value == 0
+
+    with pytest.raises(KeyError):
+        counters["no such counter"]
+
+    counters["some random name"] = 2
+    assert cnt_3.value == 2
+    assert counters["some random name"].value == 2
+    assert counters["some random name"] == 2
```
puddly, line 92: Does `Counter.__eq__` having a special case for `int` have any use other than this equality shorthand? It might be better to just use `.value` everywhere.

zigpy/application/state.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+@dataclass
+class NetworkInformation:
+    """Network information."""
+
+    extended_pan_id: Optional[t.ExtendedPanId] = None
+    pan_id: Optional[t.PanId] = 0xFFFE
+    nwk_update_id: Optional[t.uint8_t] = 0x00
+    nwk_manager_id: Optional[t.NWK] = t.NWK(0xFFFE)
```
puddly, line 34: Is this an EZSP-specific property? Or for distributed networks?

zigpy/application/state.py
```diff
@@ -0,0 +1,220 @@
...
+
+import zigpy.types as t
+import zigpy.zdo.types as zdo_t
+
+
+@dataclass
+class NodeInfo:
+    """Controller Application network Node information."""
+
+    nwk: t.NWK = t.NWK(0xFFFE)
```
puddly, line 15: Is the coordinator's `nwk` ever not `0x0000`?

puddly:
I think the counters seem pretty useful for gauging real-world link quality but I'm having a hard time figuring out the motivation behind the `NetworkInformation` and `NodeInfo` classes. Can you separate this into two PRs?

Having the network information be attributes/properties of the `ControllerApplication` class seemed pretty simple to me. Nesting them inside of objects doesn't seem to make things any easier (i.e. `app.extended_pan_id` vs `app.state.network_information.ieee) and the extra properties on these objects seem like they always have the same values (i.e. `NodeInfo.nwk == 0x0000` and `NodeInfo.logical_type == LogicalType.Coordinator`)? Or are you thinking of allowing zigpy to configure radios as routers as well as coordinators?

tests/test_app_state.py
Adminiuga: Just a shorthand. Probably doesn't make much sense without __le__ and others

zigpy/application/state.py
Adminiuga: This is Network Manager and is Zigbee specific. Part of the persistent data, each node holds

zigpy/application/state.py
Adminiuga: it is not, but NCP could be a router or an end device. But not now.

Adminiuga:
> Or are you thinking of allowing zigpy to configure radios as routers as well as coordinators?

Yes. Should not assume zigpy will always be a coordinator.

I also would like to be able just to do `_LOGGER.info("network info: %s", network_inf)` in ZHA for it to print all the relevant network information without assembling it by pieces. 

NWK and IEEE have nothing to do with the actual "network information" and keeping them in a flat structure is meh.

zigpy/application/state.py
puddly: I think this change is useful. What do you think about adding a few more fields?

```Python
@dataclass
class NodeInfo:
    """Controller Application network Node information."""

    nwk: t.NWK | None
    ieee: t.EUI64
    logical_type: zdo_t.LogicalType

    # New
    aps_link_key: t.KeyData | None
    aps_tx_counter: t.uint32_t | None
    aps_rx_counter: t.uint32_t | None


@dataclass
class NetworkInformation:
    """Network information."""

    extended_pan_id: t.ExtendedPanId
    pan_id: t.PanId
    nwk_update_id: t.uint8_t
    nwk_manager_id: t.NWK
    channel: t.uint8_t
    channel_mask: t.Channels

    # New
    security_level: t.uint8_t
    nwk_key: t.KeyData
    nwk_key_seq: t.uint8_t
    nwk_key_frame_counter: t.uint32_t

    associated_nodes: list[NodeInfo]
```

If radio libraries formed networks from a `NetworkInformation` object and returned one when starting up an adapter, this would allow for seamless backup/restore between coordinators and for network settings to easily be changed.

Adminiuga: That's fine by me. Originally this was intended to be passed to UI to display information about the controller, that's why sensitive info was excluded, but we can filter that out and have a single object for both purposes.

puddly: Do you want me to make a separate PR with just the network info changes or do you want to merge one of your counter branches instead?

Adminiuga: I need to create a new PR. probably over the weekend. I'll incorporate your changes

Adminiuga:
```
    # New
    aps_link_key: t.KeyData | None
    aps_tx_counter: t.uint32_t | None
    aps_rx_counter: t.uint32_t | None
```

Should this be a Trust Center Link Key instead? And maybe add a list or dict of APS link keys, because usually APS key is unique per specific destination? i.e.  if you a device on the networks wants to use encrypted APS with another device (not Coordinator), then it is going to use the APS key specific to that destination device only?

Adminiuga:
closing in favor of #787

PR #597, opened 2020-12-23
Single function for OTA image parsing, CachedImage simplification, and Hue SBL images

puddly:
This change centralizes all OTA image parsing into a `zigpy.ota.image.parse_ota_image` function that properly deals with IKEA containers and now Hue SBL images. A side effect of this change is allowing `FileProvider` to properly unwrap IKEA images (rather than scanning for the Zigbee OTA header in the first 512 bytes) and all future weird file types without any necessary changes.

`CachedImage` was rewritten to be a container around any type of image that has a header and can be serialized. I've also added a `CachedImage.cached_data`, which prevents the wrapped image from being repeatedly serialized every time a block is requested.

### DIFF
--- a/zigpy/ota/__init__.py
+++ b/zigpy/ota/__init__.py
@@ -26,6 +26,7 @@ class CachedImage:
 
     image = attr.ib(default=None)
     expires_on = attr.ib(default=None)
+    cached_data = attr.ib(default=None)
 
     @classmethod
     def new(cls, img):
@@ -75,13 +76,13 @@ class CachedImage:
         ):
             self.expires_on += DELAY_EXPIRATION
 
-        # XXX: this should be cached
-        data = self.image.serialize()
+        if self.cached_data is None:
+            self.cached_data = self.image.serialize()
 
-        if offset > len(data):
+        if offset > len(self.cached_data):
             raise ValueError("Offset exceeds image size")
 
-        return data[offset : offset + min(self.MAXIMUM_DATA_SIZE, size)]
+        return self.cached_data[offset : offset + min(self.MAXIMUM_DATA_SIZE, size)]
 
 
 class OTA(zigpy.util.ListenableMixin):
--- a/zigpy/ota/__init__.py
+++ b/zigpy/ota/__init__.py
@@ -6,28 +6,31 @@ from typing import Optional
 import attr
 
 from zigpy.config import CONF_OTA, CONF_OTA_DIR, CONF_OTA_IKEA, CONF_OTA_LEDVANCE
-from zigpy.ota.image import ImageKey, OTAImage
+from zigpy.ota.image import ImageKey
 import zigpy.ota.provider
 from zigpy.ota.validators import check_invalid
+import zigpy.types as t
 from zigpy.typing import ControllerApplicationType
 import zigpy.util
 
 LOGGER = logging.getLogger(__name__)
 
-DELAY_EXPIRATION = datetime.timedelta(hours=2)
 TIMEDELTA_0 = datetime.timedelta()
+DELAY_EXPIRATION = datetime.timedelta(hours=2)
 
 
 @attr.s
-class CachedImage(OTAImage):
+class CachedImage:
+    MAXIMUM_DATA_SIZE = 40
     DEFAULT_EXPIRATION = datetime.timedelta(hours=18)
 
+    image = attr.ib(default=None)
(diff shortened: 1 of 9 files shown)


juched78:
Was curious so I tried pulling down the 4 changed files and inserted them into my HA docker image.  Tried the update on my RWL020 dimmer switch which has an SBL ota file.  

Logs

`2020-12-23 11:10:11 DEBUG (MainThread) [zigpy.zcl] [0x8e19:2:0x0019] ZCL deserialize: <ZCLHeader frame_control=<FrameControl frame_type=CLUSTER_COMMAND manufacturer_specific=False is_reply=False disable_default_response=False> manufacturer=None tsn=210 command_id=1>
2020-12-23 11:10:11 DEBUG (MainThread) [zigpy.zcl] [0x8e19:2:0x0019] ZCL request 0x0001: [0, 4107, 265, 1107314102, None]
2020-12-23 11:10:11 DEBUG (MainThread) [zigpy.zcl] [0x8e19:2:0x0019] OTA query_next_image handler for 'Philips RWL020': field_control=0, manufacture_id=4107, image_type=265, current_file_version=1107314102, hardware_version=None
2020-12-23 11:10:11 ERROR (MainThread) [zigpy.zcl] [0x8e19:2:0x0019] Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/zigpy/zcl/clusters/general.py", line 993, in _handle_cluster_request
    await self._handle_query_next_image(*args, tsn=tsn)
  File "/usr/local/lib/python3.8/site-packages/zigpy/zcl/clusters/general.py", line 1043, in _handle_query_next_image
    img.header.image_size,
AttributeError: 'CachedImage' object has no attribute 'header'`

Perhaps expected, but files outside of ota folder use CachedImage class and now need updating.  

Just a data point.  Thank you for making these changes.  Really appreicated.

puddly:
Whoops, looks like I need more unit tests. Thanks for trying this out.

juched78:
Replaced __init__.py and tried again.  It took a while but successfully upgraded my Phillips hue dimmer switch.

Thank you!

Adminiuga (approved):
LGTM

Adminiuga:
For signed images, isn't OTA header part of the signed data? 
in other words, would we be better off just storing the blob starting with OTA header magic number till the end of file?

puddly:
The `.signed` images can hold any data and in this case just wrap the Zigbee OTA image. They're completely unwrapped before parsing so you can pretend the error is with just the raw `.ota` file.

I guess we instead could send `len(image.serialize())` as the image length instead of `image.header.image_size` during the OTA process? With this fix, no more "bad" IKEA images remain so we in theory don't need the validation code anymore (but we still need a way to prevent OTA loops, maybe by blacklisting images after failed updates).

zigpy/ota/__init__.py
```diff
@@ -6,42 +6,87 @@
...
+    MAXIMUM_DATA_SIZE = 40
     DEFAULT_EXPIRATION = datetime.timedelta(hours=18)
 
+    image = attr.ib(default=None)
     expires_on = attr.ib(default=None)
+    cached_data = attr.ib(default=None)
 
     @classmethod
-    def new(cls, img: OTAImage):
+    def new(cls, img):
```
Adminiuga, line 32: what's the type of `img` ? can we add it?

puddly: Whoops. I think I removed it because there was a circular import at some point in time. It'd be `BaseOTAImage`.

Adminiuga: Yeah, Circular imports with typing are PITA. Feel free to add that type to `zigpy.typing` then

Adminiuga:
overall looks good.
Are you running this in the prod?

puddly:
Yeah, I used it to re-update my IKEA plug yesterday. The file provider also successfully accepts every single OTA file in https://github.com/Koenkk/zigbee-OTA/ and all of the `.signed` IKEA files straight from IKEA's servers, properly rejecting only the gateway image.

puddly:
My network is somewhat homogeneous because IKEA is the only brand that I can update so maybe give it a shot yourself if you have any Hue devices that use the `.sbl-ota` images.

Adminiuga:
is this good to go or are you planning any other tweaks?

puddly:
It's good to go.

PR #609, opened 2021-01-07
Update README.md with warning about Zigbee to WiFi bridges

Hedda:
Update README.md with warning about Zigbee to WiFi bridges


Adminiuga:
It should be in bellows, not zigpy

Hedda:
> It should be in bellows, not zigpy

General warning against using WiFi bridge or is serial over WiFi not an issue with any manufacturers other than Silicon Labs?

Issue #592, opened 2020-12-20
Delayed message processing (Tuya)


xonestonex:
Hi,

I have a tuya radiator thermostat and see a huge delay in message processing of this device compared to Zigbee2Mqtt. 
Setting the temperature on the device is a good example for this problem. The temperature on the device is changed in 0.5 degree steps. Each step produces at least 3 messages (configured temperature, current measured temperature, device state, …). Zigpy somehow queues the messages and processes only one event per 10 seconds.
Changing the temperature by 5° results in a delay of at least 7 minutes until homeassistant shows the right value. Zigbee2Mqtt processes the messages immediately.
I’m see this behavior with a cc2531 as well a EmberZNet (Sonoff ZBBridge) coordinator.

[homeassistant.log](https://github.com/zigpy/zigpy/files/5721010/homeassistant.log)
[zigbee2mqtt.log](https://github.com/zigpy/zigpy/files/5721013/zigbee2mqtt.log)

puddly:
Your device only seems to be sending the attribute reports every minute or so:

```Python
2020-12-20 12:49:22 DEBUG (MainThread) [homeassistant.components.zha.core.channels.base] [0xE1D4:1:0x0201]: Attribute report 'SiterwellThermostat'[occupied_heating_setpoint] = 550
2020-12-20 12:50:14 DEBUG (MainThread) [homeassistant.components.zha.core.channels.base] [0xE1D4:1:0x0201]: Attribute report 'SiterwellThermostat'[occupied_heating_setpoint] = 600
2020-12-20 12:51:07 DEBUG (MainThread) [homeassistant.components.zha.core.channels.base] [0xE1D4:1:0x0201]: Attribute report 'SiterwellThermostat'[occupied_heating_setpoint] = 650
2020-12-20 12:51:59 DEBUG (MainThread) [homeassistant.components.zha.core.channels.base] [0xE1D4:1:0x0201]: Attribute report 'SiterwellThermostat'[occupied_heating_setpoint] = 700
2020-12-20 12:52:52 DEBUG (MainThread) [homeassistant.components.zha.core.channels.base] [0xE1D4:1:0x0201]: Attribute report 'SiterwellThermostat'[occupied_heating_setpoint] = 750
2020-12-20 12:53:44 DEBUG (MainThread) [homeassistant.components.zha.core.channels.base] [0xE1D4:1:0x0201]: Attribute report 'SiterwellThermostat'[occupied_heating_setpoint] = 800
[...]

PR #622, opened 2021-01-15
Send default response for cluster requests.

xonestonex:
Send a default response if a response has been requested.

Tuya devices wait with a timeout for the response, until a next request is send.
See: https://github.com/zigpy/zigpy/issues/592

### DIFF
--- a/zigpy/zcl/__init__.py
+++ b/zigpy/zcl/__init__.py
@@ -192,6 +192,14 @@ class Cluster(util.ListenableMixin, util.CatchingTaskMixin, metaclass=Registry):
         if hdr.frame_control.is_cluster:
             self.handle_cluster_request(hdr.tsn, hdr.command_id, args)
             self.listener_event("cluster_command", hdr.tsn, hdr.command_id, args)
+            if not hdr.frame_control.disable_default_response:
+                self.debug("ZCL default response 0x%04x: %d", hdr.command_id, hdr.tsn)
+                schema = foundation.COMMANDS[foundation.Command.Default_Response][0]
+                self.create_catching_task(
+                    self.reply(
+                        True, foundation.Command.Default_Response, schema, hdr.command_id, tsn=hdr.tsn
+                    )
+                )
             return
         self.listener_event("general_command", hdr, args)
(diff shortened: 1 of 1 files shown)


Adminiuga:
1st -- Thank you for contribution.
But short version -- this would break a few things and cannot be accepted as is. In the short term, this fix should be implemented in zha-device-handlers for Tuya manufacturer specific cluster in `handle_cluster_request()`

Long version: I don't think you can/should send the default_response command/reply to all cluster requests with the `disable_default_response == False` IIRC you send the default_response to a cluster command, only if there's no "cluster specific" response to the received command. For example, when I send a "get_groups" command to the group cluster, I do not get the default_response, but I get the "get_groups_response" command instead.  So, at very least this code should check what kind of cluster it is server/client to know whether the received command is a server/cluster command and lookup if there's a response command and send the default response only if there's no response command defined. But even with that, I'm split on the best approach and ATM leaning toward the individual cluster decide what response to send.

Adminiuga:
> But even with that, I'm split on the best approach and ATM leaning toward the individual cluster decide what response to send.

By that I mean the "response" to the received command may come from:
1. `handle_cluster_request()` -- method, like time or OTA clusters
2. in certain cases IMO it would be better to let `cluster_command()` event listener to generate a reply and this approach would block such cases

xonestonex:
My problem is, that disable_default_response is not passed to handle_cluster_request(). So I don't know that I need to create a default response in the manufacturer specific cluster in zha-device-handlers.

Adminiuga:
Does the tuya ever sends a "command" with the default response disabled? Does it actually needs the "default response" or is there a cluster specific response command? i.e. how does tuya gateway responds to those commands?

For tuya, i guess we could always send that response, but I'd like to know if it has the specific response command or not and how the actual gateway responds.

In the long run, I think we should change the zigpy API and send `hdr` instance, instead of just the `hdr.tsn` to the `handle_cluster_request()` method.

xonestonex:
It looks like, that tuya always expects a default response for client commands and also sends a default response for server commands. 
I will close this PR now and implement the change in the manufacturer specific cluster in zha-device-handlers.
Thanks.

PR #625, opened 2021-01-16
Make Bool type an enum8

Adminiuga:
Make the `zigpy.types.named.Bool` an `enum8` type, to allow other than 0 & 1 value for the bool type, as some devices are overloading this type with other values, requiring more complex handling/work arounds


Adminiuga:
Helps with:
- https://github.com/zigpy/zha-device-handlers/issues/719
- https://github.com/zigpy/zha-device-handlers/pull/715
- https://github.com/zigpy/zha-device-handlers/issues/713

PR #632, opened 2021-01-20
Update README.md as CC253x with Z-Stack 3.0.x firmware working via zigpy-znp

Hedda:
@puddly OK to update zigpy README.md as CC2538/CC2531/CC2530 with Z-Stack 3.0.x firmware now working via zigpy-znp?


Adminiuga:
IMO Doesn't make sense duplicating the list of supported hardware in Zigpy's Readme and in the radio lib readme. Zigpy readme just should contain the links to the libs and libs readme should define the list of supported hardware.

MattWestb:
@Adminiuga I think its time moving some detailed info to the wiki and doing the main page more slimmed and the wiki more detailed with deeprt infos and guides.
And i knowing im not the person for writing the wiki :-((

Hedda:
> IMO Doesn't make sense duplicating the list of supported hardware in Zigpy's Readme and in the radio lib readme. Zigpy readme just should contain the links to the libs and libs readme should define the list of supported hardware.

IMHO, I would argue that first impressions are very important and we could probably assume that at least some new developers that stumble on zigpy will read its README.md file and perhaps even make a decision on whether to research further based on the information in it before digging deeper into its dependency on radio libraries for adapter specific code.

Hedda:
Closing as https://github.com/zigpy/zigpy/pull/705 makes this PR obsolete.

PR #633, opened 2021-01-20
Support wildcard manufacturer quirks

Adminiuga:
`CustomDevice` with a "model" only signature key now will match any manufacturer, assuming the rest of the signature matches.


TheJulianJES:
Does it make sense to also support wildcard model quirks?
For example: the many Hue lights could have manufacturer Philips and Signify and then match using the rest of the signature.

Adminiuga:
That's already was supported. If quirk has only `manufacturer` specified, then it would check all models for that manufacturer. Just when it was implemented, i never expected there to be the same "models" from different manufacturers.

PR #639, opened 2021-01-21
Refactor Cluster.handle_cluster_request() method

Adminiuga:
Change method signature, send the actual ZCL header to provide more context to the handler.
This PR is a result to the [comment](https://github.com/zigpy/zigpy/pull/622#issuecomment-761615829) and is aimed to provide more context to the command handler, like whether a default response should be suppressed or not.


Adminiuga:
This is a breaking change for `zha-device-handlers` and a separate PR will be submitted to address these changes.

PR #659, opened 2021-02-10
Do not catch `CancelledError` with `except Exception` in Python 3.7

puddly:
Ran into a bizarre failing unit test in zigpy-znp caused by `asyncio.CancelledError` being an `Exception` subclass in Python 3.7 but a `BaseException` subclass in Python 3.8 and beyond. This causes coroutines that normally handle exceptions (e.g. infinite loops) to never actually quit when cancelled.

For example, the following task will not stop running in Python 3.7, but will in Python 3.8 and 3.9:

```Python
import asyncio

async def task():
    while True:
        try:
            print("Running...")
            await asyncio.sleep(1)
        except Exception as e:
            print(f"Caught an exception: {e!r}")
        
        await asyncio.sleep(2)

async def main():
    t = asyncio.create_task(task())
    await asyncio.sleep(4)
    print("Stopping task")
    t.cancel()
    await asyncio.sleep(10)

if __name__ == "__main__":
    asyncio.run(main())
```

### DIFF
--- a/zigpy/device.py
+++ b/zigpy/device.py
@@ -139,6 +139,8 @@ class Device(zigpy.util.LocalLogMixin, zigpy.util.ListenableMixin):
                 continue
             try:
                 await ep.initialize()
+            except asyncio.CancelledError:
+                raise
             except Exception as exc:
                 self.warning("Endpoint %s initialization failure: %s", endpoint_id, exc)
                 break
--- a/zigpy/util.py
+++ b/zigpy/util.py
@@ -286,9 +286,7 @@ class CatchingTaskMixin(LocalLogMixin):
             return await target
         except exceptions:
             pass
-        except asyncio.CancelledError:
-            raise
-        except Exception:  # pylint: disable=broad-except
+        except (Exception, asyncio.CancelledError):  # pylint: disable=broad-except
             # Do not print the wrapper in the traceback
             frames = len(inspect.trace()) - 1
             exc_msg = traceback.format_exc(-frames)
--- a/zigpy/util.py
+++ b/zigpy/util.py
@@ -286,6 +286,8 @@ class CatchingTaskMixin(LocalLogMixin):
             return await target
         except exceptions:
             pass
+        except asyncio.CancelledError:
+            raise
         except Exception:  # pylint: disable=broad-except
             # Do not print the wrapper in the traceback
             frames = len(inspect.trace()) - 1
--- a/zigpy/appdb.py
+++ b/zigpy/appdb.py
@@ -92,6 +92,8 @@ class PersistingListener(zigpy.util.CatchingTaskMixin):
                     args,
                     str(exc),
                 )
+            except asyncio.CancelledError:
+                raise
             except Exception as ex:
                 LOGGER.error(
                     "Unexpected error while processing %s(%s): %s", cb_name, args, ex
--- a/zigpy/application.py
+++ b/zigpy/application.py
@@ -67,6 +67,8 @@ class ControllerApplication(zigpy.util.ListenableMixin, abc.ABC):
         if start_radio:
             try:
                 await app.startup(auto_form)
+            except asyncio.CancelledError:
+                raise
             except Exception:
                 LOGGER.error("Couldn't start application")
                 await app.pre_shutdown()
--- a/zigpy/endpoint.py
+++ b/zigpy/endpoint.py
@@ -59,6 +59,8 @@ class Endpoint(zigpy.util.LocalLogMixin, zigpy.util.ListenableMixin):
                 return
             elif sdr[0] != zdo_status.SUCCESS:
                 raise Exception("Failed to retrieve service descriptor: %s", sdr)
+        except asyncio.CancelledError:
+            raise
         except Exception:
             self.warning(
                 "Failed to discover endpoint_id %s", self.endpoint_id, exc_info=True


Adminiuga:
Wanna add coverage?

puddly:
Looking closer, maybe the Python 3.7 behavior is better in some cases?? Like `appdb.PersistingListener._worker` processing all of its tasks before shutting down or if startup fails in `ControllerApplication.new`, `pre_shutdown` will still be called?

Adminiuga:
We're not cancelling `appdb.PersistingListener._worker` from zigpy, are we? i.e. when it is cancelled, then it is already to late and to do anything and bailing out is probably fine.

Adminiuga:
btw, i've modified the code like
```
import asyncio

async def task():
    while True:
        try:
            print("Running...")
            await asyncio.sleep(1)
        except Exception as e:
            print(f"Caught an exception: {e!r}")

        await asyncio.sleep(1)

async def main():
    t = asyncio.create_task(task())
    await asyncio.sleep(4)
    print("Stopping task")
    t.cancel()
    print(f"Task stopped, sleeping for another 5s. Tasks: {asyncio.all_tasks()}")   
    await asyncio.sleep(5)
    print("done sleeping, let's check the task")
    await t

if __name__ == "__main__":
    asyncio.run(main())
```

and looks like the task actually stops on 3.7.3, 3.8.x and 3.9.1, but on 3.9.1 there are two exceptions:
```
Stopping task
Task stopped, sleeping for another 5s. Tasks: {<Task pending name='Task-2' coro=<task() running at /tmp/z.py:11> wait_for=<Future cancelled>>, <Task pending name='Task-1' coro=<main() running at /tmp/z.py:18> cb=[_run_until_complete_cb() at /usr/lib/python3.9/asyncio/base_events.py:184]>}
done sleeping, let's check the task
  File "/tmp/z.py", line 11, in task
    await asyncio.sleep(1)
  File "/usr/lib/python3.9/asyncio/tasks.py", line 654, in sleep
    return await future
asyncio.exceptions.CancelledError

During handling of the above exception, another exception occurred:

  File "/tmp/z.py", line 21, in main
    await t
asyncio.exceptions.CancelledError

During handling of the above exception, another exception occurred:

  File "/tmp/z.py", line 24, in <module>
    asyncio.run(main())
  File "/usr/lib/python3.9/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/usr/lib/python3.9/asyncio/base_events.py", line 642, in run_until_complete 
    return future.result()
asyncio.exceptions.CancelledError
```

Adminiuga:
the problem you might be running into, could be related to `aiosqlite`, it a test fails and you don't shutdown the persistent listener, then the test never exits, because aios1qlite thread never finishes, see https://github.com/omnilib/aiosqlite/pull/82

puddly:
I think that's just a mistake in my example, the outermost `await asyncio.sleep(1)` call is unnecessary. Removing it causes your modified code to still break on 3.7:

```
Running...
Running...
Running...
Running...
Stopping task
Task stopped, sleeping for another 5s. Tasks: {<Task pending coro=<task() running at <ipython-input-4-dcfbb2a7e421>:7> wait_for=<Future cancelled>>, <Task pending coro=<main() running at <ipython-input-4-dcfbb2a7e421>:16> cb=[_run_until_complete_cb() at /usr/local/opt/python@3.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/base_events.py:157]>}
Caught an exception: CancelledError()
Running...
Running...
Running...
Running...
Running...
done sleeping, let's check the task
Running...
Running...
Running...
Running...
```

Works as expected on Python 3.9:

```
Running...
Running...
Running...
Running...
Stopping task
Task stopped, sleeping for another 5s. Tasks: {<Task pending name='Task-16' coro=<task() running at <ipython-input-1-dcfbb2a7e421>:7> wait_for=<Future cancelled>>, <Task pending name='Task-15' coro=<main() running at <ipython-input-1-dcfbb2a7e421>:16> cb=[_run_until_complete_cb() at /usr/local/Cellar/python@3.8/3.8.6_1/Frameworks/Python.framework/Versions/3.8/lib/python3.8/asyncio/base_events.py:184]>}
done sleeping, let's check the task
---------------------------------------------------------------------------
CancelledError                            Traceback (most recent call last)
```

Adminiuga:
huh, interesting. Can confirm.

zigpy/device.py
```diff
@@ -139,6 +139,8 @@ async def _initialize(self):
                 continue
             try:
                 await ep.initialize()
+            except asyncio.CancelledError:
+                raise
```
Adminiuga, line 143: should it just log a warning and `break` instead like a regular exception? I don't think we directly cancel `endpoint.initialize()` but we may cancel `device.initialze()` should we guard there instead?

puddly: Hmm. I think there is a difference between `ep.initialize()` failing (i.e. when the device falls asleep and the radio library throws an exception due to the message being undeliverable) and `device.cancel_initialization()` being called, cancelling the initialization task. The former is common, but the latter seems like it shouldn't be swallowed and then `self.application.remove` be awaited, potentially never cancelling the task if that coroutine takes a long time.

Also, the device object removing itself from its `ControllerApplication` may not be a good idea?? It seems like in this specific case, it would be better for the application to schedule a device initialization in a dedicated task and then remove the device if the initialization fails, instead of the device having a method that does all of this on its own? The `device_init_failure` event is not used by ZHA or anything else and seems like a substitute for just letting an exception happen.

zigpy/util.py
```diff
@@ -286,6 +286,8 @@ async def catching_coro(
             return await target
         except exceptions:
             pass
+        except asyncio.CancelledError:
+            raise
```
Adminiuga, line 290: should we raise here or just handle the same way as the broad Exception? For catching coros that was the intention, not to raise any exceptions so you don't get "exception was never retrieved"

zigpy/device.py
Adminiuga: I guess my question is: while `device.initialize()` is  awaiting on `ep.initialize()` and `dev.initialize()` is cancelled, is ep.initialize() cancelled too or is it allowed to finish?

But instead, maybe instead of raising and catching a wide exception, we should catch `ZigbeeException` and `asyncio.Timeout` exceptions allowing anything else rise? and on failures, instead of throwing a `Exception` we should raise a more specific `ControllerException` ?

puddly: Hmm. I expect a task to immediately get cancelled when it's run unless it's doing something extremely critical when cleaning up.

By catching `ZigbeeException` and `asyncio.Timeout`, do you mean doing what currently happens, which is logging a warning and returning? To be honest, I think all of these `try` blocks that swallow exceptions and log them instead should be removed entirely. If whatever is running `get_node_descriptor` or `initialize` wants to itself swallow the exceptions, it should. Otherwise, let the coroutines work normally and leave the higher-level application controlling zigpy to deal with the errors however it wants.

zigpy/util.py
puddly: It doesn't appear that an error is logged when either the `catching_coro` task or an inner task is cancelled:

```Python
import asyncio


async def catching_coro(target, exceptions):
    try:
        return await target
    except exceptions:
        pass
    except asyncio.CancelledError:  # For Python 3.7
        raise
    except Exception as e:
        print(f"Caught exception: {e}")


async def busy():
    while True:
        print('Busy')
        await asyncio.sleep(0.5)

async def main():
    busy_task = asyncio.create_task(busy())
    catching_task = asyncio.create_task(catching_coro(busy_task, (RuntimeError,)))

    await asyncio.sleep(1)

    print('Cancelling')
    busy_task.cancel()
    catching_task.cancel()

    await asyncio.sleep(1)
    print('Done')


if __name__ == '__main__':
    asyncio.run(main())
```

puddly: Ignore that, it does get logged if you `await catching_task`. It doesn't if you catch the `CancelledError`.

Yeah, maybe here it does make sense to catch both here?

zigpy/util.py
```diff
@@ -286,6 +286,8 @@ async def catching_coro(
...
         except Exception:  # pylint: disable=broad-except
```
Adminiuga, line 291:
```suggestion
        except (Exception, asyncio.CancelledError):  # pylint: disable=broad-except
```

PR #656, opened 2021-02-01
[WIP] Add ZGP (continue of PR #519)

zoic21:

PR #669, opened 2021-02-20
Add support for bitfields to Structs

puddly:
Inspired by #668 (and #656), I'm going to try and add support for integer types of arbitrary bit lengths. My goal is to make specifying and working with bitfields no different from working with any other types:

```Python
class LogicalType(t.enum3):
    Coordinator = 0b000
    Router = 0b001
    EndDevice = 0b010
    Reserved3 = 0b011
    Reserved4 = 0b100
    Reserved5 = 0b101
    Reserved6 = 0b110
    Reserved7 = 0b111


class MACCapabilityFlags(t.bitmap8):
    AlternatePanCoordinator = 0b00000001
    FullFunctionDevice = 0b00000010
    MainsPowered = 0b00000100
    RxOnWhenIdle 0b00001000
    Reserved4 = 0b00010000
    Reserved5 = 0b00100000
    SecurityCapable = 0b01000000
    AllocateAddress = 0b10000000


class FrequencyBand(t.bitmap5):
    Freq868MHz = 0b00001
    Reserved1 = 0b00010
    Freq902MHz = 0b00100
    Freq2400MHz = 0b01000
    Reserved4 = 0b10000


class NodeDescriptor(t.Struct):
    # All bitfields will be assumed to packed into segments that start and end on a whole byte

    # byte 1
    logical_type: LogicalType
    complex_descriptor_available: t.uint1_t
    user_descriptor_available: t.uint1_t
    reserved: t.uint3_t

    # byte 2
    aps_flags: t.uint3_t
    frequency_band: FrequencyBand

    mac_capability_flags: MACCapabilityFlags
    manufacturer_code: t.uint16_t
    maximum_buffer_size: t.uint8_t
    maximum_incoming_transfer_size: t.uint16_t
    server_mask: t.uint16_t
    maximum_outgoing_transfer_size: t.uint16_t
    descriptor_capability_field: t.uint8_t
```

Unfortunately, this change will likely cause downstream breakage for code that relies on the existing public interface of these types of objects (i.e. the `byte1` and `byte2` attributes). Bitfields also have a bit of overlap with the existing bitmap types.

Thoughts?

In this current iteration, I just subclassed `t.Struct` and made a `t.BitStruct` and it works fine, but it introduces an extra layer of attributes and breaks code that relies on the properties returning `None` when the field is not populated. Having `t.Struct` deal with bit types directly is not that much more difficult but care must be taken to keep this from impacting performance.


puddly:
I've migrated both `zdo_t.Neighbors` and `zdo_t.NodeDescriptor`, which were the two most frequent structures in need of bitfields. I haven't looked in the ZCL yet but I think they are also in use in some security contexts that can be added later.

With the bitfields now properly exposed, a lot of properties (`NodeDescriptor.is_*`) are redundant but they're currently in use by ZHA. All of that information is available through the `logical_type` struct member and the `mac_capability_flags` bit enum.

Also, the node descriptor is stored in the database with a `packed` field. I wasn't sure how to handle database migrations so for now I've just written a simple compatibility layer that deals with the `packed` field, without incrementing the database schema version number.

At the moment everything passes unit tests but I'll be dogfooding this on my own network for a while. I will have to check if this will affect radio libraries that directly use zigpy's types.

puddly:
@Adminiuga this PR is basically done with the exception of how to handle database schema changes. Since fields like `Neighbors.packed` and `NodeDescriptor.byte1`/`byte2` are now properly expanded, the current DB schema for `neighbors` doesn't fit because it has a `packed` column. I see three possible solutions:

 1. Expand the schema to include more columns and write appropriate migrations. This is what I've currently implemented but are the individual columns ever even used? Possibly by other applications (do those exist?)?
 2. Downgrade the schema to just a blob `value` column, like `node_descriptors` and `relays`. Also requires migrations but simpler.
 3. Don't touch the DB and manually translate back and forth between `packed` and its respective bitfields in Python. I had this originally but it's extra translation code and database migrations can't be avoided forever.

Any ideas? I'm kinda leaning towards 2 for simplicity.

Adminiuga:
Currently schema changes are a major pita in terms is backwards compatibility.
We could create a new table for new node descriptor format? That would maintain backward compatibility at least for existing devices. Newer ones are going to have their descriptor only in the new table

Adminiuga:
And if the new table is a blob, then it is more future proof, but less human readable

puddly:
I'll have to look more closely at the `aiosqlite`/`sqlite3` transaction corner cases but the migration code [I have right now](https://github.com/puddly/zigpy/blob/puddly/bitstructs/zigpy/appdb.py#L624) migrated my database without any issues and doesn't do anything that esoteric that would break with older SQLite versions. I've added a few unit tests where migrations intentionally fail and no data is changed, as expected.

Adminiuga:
I mean, once the user upgrades and migrates to the new schema, if for any reason he decides to go back to previous versions, then the older zigpy version won't be able to parse the table and load data correctly.

puddly:
Ah. That is not the direction I had in mind when thinking of migrations.

For now, let's just leave the DB structure alone and convert back and forth between `packed` and the unpacked bitfields. Migrations can be looked at in a separate PR.

Adminiuga:
> Ah. That is not the direction I had in mind when thinking of migrations.

The only reason, is that with every HA release, a bunch of ConBee users would have issues, regardless if there were or were no changes in zigpy-deconz and they would try to go back to previous version. And I just don't want to deal with the fallout. Otherwise I don't have a strong preference.
And as a middle ground we could just use a new table name for node descriptors/neighbours, until there's a better schema update/rollback options. At least this should solve problem of reverting back to older vesions.

puddly:
Sounds good. I've added `new_*` tables to replace the in-place migration.

PR #700, opened 2021-03-19
Enabling firmware down grading from local file.

MattWestb:
Changing OTA so all not equal versions of firmware version is triggering OTA updating of devices.
For normal updating with online requests shall not being any problem.
If disabling online updates and putting one firmware in the "local OTA update folder" and the firmware version is not equal as in the device is the system saying update is true and sending the local file to the device then its requesting OTA updates.

I have trying up and downgrading one IKEA E27 opal 1000lm 6 times and its working great and is not updating the firmware then the version is the same.

As long the user is not configuring one local OTA folder and  / or not putting any valid OTA files for the device in the system its not making any problems.

Its making much easier testing changed functions in devices that can being firmware depending and also to restoring  "the default scenes" that IKEA is using for there 5 button  remotes and getting it working out of the box if putting the magic group in the light.

### DIFF
--- a/zigpy/ota/__init__.py
+++ b/zigpy/ota/__init__.py
@@ -57,7 +57,7 @@ class CachedImage:
         if self.key != ImageKey(manufacturer_id, img_type):
             return False
 
-        if ver >= self.version:
+        if ver == self.version:
             return False
 
         if (


MattWestb:
One problem can being if changing OTA feed and some files is dont have the same version and the system is doing one downgrading of some device that is changed between the feeds. 

But  the IKEA feed is normally the same but is then new test is coming and some weeks later and the production files is added and they is the same files a gen.

puddly:
Would this not enter into an upgrade-downgrade loop if you have two OTA files for the same device?

```yaml
ota_files/
  v123.ota
  v124.ota
```

It would downgrade to 123, then upgrade to 124, then downgrade to 123, etc.

Maybe adding another folder for IEEE-specific OTA files? This is even hackier but it could work:

```yaml
advanced_ota_files/
  00124b001cce3385_v123.ota  # prefixed with the device's IEEE
```

Zigpy would then delete/rename the OTA file after upgrading to prevent loops? Some OTA files (e.g. the IKEA `test` channel ones) have the same version number but different firmware.

Really, all of this stuff is a poor substitute for RPC...

Adminiuga:
Yes. Need a mechanism to allow downgrade for specific devices or version pinning for specific devices.

MattWestb:
I think all is possible to do if can doing coding well and have one bright head (and dont have non of that).

The upgrade loop i have not testing then its one advanced function that novice user shall not using and shall reading the wiki before start using this kind of functions.
I have not testing local versions and feed if getting upgrade loop.

But I 110% agree with you that better solution is needed in the long  run for not getting much issues then user having problems.

Cope many copies of one file i dont like, if you is have 20 IKEA GU10 CWS that you like downgrading you need doing 20 copies of the original OTA file.

Perhaps doing one list of IEEE and firmware that the device shall having if the system is finding one firmware online or local.
Like `./config/OTA_Files/Version.lst` and the local files in the directory and the `Version.ls` is only IEEE:Firmware version on one line / device that you like forcing the firmware version of if the system can finding it (local or in feed) or not doing anything and if the device is not in the list then using the highest file version that is matching.

MattWestb:
I have doing one extension of the Wiki for device OTA with [firmware downgrading](https://github.com/zigpy/zigpy/wiki/OTA-Device-Firmware-Updates#downgrading-your-device-firmware) and i think its enough for the moment for advanced user getting it done if they is needing doing firmware downgrading with some sort off control. 
Better is implanting configuration of it in HA config and / or HA services or the more advanced forced by IEEE as supposed.

Then i cant writing the advanced code that is needed and its for the moment is a working workaround i dont see any need for this PR so i closing it.

Issue #710, opened 2021-04-08
Old zigpy databases without foreign keys can have inconsistent `attributes` table state after restart


puddly:
A couple of users have complained that ZHA fails to start up after an update from `2021.3.x` to `2021.4.0` with the following error:

```Python
2021-04-08 20:58:25 ERROR (MainThread) [homeassistant.config_entries] Error setting up entry HubZ for zha
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/config_entries.py", line 249, in async_setup
    result = await component.async_setup_entry(hass, self)  # type: ignore
  File "/usr/src/homeassistant/homeassistant/components/zha/__init__.py", line 99, in async_setup_entry
    await zha_gateway.async_initialize()
  File "/usr/src/homeassistant/homeassistant/components/zha/core/gateway.py", line 157, in async_initialize
    self.application_controller = await app_controller_cls.new(
  File "/usr/local/lib/python3.8/site-packages/zigpy/application.py", line 64, in new
    await app._load_db()
  File "/usr/local/lib/python3.8/site-packages/zigpy/application.py", line 56, in _load_db
    await self._dblistener.load()
  File "/usr/local/lib/python3.8/site-packages/zigpy/appdb.py", line 476, in load
    await self._load_attributes()
  File "/usr/local/lib/python3.8/site-packages/zigpy/appdb.py", line 491, in _load_attributes
    dev = self._application.get_device(ieee)
  File "/usr/local/lib/python3.8/site-packages/zigpy/application.py", line 392, in get_device
    return self.devices[ieee]
KeyError: 90:fd:9f:ff:fe:71:71:70
```

Nothing interesting changed between those two HA releases so I believe this was "caused" by just restarting ZHA.

Here is one such database: [zigbee-attrs1.db.zip](https://github.com/zigpy/zigpy/files/6281140/zigbee-attrs1.db.zip)

Notably, there is in fact no device with that IEEE address, and yet a row for it appears in the `attributes` table:

```Console
$ sqlite3 ~/Downloads/zigbee-attrs1.db .dump | grep -i 90:fd:9f:ff:fe:71:71:70
INSERT INTO attributes VALUES('90:fd:9f:ff:fe:71:71:70',1,6,0,0);
INSERT INTO attributes VALUES('90:fd:9f:ff:fe:71:71:70',1,8,0,75);
```

[...]

PR #711, opened 2021-04-09
Migrate appdb to versioned tables to support zigpy upgrades/downgrades

puddly:
Every table is now versioned, allowing for zigpy to be upgraded/downgraded (for probably the wrong reasons) without breaking the database. Since we need to keep track of all past schemas, I've moved all table/index creation logic into dedicated `schema_v*.sql` files (that IMO are much easier to work with), which will allow for properly migrating from any past DB version.

The v5 schema migration ignores all non-`device` rows that violate schema constraints, which will "fix" #710 but investigating the root cause still needs to be done.


Adminiuga:
for https://github.com/zigpy/zigpy/issues/710 I think we should drop devices if we get any exception during the load and do some sanity checks maybe, like dropping the devices without a node descriptor, endpoints?

This looks good.

puddly:
I added foreign key constraints to the `neighbors` table (https://github.com/zigpy/zigpy/pull/711/commits/91a6c3bd245d0aee1897b15591280f5cb1603cc0) to remove the need to check for neighbor IEEEs existing in the database but it's causing one unit test to fail (and lock up pytest...).

Is this going to cause any issues (i.e. if a network has devices unknown to zigpy)?

Adminiuga:
Well, that one was omitted on purpose. So we do want to make sure that device reporting the lqi_mgmt is in the database, but what the device reports, e.g. an non existent ieee does not have to be in the database. We still create that neighbor entry, but that entry does not have a reference to a zigpy device

zigpy/appdb_schemas/schema_v5.sql
```diff
@@ -0,0 +1,168 @@
...
+-- attributes
+CREATE TABLE IF NOT EXISTS attributes_cache_v5 (
+    ieee ieee NOT NULL,
+    endpoint_id INTEGER NOT NULL,
+    cluster INTEGER NOT NULL,
+    attrid INTEGER NOT NULL,
+    value BLOB NOT NULL,
+
+    FOREIGN KEY(ieee, endpoint_id, cluster)
+        REFERENCES in_clusters_v5(ieee, endpoint_id, cluster)
```
Adminiuga, line 121: Some quirks report attributes on non existent (virtual) cluster ids, which breaks the foreign key. IIRC xiaomi contact sensor or could be other

puddly: Is it enough to relax the constraint to an `endpoints` foreign key?

Adminiuga: TBH, i don't remember. ATM i don't see why not.

puddly: It doesn't look like any quirks exist that add new endpoints, only ones remove existing ones for some reason (Gledopto and Aqara `CtrlNeutral`).

puddly:
Some more testing with @dmulcahey's database has yielded an unexpected problem: devices can still be functional *without a node descriptor*. They would be deleted during the DB v5 migration.

I see a few paths here:

 1. Allow the deletion to happen, since the device is only functional by coincidence. This is the easiest but will frustrate users, since they'll upgrade and devices will disappear (and likely will downgrade). It would be useful to pair this with some sort of UI feedback or prominent log entries.
 2. Relax the database constraints and allow devices with empty node descriptors to exist at runtime. I'm not a fan of this approach: any code using the node descriptor will have to handle `None` as a possible field value.
 3. Populate them with a dummy node descriptor. FFDs seem like they can't get into this state so we can just use RFDs and everything should "just work" after that.
 4. Keep track of attribute reporting status and binding state of every cluster (succeeded, not supported, failed), allowing zigpy to relentlessly initialize devices that are not fully setup every time they appear awake. Re-requesting the node descriptor (or just performing initialization from scratch in this case) will not be a difficult change but there are no guarantees that these corrupted entries correspond to online devices.

Thoughts?

MattWestb:
If trying configuration SED its little critical catching them "online".
Any thought implanting sending config to SED then they is doing check-in if they is having one ZB3 router as parent ?

I think the mechanism is not implanted in Zigpy /ZHA ("Reconfigure device" was of sync and was trying sending things that was always failing in 2021.4 so i think its not implanted) but shall working without problem if its being correct implanted.

As normal speculation without having the knowledge of coding from my side ;-((

puddly:
I've gone with approach 3 for now and have given them all a dummy node descriptor, along with logging an error message for each device at startup. If the devices are functional, I think it's best to wait a release or two before deleting them from user databases. This would give people a chance to fix things if they care about the device.

The lack of a valid node descriptor also causes an annoying issue: if the device re-joins the network with a new NWK address, the invalid node descriptor will cause the device's NWK in the database to never be updated, causing the device to be "unknown" to zigpy after a restart.

We could poll the node descriptor during startup to reduce the number of devices that need to be re-joined, but since dmulcahey's database is the only one that I've seen with this problem (and it affected a single device), I'm honestly not sure what the impact of this really is. Probably not worth the effort, especially since the device's NWK may be wrong.

> Any thought implanting sending config to SED then they is doing check-in if they is having one ZB3 router as parent ?

I haven't tested this but I have a feeling that some devices poll so infrequently that even their parent won't hold a message long enough. From zigpy's perspective we only can tell that a device is awake when it sends something back to the coordinator so @dmulcahey's idea of a a per-device request queue with smarter retrying will help a lot. Also, having more robust routing error resolution (falling back to IEEE addressing, scanning for changed NWKs, route re-discovery, radio-specific fixes, etc.) will need to be done as well, which again is per-device.

MattWestb:
If using pull control then its the SED doing check in to the coordinator (not to its parent / router) (and its not on pull that is being dome to its parent /router) that is setting the SED in fast pull and can sending data / commands and then its finished the coordinator is sending on stop fast pull and the SED is doing the normal sleep until next check in or if the router have cashed commands that is cached in the router between normal pulls.
The router is more or less transparent then its is only doing the normal cashing of commands.
The coordinator is holding all commands to the SED and delivering them then the SED is checking in so its can taking some time but its on secure delivery as long the mesh is not broke and the coordinator is working.

And great work with the DB up / downgrading !!

Adminiuga:
@dmulcahey what devices of yours does not have a node desriptor? AFAIK there only XBee modules which lack node descriptor and I don't remember if they provide a node descriptor via a quirk.

> The lack of a valid node descriptor also causes an annoying issue: if the device re-joins the network with a new NWK address, the invalid node descriptor will cause the device's NWK in the database to never be updated, 

@puddly how is this happening? 

IMO, the DB schema should allow a device without a node descriptor, but a node_descriptor without a valid device should not exist. 

Do you want to revert https://github.com/zigpy/zigpy/pull/520 ? It would update Node Descriptor on devices, if there's an incoming message and node descriptor is missing, but it had a problem when a brand new device joins, in certain cases it could request the node descriptor twice

puddly:
Here's the device info:

```Python
{
  'manufacturer': 'CentraLite',
  'model': '3320-L',
  #'node_desc': None,  # Not in the database
  'endpoints': {
    1: {
      'profile_id': 260,
      'device_type': <DeviceType.IAS_ZONE: 1026> ,
      'input_clusters': [0, 1, 3, 32, 1026, 1280, 2821],
      'output_clusters': [25]
    },
    2: {
      'profile_id': 49887,
      'device_type': 12,
      'input_clusters': [0, 1, 3, 2821, 64527],
      'output_clusters': [3]
    }
  }
}
```

And as SQL:

```SQL
INSERT INTO devices VALUES('IEEE_ADDR',50382,2);
INSERT INTO endpoints VALUES('IEEE_ADDR',1,260,1026,1);
INSERT INTO endpoints VALUES('IEEE_ADDR',2,49887,12,1);
INSERT INTO clusters VALUES('IEEE_ADDR',1,0);
INSERT INTO clusters VALUES('IEEE_ADDR',1,1);
INSERT INTO clusters VALUES('IEEE_ADDR',1,3);
INSERT INTO clusters VALUES('IEEE_ADDR',1,32);
INSERT INTO clusters VALUES('IEEE_ADDR',1,1026);
INSERT INTO clusters VALUES('IEEE_ADDR',1,1280);
INSERT INTO clusters VALUES('IEEE_ADDR',1,2821);
INSERT INTO clusters VALUES('IEEE_ADDR',2,0);
INSERT INTO clusters VALUES('IEEE_ADDR',2,1);
INSERT INTO clusters VALUES('IEEE_ADDR',2,3);
INSERT INTO clusters VALUES('IEEE_ADDR',2,2821);
INSERT INTO clusters VALUES('IEEE_ADDR',2,64527);
INSERT INTO output_clusters VALUES('IEEE_ADDR',1,25);
INSERT INTO output_clusters VALUES('IEEE_ADDR',2,3);
INSERT INTO attributes VALUES('IEEE_ADDR',1,0,4,'CentraLite');
INSERT INTO attributes VALUES('IEEE_ADDR',1,0,5,'3320-L');
INSERT INTO attributes VALUES('IEEE_ADDR',1,1280,1,21);
INSERT INTO attributes VALUES('IEEE_ADDR',1,1,49,2);
INSERT INTO attributes VALUES('IEEE_ADDR',1,1,51,1);
INSERT INTO attributes VALUES('IEEE_ADDR',1,0,7,3);
INSERT INTO attributes VALUES('IEEE_ADDR',1,1280,2,36);
INSERT INTO attributes VALUES('IEEE_ADDR',1,1280,0,1);
INSERT INTO attributes VALUES('IEEE_ADDR',2,0,7,3);
INSERT INTO attributes VALUES('IEEE_ADDR',1,1,62,0);
INSERT INTO attributes VALUES('IEEE_ADDR',1,1,32,28);
INSERT INTO attributes VALUES('IEEE_ADDR',1,1,33,200);
INSERT INTO attributes VALUES('IEEE_ADDR',1,1026,0,1917);
```

It should be a "just works" device, but the database is somehow missing the row for the node descriptor for this device. It's not corrupted either.

> how is this happening?

I haven't run into the issue myself. I believe this line will prevent the `raw_device_initialized` signal handler from saving the `Device` object with the updated NWK address: https://github.com/zigpy/zigpy/blob/103ac0052cd155b861fa9a9a1d46dc88cafcef51/zigpy/appdb.py#L402

Adminiuga:
I have quite a few centralite 3320-L devices and they definitely support Node Descriptor:

```yaml
{
  "node_descriptor": "NodeDescriptor(byte1=2, byte2=64, mac_capability_flags=128, manufacturer_code=49887, maximum_buffer_size=82, maximum_incoming_transfer_size=82, server_mask=0, maximum_outgoing_transfer_size=82, descriptor_capability_field=0, *allocate_address=True, *complex_descriptor_available=False, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False, *is_valid=True, *logical_type=<LogicalType.EndDevice: 2>, *user_descriptor_available=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0402",
      "in_clusters": ["0x0000", "0x0001", "0x0003", "0x0020", "0x0402", "0x0500", "0x0b05"],
      "out_clusters": ["0x0019"]
    },
    "2": {
      "profile_id": 49887,
      "device_type": "0x000c",
      "in_clusters": ["0x0000", "0x0003", "0x0b05", "0xfc0f"],
      "out_clusters": ["0x0003"]
    }
  },
  "manufacturer": "CentraLite",
  "model": "3320-L",
  "class": "zhaquirks.centralite.ias.CentraLiteIASSensor"
}
````

Node Descriptor is a mandatory field in zigbee and I haven't seen devices not supporting it, except XBee modules. We should not save devices without it, but should figure out why the node descriptor was not migrated. Since https://github.com/zigpy/zigpy/pull/520 we consider device not fully discovered without a node descriptor and don't save it.

**Comment by puddly** [comment]
This device is from @dmulcahey's current `zigbee.db` file. The database's `user_version` is `3` and it has no `_v4` tables so the node descriptor just was never inserted into the database in the first place, either because the device is old and was persisted improperly during joining with an older version of zigpy, or because zigpy currently has a bug that causes the node descriptor to not be persisted in some strange circumstances.

In either case, there are likely more users who are using devices with missing node descriptors. I don't think deleting these devices is a good solution.

**Comment by Adminiuga** [comment]
if we don't have the node descriptor for the device, then it should be kept as None, and requested again next time, when the  device send an inbound packet, i.e. revert #520, We should update NWK correctly though. 
If you use the fake node descriptor, is it going to be written to the database next time the device changes the NWK?

**Comment by puddly** [comment]
> if we don't have the node descriptor for the device, then it should be kept as None, and requested again next time, when the device send an inbound packet, i.e. revert #520, We should update NWK correctly though.

Sounds good.

Is there any case in which the node descriptor is not `is_valid` and *not* `None`? Really, `Struct.is_valid` exists only for `node_desc`, so it seems like it can be removed entirely. The node descriptor is deserialized from a bytestring when loaded from the database or from an over-the-air packet, so I can't think of any instance for which `NodeDescriptor.deserialize(...)[0].is_valid` is not `True`.

> If you use the fake node descriptor, is it going to be written to the database next time the device changes the NWK?

No, `PersistingListener._save_node_descriptor` in this PR [does nothing](https://github.com/zigpy/zigpy/pull/711/files#diff-4467528e1dd4e41467c02e93fefd5b85d0006e42c46349ea1032bf2720a9da5dR338-R345) if the device's node descriptor is the fake node descriptor.

**Comment by puddly** [comment]
I've attached another user report of missing node descriptors. This time it's a full log: [no_node_desc_log.txt](https://github.com/zigpy/zigpy/files/6504637/no_node_desc_log.txt).

Here's an excerpt. This is a common joining pattern for IKEA devices:

```python
# Device is sent the transport key
2021-05-18 14:23:40 DEBUG (MainThread) [bellows.zigbee.application] Received trustCenterJoinHandler frame with [0x9d54, aa:bb:cc:dd:00:11:22:33, <EmberDeviceUpdate.STANDARD_SECURITY_UNSECURED_JOIN: 1>, <EmberJoinDecision.USE_PRECONFIGURED_KEY: 0>, 0x3a2c]
2021-05-18 14:23:40 INFO (MainThread) [zigpy.application] Device 0x9d54 (aa:bb:cc:dd:00:11:22:33) joined the network
2021-05-18 14:23:40 DEBUG (MainThread) [bellows.ezsp.protocol] Send command findKeyTableEntry: (aa:bb:cc:dd:00:11:22:33, True)

# Zigpy immediately begins sending requests. These almost never succeed.
2021-05-18 14:23:40 INFO (MainThread) [zigpy.device] [0x9d54] Requesting 'Node Descriptor'
2021-05-18 14:23:40 DEBUG (MainThread) [zigpy.util] Tries remaining: 2
2021-05-18 14:23:40 DEBUG (MainThread) [zigpy.device] [0x9d54] Extending timeout for 0xea request
2021-05-18 14:23:40 DEBUG (MainThread) [bellows.zigbee.application] Extending timeout for aa:bb:cc:dd:00:11:22:33/0x9d54
2021-05-18 14:23:40 DEBUG (MainThread) [bellows.ezsp.protocol] Send command setExtendedTimeout: (aa:bb:cc:dd:00:11:22:33, True)
2021-05-18 14:23:40 DEBUG (MainThread) [bellows.ezsp.protocol] Send command sendUnicast: (<EmberOutgoingMessageType.OUTGOING_DIRECT: 0>, 0x9D54, EmberApsFrame(profileId=0, clusterId=2, sourceEndpoint=0, destinationEndpoint=0, options=<EmberApsOption.APS_OPTION_ENABLE_ROUTE_DISCOVERY|APS_OPTION_RETRY: 320>, groupId=0, sequence=234), 235, b'\xeaT\x9d')

# Device is sent the key again, four seconds later
2021-05-18 14:23:44 DEBUG (MainThread) [bellows.zigbee.application] Received trustCenterJoinHandler frame with [0x9d54, aa:bb:cc:dd:00:11:22:33, <EmberDeviceUpdate.STANDARD_SECURITY_UNSECURED_JOIN: 1>, <EmberJoinDecision.USE_PRECONFIGURED_KEY: 0>, 0x3a2c]
2021-05-18 14:23:44 INFO (MainThread) [zigpy.application] Device 0x9d54 (aa:bb:cc:dd:00:11:22:33) joined the network

# Zigpy doesn't want to re-initialize it
2021-05-18 14:23:44 DEBUG (MainThread) [zigpy.application] Skip initialization for existing device aa:bb:cc:dd:00:11:22:33
2021-05-18 14:23:44 DEBUG (MainThread) [bellows.ezsp.protocol] Send command findKeyTableEntry: (aa:bb:cc:dd:00:11:22:33, True)
2021-05-18 14:23:44 DEBUG (MainThread) [bellows.zigbee.application] Received incomingSenderEui64Handler frame with [aa:bb:cc:dd:00:11:22:33]
2021-05-18 14:23:44 DEBUG (MainThread) [bellows.zigbee.application] Received incomingMessageHandler frame with [<EmberIncomingMessageType.INCOMING_BROADCAST: 4>, EmberApsFrame(profileId=0, clusterId=19, sourceEndpoint=0, destinationEndpoint=0, options=<EmberApsOption.APS_OPTION_SOURCE_EUI64: 1024>, groupId=0, sequence=63), 33, -83, 0x9d54, 255, 255, b'\x81T\x9d\x33\x22\x11\x00\xdd\xcc\xbb\xaa\x8e']

# The device immediately announces itself this time but four seconds after the initial join
2021-05-18 14:23:44 INFO (MainThread) [bellows.zigbee.application] ZDO Device announce: 0x9d54, aa:bb:cc:dd:00:11:22:33
2021-05-18 14:23:44 INFO (MainThread) [zigpy.application] Device 0x9d54 (aa:bb:cc:dd:00:11:22:33) joined the network
2021-05-18 14:23:44 DEBUG (MainThread) [zigpy.application] Skip initialization for existing device aa:bb:cc:dd:00:11:22:33
2021-05-18 14:23:44 DEBUG (MainThread) [zigpy.zdo] [0x9d54:zdo] ZDO request ZDOCmd.Device_annce: [0x9D54, aa:bb:cc:dd:00:11:22:33, 142]

# The first node descriptor request fails here
2021-05-18 14:23:45 DEBUG (MainThread) [zigpy.device] [0x9d54] Delivery error for seq # 0xea, on endpoint id 0 cluster 0x0002: message send failure
2021-05-18 14:23:45 DEBUG (MainThread) [zigpy.device] [0x9d54] Extending timeout for 0xf1 request
2021-05-18 14:23:45 DEBUG (MainThread) [bellows.zigbee.application] Extending timeout for aa:bb:cc:dd:00:11:22:33/0x9d54
2021-05-18 14:23:45 DEBUG (MainThread) [bellows.ezsp.protocol] Send command setExtendedTimeout: (aa:bb:cc:dd:00:11:22:33, True)

# The second node descriptor request is sent almost two seconds after the device joined for real
2021-05-18 14:23:45 DEBUG (MainThread) [bellows.ezsp.protocol] Send command sendUnicast: (<EmberOutgoingMessageType.OUTGOING_DIRECT: 0>, 0x9D54, EmberApsFrame(profileId=0, clusterId=2, sourceEndpoint=0, destinationEndpoint=0, options=<EmberApsOption.APS_OPTION_ENABLE_ROUTE_DISCOVERY|APS_OPTION_RETRY: 320>, groupId=0, sequence=241), 242, b'\xf1T\x9d')
2021-05-18 14:23:46 DEBUG (MainThread) [bellows.zigbee.application] Received incomingMessageHandler frame with [<EmberIncomingMessageType.INCOMING_UNICAST: 0>, EmberApsFrame(profileId=0, clusterId=2, sourceEndpoint=0, destinationEndpoint=0, options=<EmberApsOption.APS_OPTION_ENABLE_ROUTE_DISCOVERY|APS_OPTION_RETRY: 320>, groupId=0, sequence=64), 132, -62, 0x9d54, 255, 255, b'\x01\x00\x00']
2021-05-18 14:23:46 DEBUG (MainThread) [zigpy.zdo] [0x9d54:zdo] ZDO request ZDOCmd.Node_Desc_req: [0x0000]
2021-05-18 14:23:46 DEBUG (MainThread) [zigpy.zdo] [0x9d54:zdo] Unsupported ZDO request:ZDOCmd.Node_Desc_req
2021-05-18 14:23:48 DEBUG (MainThread) [bellows.zigbee.application] Received incomingRouteRecordHandler frame with [0x9d54, aa:bb:cc:dd:00:11:22:33, 132, -62, [0x3a2c, 0xe205]]
2021-05-18 14:23:48 DEBUG (MainThread) [bellows.zigbee.application] Processing route record request: (0x9d54, aa:bb:cc:dd:00:11:22:33, 132, -62, [0x3a2c, 0xe205])

# The radio seems to struggle to find a way to contact the device after it re-joined
2021-05-18 14:23:48 DEBUG (MainThread) [bellows.zigbee.application] Received incomingRouteErrorHandler frame with [<EmberStatus.MANY_TO_ONE_ROUTE_FAILURE: 170>, 0x9d54]
2021-05-18 14:23:48 DEBUG (MainThread) [bellows.zigbee.application] Processing route error: status=EmberStatus.MANY_TO_ONE_ROUTE_FAILURE, nwk=0x9d54
2021-05-18 14:23:49 DEBUG (MainThread) [bellows.zigbee.application] Received incomingRouteRecordHandler frame with [0x9d54, aa:bb:cc:dd:00:11:22:33, 132, -62, [0x3a2c, 0xe205]]
2021-05-18 14:23:49 DEBUG (MainThread) [bellows.zigbee.application] Processing route record request: (0x9d54, aa:bb:cc:dd:00:11:22:33, 132, -62, [0x3a2c, 0xe205])
2021-05-18 14:23:50 DEBUG (MainThread) [bellows.zigbee.application] Received incomingMessageHandler frame with [<EmberIncomingMessageType.INCOMING_UNICAST: 0>, EmberApsFrame(profileId=0, clusterId=2, sourceEndpoint=0, destinationEndpoint=0, options=<EmberApsOption.APS_OPTION_ENABLE_ROUTE_DISCOVERY|APS_OPTION_RETRY: 320>, groupId=0, sequence=64), 132, -62, 0x9d54, 255, 255, b'\x01\x00\x00']
2021-05-18 14:23:50 DEBUG (MainThread) [zigpy.zdo] [0x9d54:zdo] ZDO request ZDOCmd.Node_Desc_req: [0x0000]
2021-05-18 14:23:50 DEBUG (MainThread) [zigpy.zdo] [0x9d54:zdo] Unsupported ZDO request:ZDOCmd.Node_Desc_req
2021-05-18 14:23:50 DEBUG (MainThread) [bellows.zigbee.application] Received zigbeeKeyEstablishmentHandler frame with [aa:bb:cc:dd:00:11:22:33, <EmberKeyStatus.TC_RESPONDED_TO_KEY_REQUEST: 6>]
2021-05-18 14:23:50 DEBUG (MainThread) [bellows.zigbee.application] Received incomingRouteRecordHandler frame with [0x9d54, aa:bb:cc:dd:00:11:22:33, 132, -62, [0x3a2c, 0xe205]]
2021-05-18 14:23:50 DEBUG (MainThread) [bellows.zigbee.application] Processing route record request: (0x9d54, aa:bb:cc:dd:00:11:22:33, 132, -62, [0x3a2c, 0xe205])
2021-05-18 14:23:50 DEBUG (MainThread) [bellows.zigbee.application] Received incomingRouteErrorHandler frame with [<EmberStatus.SOURCE_ROUTE_FAILURE: 169>, 0x9d54]
2021-05-18 14:23:50 DEBUG (MainThread) [bellows.zigbee.application] Processing route error: status=EmberStatus.SOURCE_ROUTE_FAILURE, nwk=0x9d54
2021-05-18 14:23:50 DEBUG (MainThread) [bellows.zigbee.application] Received incomingRouteErrorHandler frame with [<EmberStatus.MANY_TO_ONE_ROUTE_FAILURE: 170>, 0x9d54]
2021-05-18 14:23:50 DEBUG (MainThread) [bellows.zigbee.application] Processing route error: status=EmberStatus.MANY_TO_ONE_ROUTE_FAILURE, nwk=0x9d54

# Second node descriptor request fails
2021-05-18 14:23:50 DEBUG (MainThread) [zigpy.device] [0x9d54] Delivery error for seq # 0xf1, on endpoint id 0 cluster 0x0002: message send failure
2021-05-18 14:23:50 WARNING (MainThread) [zigpy.device] [0x9d54] Requesting Node Descriptor failed
  File "/usr/local/lib/python3.8/site-packages/zigpy/device.py", line 102, in get_node_descriptor
    status, _, node_desc = await self.zdo.Node_Desc_req(
  File "/usr/local/lib/python3.8/site-packages/zigpy/util.py", line 110, in retry
    r = await func()
  File "/usr/local/lib/python3.8/site-packages/zigpy/device.py", line 216, in request
    raise zigpy.exceptions.DeliveryError(
zigpy.exceptions.DeliveryError: [0x9d54:0:0x0002]: Message send failure

# Initialization continues???
2021-05-18 14:23:50 INFO (MainThread) [zigpy.device] [0x9d54] Discovering endpoints
```

Some thoughts:

 1. Zigpy needs to cancel initialization when a device leaves or re-joins and start again. Retrying sort of works but for end-devices it sometimes takes too long for the retrying delay to elapse.
 2. Routing errors seem to be common immediately after a device joins. Is there any way in bellows to proactively request a route immediately after a device joins? Or perhaps delay requests as long as a device is unreachable so that the retry delay timers synchronize once a single request succeeds?
 3. We need to better handle the two "stages" that some radios expose: in zigpy-znp I start a five second timer when the radio sends a trust center join indication. If a ZDO device announcement is received before the timer expires, `handle_join` is called immediately. Otherwise, it'll be called after the five seconds elapse.
    Perhaps it is better to always wait until a ZDO device announcement occurs? Are there any devices that do not send this message?
 4. Grabbing the node descriptor and other basic info is extremely important. Maybe we need to have it retry continuously and reject the device if it doesn't work after like 30 seconds? Are there devices that don't respond *at all* to node descriptor requests?

Adminiuga:
Yeah, this is a bug. We should not save the device which does not have a node descriptor or for which we failed to get a node descriptor. I think it is allowed to slide, because we used to request the node descriptor next time we'd hear from the device, which we don't do anymore, so if getting node descriptor fails, we should abort. 

The question I have: if we're initializing the devie and new `handle_join()` is issued? should we cancell the current init and start a new one?

> `# Device is sent the key again, four seconds later`
> ...
> `# Zigpy doesn't want to re-initialize it`

+1, we should cancel and start a new cycle

>Zigpy needs to cancel initialization when a device leaves or re-joins and start again. Retrying sort of works but for end-devices it sometimes takes too long for the retrying delay to elapse.
 +1

> Routing errors seem to be common immediately after a device joins. Is there any way in bellows to proactively request a route immediately after a device joins? Or perhaps delay requests as long as a device is unreachable so that the retry delay timers synchronize once a single request succeeds?

I need to check if we could send a many to one routing request. Do you know if this environment has source routing enabled?

> We need to better handle the two "stages" that some radios expose: in zigpy-znp I start a five second timer when the radio sends a trust center join indication. If a ZDO device announcement is received before the timer expires, handle_join is called immediately. Otherwise, it'll be called after the five seconds elapse (or cancelled if the device re-joins).
Perhaps it is better to always wait until a ZDO device announcement occurs? Are there any devices that do not send this message?

I'm only aware of Jasco HA1.2 switches which don't send device announcement for some reason. I think we could wait for the device announcement and start initialization then. May work better for ZB3 devices till they complete the querying coordinator.

> Grabbing the node descriptor and other basic info is extremely important. Maybe we need to have it retry continuously and reject the device if it doesn't work after like 30 seconds? Are there devices that don't respond at all to node descriptor requests?

Yes, it is important info and we should not save device without the node descriptor. I guess we could try continuously, but IMO if the device does not respond to 3 request it is unlikely to respond.

puddly:
> The question I have: if we're initializing the devie and new handle_join() is issued? should we cancell the current init and start a new one?

Yeah, that seems easiest. The current initialization is sort of reentrant so it should just start where it left off (if it did anything the first time). I will have to check how `asyncio.CancelledError` propagates through the various request functions that aren't themselves coroutines (e.g. https://github.com/zigpy/zigpy/blob/dev/zigpy/zdo/__init__.py#L41).

> Do you know if this environment has source routing enabled?

Doesn't look like it, it's a plug and play setup. `Set concentrator type:` isn't in the debug log either.

> I think we could wait for the device announcement and start initialization then. May work better for ZB3 devices till they complete the querying coordinator.

If cancelling the device initialization task can propagate and cancel all of the underlying radio library requests, we should be able to start initialization immediately every time a device joins or announces. That seems like it would handle both cases.

I'll have to test how this behaves with underpowered radios like the CC2531, which have a small transmit buffer and the software limiting may not work play well with sending a bunch of requests like this and cancelling them.

MattWestb:
The request and verifying of new TC link key is one ZB3 function and if its successful (that it shall being) the joining device is sending on ZDO with permit join with payload 180 seconds.

Before the ZDO is sent the devices is not complete in the network and cant answer commands from the coordinator.

So if receiving one device discretion request from the device it is starting the link exchange and shall not being bombed with commands before its sending the ZDO PJ 180. 

Look in 3.3.3 Requesting a New Link Key after Joining 
[an1233-zigbee-security.pdf](https://github.com/zigpy/zigpy/files/6506321/an1233-zigbee-security.pdf)

dmulcahey:
most IKEA devices seem to always join 2x with a nwk id change at least from what I have seen…

MattWestb:
Only the old motion sensor and the old CWS1 (RGBW 600 Lm) is still on ZLL but still on Zigbee PRO stack so all other is trying getting one new TCLK.

EmberZNet version: 6.0 is using  EZSP protocol v6 and is Zigbee PRO that is one must for Zigbee 3 but i dont knowing if its R21, R20 or R19 but i think it possible doing TCLK exchange with it.

I think the safest testing no ZB3 joining is using on CC-2531 with HA1.2 firmware.
If i remember is IKEA ZB3 device only one joining then they is asking of node discretion of the coordinator and its have one flag that is saying its not supporting TCLK changing then using on CC-2531.

Or using one no ZB3 Xiaomi device that is HA 1.X.

MattWestb:
From the release notes "EmberZNet SDK 6.0.3.0 GA":
`Green Power updates for Zigbee 3.0 and Green Power Proxy support`
So if Zigpy / bellows is doing it right its forming the network as Zigbee 3.

MattWestb:
It shall being possible setting TCLK request or not in the EZSP stack:

>On EZSP-based NCP platforms, the application will not be asked directly to make a decision about whether or not to provide link keys. So in order to allow trust center link keys, EzspPolicyId EZSP_TC_KEY_REQUEST_POLICY must be set to allow EZSP_ALLOW_TC_KEY_REQUESTS. For Application link keys use EZSP_APP_KEY_REQUEST_POLICY.

puddly:
Let's move the device initialization discussion over to https://github.com/zigpy/zigpy/pull/744. The appdb changes are for the most part independent.

puddly:
For this missing node descriptor problem, do you want to keep `self.node_desc = zdo_t.NodeDescriptor()` as we had before or explicitly use `self.node_desc = None`? Right now I have the "dummy" valid node descriptor in this PR and it's functional but it we're already introducing breaking changes, it may be easier to be explicit and go with `None`.

This would also be a good opportunity to remove all of the unnecessary `is_*` properties, since the application can then directly use `NodeDescriptor.logical_type`:

```Python
    @property
    def is_end_device(self) -> bool | None:
        if self.logical_type is None:
            return None

        return self.logical_type == LogicalType.EndDevice
```

Adminiuga:
Need to update all the apps first, before removing this. Pretty sure zha component and bellows stills uses is_* properties.
On the other hand, i kind like is_properties. Without those, the app needs to import Logical Device Type enum and then do comparison in the code vs device.node_desc and device.is_end_device

puddly:
Yeah, the `is_` properties do make it easier to use. I guess my real issue with them is that #668 now makes the node descriptor's string representation super verbose, but we can probably revert that PR once the few remaining bitfield types that don't actually use bitfields are expanded.

Let's just leave them alone for now and deal with the `None` vs `zdo_t.NodeDescriptor()` choice, it's I think the last issue with this PR. Once it's merged we can start extending the database schema to properly support ZGP devices, maybe add a new column to each endpoint/cluster for ZHA initialization status (so ZHA isn't stuck with half-configured devices), stuff like that.

Adminiuga:
I'm fine with node descriptor to be None, just need find all instances where it is being used and correct. Probably in bellows in zha component

Adminiuga:
Is this ready? Are you planning any more changes?

puddly:
It's good to go.

PR #744, opened 2021-05-19
Improved device initialization

puddly:
This is a continuation of the discussion started with https://github.com/zigpy/zigpy/pull/711#issuecomment-843597383 and my attempt at implementing some of the changes.

It's currently a WIP and no unit tests have been updated. That being said, IKEA remotes and bulbs feel like they join much more quickly.

Changes:
 - No errors are caught during device initialization, the whole process has to work.
 - Device initialization is cancelled and re-started every time a device joins or announces itself.
   - Initialization more or less starts where it left off after every restart and can safely be called on already-initialized devices, in case they have no node descriptor or are missing some sort of information. This is done at startup and after any message is received from a partially-initialized device.
 - A new `status_changed` device event that's sent after each stage during initialization. Could be used for more ZHA progress.
 - NWK changes are explicitly saved to the database in a new event.

TODO:
 - Some sort of event to signal that initialization has failed. That way, the user will know that the device needs to be re-joined (maybe with a message telling them to keep the device awake if it's a sensor).
 - See if delaying requests until after the device announces itself (or a few seconds have elapsed since the TC indication) is a better approach.

### DIFF
--- a/zigpy/application.py
+++ b/zigpy/application.py
@@ -157,6 +157,9 @@ class ControllerApplication(zigpy.util.ListenableMixin, abc.ABC):
         if not dev:
             LOGGER.debug("Device not found for removal: %s", ieee)
             return
+
+        dev.cancel_initialization()
+
         LOGGER.info("Removing device 0x%04x (%s)", dev.nwk, ieee)
         asyncio.create_task(self._remove_device(dev))
         if dev.node_desc.is_valid and dev.node_desc.is_end_device:
--- a/zigpy/application.py
+++ b/zigpy/application.py
@@ -212,39 +212,62 @@ class ControllerApplication(zigpy.util.ListenableMixin, abc.ABC):
             "handle_message", sender, profile, cluster, src_ep, dst_ep, message
         )
 
-        if not sender.is_initialized:
+        if sender.is_initialized:
+            return sender.handle_message(
+                profile,
+                cluster,
+                src_ep,
+                dst_ep,
+                message,
+                dst_addressing=dst_addressing,
+            )
+
+        LOGGER.debug(
+            "Received frame on uninitialized device %s"
+            " from ep %s to ep %s, cluster %s: %r",
+            sender,
+            src_ep,
+            dst_ep,
+            cluster,
+            message,
+        )
+
+        if (
+            dst_ep == 0
+            or sender.all_endpoints_init
+            or (
+                sender.has_non_zdo_endpoints
+                and cluster == zigpy.zcl.clusters.general.Basic.cluster_id
+            )
+        ):
+            # Allow the following responses:
+            #  - any ZDO
+            #  - ZCL if endpoints are initialized
+            #  - ZCL from Basic cluster if endpoints are initializing
+
             if not sender.initializing:
                 sender.schedule_initialize()
 
-            if not sender.has_node_descriptor and dst_ep != 0:
-                # only allow ZDO responses while initializing device
-                LOGGER.debug(
-                    "Received frame on uninitialized device %s for endpoint: %s",
-                    sender,
-                    dst_ep,
-                )
-                zigpy.quirks.handle_message_from_uninitialized_sender(
-                    sender, profile, cluster, src_ep, dst_ep, message
-                )
-                return
-            elif dst_ep != 0 and cluster != 0:
-                # only allow access to basic cluster while initializing endpoints
-                LOGGER.debug(
-                    "Received frame on uninitialized device %s endpoint %s for cluster: %s",
-                    sender,
-                    dst_ep,
-                    cluster,
-                )
-                return
+            return sender.handle_message(
+                profile,
+                cluster,
+                src_ep,
+                dst_ep,
+                message,
+                dst_addressing=dst_addressing,
+            )
 
-        return sender.handle_message(
-            profile,
-            cluster,
-            src_ep,
-            dst_ep,
-            message,
-            dst_addressing=dst_addressing,
+        # Give quirks a chance to fast-initialize the device (at the moment only Xiaomi)
+        zigpy.quirks.handle_message_from_uninitialized_sender(
+            sender, profile, cluster, src_ep, dst_ep, message
         )
+
+        # Reload the sender device object, in it was replaced by the quirk
+        sender = self.get_device(ieee=sender.ieee)
+
+        # If the quirk did not fast-initialize the device, start initialization
+        if not sender.initializing and not sender.is_initialized:
+            sender.schedule_initialize()
 
     def handle_join(self, nwk: t.NWK, ieee: t.EUI64, parent_nwk: t.NWK) -> None:
         """
--- a/zigpy/application.py
+++ b/zigpy/application.py
@@ -1,7 +1,9 @@
+from __future__ import annotations
+
 import abc
 import asyncio
 import logging
-from typing import Any, Dict, Optional, Union
+from typing import Any, Optional
 
 import zigpy.appdb
 import zigpy.config
@@ -25,9 +27,9 @@ class ControllerApplication(zigpy.util.ListenableMixin, abc.ABC):
     SCHEMA = zigpy.config.CONFIG_SCHEMA
     SCHEMA_DEVICE = zigpy.config.SCHEMA_DEVICE
 
-    def __init__(self, config: Dict):
+    def __init__(self, config: dict):
         self._send_sequence = 0
-        self.devices: Dict[t.EUI64, zigpy.device.Device] = {}
+        self.devices: dict[t.EUI64, zigpy.device.Device] = {}
         self.topology = None
         self._listeners = {}
         self._channel = None
@@ -57,22 +59,31 @@ class ControllerApplication(zigpy.util.ListenableMixin, abc.ABC):
 
     @classmethod
     async def new(
-        cls, config: Dict, auto_form: bool = False, start_radio: bool = True
-    ) -> "ControllerApplication":
+        cls, config: dict, auto_form: bool = False, start_radio: bool = True
+    ) -> ControllerApplication:
         """Create new instance of application controller."""
         app = cls(config)
         await app._load_db()
         await app.ota.initialize()
         app.topology = zigpy.topology.Topology.new(app)
-        if start_radio:
-            try:
-                await app.startup(auto_form)
-            except asyncio.CancelledError:
-                raise
-            except Exception:
-                LOGGER.error("Couldn't start application")
-                await app.pre_shutdown()
-                raise
+
+        if not start_radio:
+            return app
+
+        try:
+            await app.startup(auto_form)
+        except asyncio.CancelledError:
+            raise
+        except Exception:
+            LOGGER.error("Couldn't start application")
+            await app.pre_shutdown()
+            raise
+
+        # Re-initialize partially-initialized devices
+        for device in app.devices.values():
+            if device.is_partially_initialized:
+                LOGGER.warning("Device is partially initialized: %s", device)
+                device.schedule_initialize()
 
         return app
 
@@ -127,6 +138,8 @@ class ControllerApplication(zigpy.util.ListenableMixin, abc.ABC):
 
     def device_initialized(self, device):
         """Used by a device to signal that it is initialized"""
+        LOGGER.debug("Device is initialized %s", self)
+
         self.listener_event("raw_device_initialized", device)
         device = zigpy.quirks.get_device(device)
         self.devices[device.ieee] = device
@@ -191,7 +204,7 @@ class ControllerApplication(zigpy.util.ListenableMixin, abc.ABC):
         message: bytes,
         *,
         dst_addressing: Optional[
-            Union[t.Addressing.Group, t.Addressing.IEEE, t.Addressing.NWK]
+            t.Addressing.Group | t.Addressing.IEEE | t.Addressing.NWK
         ] = None,
     ) -> None:
         self.listener_event(
@@ -231,31 +244,41 @@ class ControllerApplication(zigpy.util.ListenableMixin, abc.ABC):
             dst_addressing=dst_addressing,
         )
 
-    def handle_join(self, nwk, ieee, parent_nwk):
+    def handle_join(self, nwk: t.NWK, ieee: t.EUI64, parent_nwk: t.NWK) -> None:
+        """
+        Called when a device joins or announces itself on the network.
+        """
+
         ieee = t.EUI64(ieee)
         LOGGER.info("Device 0x%04x (%s) joined the network", nwk, ieee)
-        if ieee in self.devices:
+
+        try:
             dev = self.get_device(ieee)
-            if dev.nwk != nwk:
-                LOGGER.debug(
-                    "Device %s changed id (0x%04x => 0x%04x)", ieee, dev.nwk, nwk
-                )
-                dev.nwk = nwk
-                dev.schedule_group_membership_scan()
-            elif dev.initializing or dev.status == zigpy.device.Status.ENDPOINTS_INIT:
-                LOGGER.debug("Skip initialization for existing device %s", ieee)
-                dev.schedule_group_membership_scan()
-                return
-        else:
+            LOGGER.debug("Device already exists")
+        except KeyError:
+            LOGGER.debug("Device is unknown, creating a new one")
             dev = self.add_device(ieee, nwk)
 
+        if dev.nwk != nwk:
+            dev.nwk = nwk
+            LOGGER.debug("Device %s changed id (0x%04x => 0x%04x)", ieee, dev.nwk, nwk)
+
         self.listener_event("device_joined", dev)
+
         dev.schedule_initialize()
+
+        # Rescan groups for devices that are already initialized
+        if dev.status == zigpy.device.Status.ENDPOINTS_INIT:
+            dev.schedule_group_membership_scan()
 
     def handle_leave(self, nwk, ieee):
         LOGGER.info("Device 0x%04x (%s) left the network", nwk, ieee)
-        dev = self.devices.get(ieee, None)
-        if dev is not None:
+
+        try:
+            dev = self.get_device(ieee)
+        except KeyError:
+            return
+        else:
             self.listener_event("device_left", dev)
 
     async def mrequest(
@@ -416,7 +439,7 @@ class ControllerApplication(zigpy.util.ListenableMixin, abc.ABC):
         dstaddr.endpoint = self.get_endpoint_id(cluster.cluster_id, cluster.is_server)
         return dstaddr
 
-    def update_config(self, partial_config: Dict[str, Any]) -> None:
+    def update_config(self, partial_config: dict[str, Any]) -> None:
         """Update existing config."""
         self.config = {**self.config, **partial_config}
 
@@ -473,5 +496,5 @@ class ControllerApplication(zigpy.util.ListenableMixin, abc.ABC):
 
     @classmethod
     @abc.abstractmethod
-    async def probe(cls, device_config: Dict[str, Any]) -> bool:
+    async def probe(cls, device_config: dict[str, Any]) -> bool:
         """API/Port probe method."""
--- a/zigpy/application.py
+++ b/zigpy/application.py
@@ -81,7 +81,7 @@ class ControllerApplication(zigpy.util.ListenableMixin, abc.ABC):
 
         # Re-initialize partially-initialized devices
         for device in app.devices.values():
-            if device.is_partially_initialized:
+            if not device.is_initialized:
                 LOGGER.warning("Device is partially initialized: %s", device)
 
         return app
@@ -137,7 +137,7 @@ class ControllerApplication(zigpy.util.ListenableMixin, abc.ABC):
 
     def device_initialized(self, device):
         """Used by a device to signal that it is initialized"""
-        LOGGER.debug("Device is initialized %s", self)
+        LOGGER.debug("Device is initialized %s", device)
 
         self.listener_event("raw_device_initialized", device)
         device = zigpy.quirks.get_device(device)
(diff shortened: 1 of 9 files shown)


MattWestb:
Key document: [Base Device Behavior Specification Version 1.0](https://zigbeealliance.org/wp-content/uploads/2019/12/docs-13-0402-13-00zi-Base-Device-Behavior-Specification-2-1.pdf) 10.2.3 Trust Center Link Keys.

All is in there but not easy understanding for "normal user" like my ;-))

zigpy/device.py
```diff
@@ -47,117 +54,150 @@ class Device(zigpy.util.LocalLogMixin, zigpy.util.ListenableMixin):
...
+        return self._initialize_task
+
+    @property
+    def is_partially_initialized(self):
+        """
+        Some devices were not properly initialized and need to have more information
+        requested after they have been joined.
+        """
+
+        if self.node_desc is None or not self.node_desc.is_valid:
```
Adminiuga, line 123: is there a case when `self.node_desc`  is None? The device class is initiallie with a NodeDescriptor object, but it is not valid and that's why it has the property `is_valid`. I guess we can remove that property, but then in order to distinguish if the node-descriptor was properly initialized, we would need to have is as None

zigpy/device.py
```diff
@@ -33,8 +35,13 @@ class Status(enum.IntEnum):
 
     # No initialization done
     NEW = 0
+
+    # Node descriptor has been read
+    NODE_DESC = 3
```
Adminiuga, line 40: What do you think about making `device.Status` an intFlag instead? We still will be able to do all these, but also could have some additional states like device left, device rejoined and/or new_device?
So when device leaves, we could just mark it as "device left" and keep it in the database, till it rejoins next time?

zigpy/device.py
puddly: I think it seems better to start off with `Device(...).node_desc is None`. I tried changing it to `None` in `Device.__init__` but a few other things broke and left it as-is here.

Really, this draft PR is a MVP and functions enough to demonstrate how the improved joining behavior works in ZHA (which is how I tested it).

 - Devices with missing node descriptors are re-initialized on startup (which queries just the node descriptor) and every time they send data back to zigpy.
 - Re-initialization works normally, so IKEA devices rejoining does nothing but ask the coordinator to fire off a few requests. When they re-join again and announce themselves, it'll work as intended, without waiting for the request retry period to elapse. Perhaps an "initialization delay" would be useful here.

zigpy/device.py
```diff
@@ -47,117 +54,150 @@ class Device(zigpy.util.LocalLogMixin, zigpy.util.ListenableMixin):
...
     async def _initialize(self):
-        if self.status == Status.NEW:
+        """
+        Discover all basic information about a device.
+        Can be called multiple times and should be reentrant.
+        """
+
+        # Some devices are improperly initialized and are missing a node descriptor
+        if self.node_desc is None or not self.node_desc.is_valid:
             await self.get_node_descriptor()
```
Adminiuga, line 156: if it raises, were that exception is caught and logged?

zigpy/device.py
```diff
@@ -47,117 +54,150 @@ class Device(zigpy.util.LocalLogMixin, zigpy.util.ListenableMixin):
...
+
+        if self.status == Status.NEW:
+            self.status = Status.NODE_DESC
```
Adminiuga, line 159: What if we do this like:

```suggestion

        if self.status == Status.NEW or self.node_desc is None or not self.node_desc.is_valid:
            await self.get_node_descriptor()
            self.status = Status.NODE_DESC
```

zigpy/device.py
puddly: If we use the upper four bits of `status` (or extend it to 16 bits?) that seems like it should work, though with functional migrations from #711 anything is possible.

My main goal with adding a new "status" value was to provide ZHA with UI progress indicators, so that it can display some more progress messages when a device is joining. Tbh I'm not sure if this is a useful addition, since in my experience either devices go through the initialization process in less than a few seconds, or they fail before it starts.

Maybe it's better to make the device's join state a new field/db column?

Adminiuga: Definitely not an extra field in sql. It is an i t enum, not an 8 bit int. With enum, the device can be only in one state. As a flag, could have some variations, e.g. fully initialized and newly joined or fully initialized but rejoined.

zigpy/device.py
Adminiuga: Should not re-initialize devices on start-up. Zha hammers the network enough without the extra traffic, better to trigger re-initialize on incoming traffic, just want to make sure that one is not re-entered.

zigpy/device.py
puddly: Can we get rid of device state/status tracking entirely?

The device status will end up in the database only when a device is fully initialized and these states can be derived from what the `Device` and `Endpoint` objects expose:

```Python
class Device:
    def __init__(self):
        # What does zigpy do with non-joined devices? Do we update the `joined` state if
        # a previously-unjoined device unexpectedly sends a request? What about if a
        # device never "leaves" but just stops responding? It seems better to store a
        # `last_seen` timestamp for the device in the database, but this will cause
        # unnecessary database writes if SQLite persists to disk frequently.
        self.joined: bool = False

        self.node_desc: zdo.types.NodeDescriptor | None = None
        self.endpoints = {...}

    @property
    def _non_zdo_endpoints(self):
        return [ep for epid, ep in self.endpoints.items() if epid != 0]

    @property
    def has_node_descriptor(self) -> bool:
        return self.node_desc not in (None, DUMMY_NODE_DESCRIPTOR)

    @property
    def did_zdo_init(self) -> bool:
        # XXX: Do any devices exist that do not respond properly to `Active_EP_rsp`?
        return bool(self._non_zdo_endpoints)

    @property
    def all_endpoints_init(self) -> bool:
        return self.did_zdo_init and all(ep.state != EndpointState.NEW for ep in self._non_zdo_endpoints)

    @property
    def is_initialized(self) -> bool:
        return self.has_node_desc and self.all_endpoints_init
```

Adminiuga: That would work too.

Adminiuga: I meant to say go for it.

puddly: Sure. Since it's going to mess with the database, I'd like to get #711 merged to make any migrations possible. Or we could just ignore the `status` column for now and merge that PR later, since that PR in its current state it has the dummy node descriptors to make upstream code temporarily work around the "missing node descriptor" problem.

zigpy/device.py
puddly: I've decorated `_initialize` with `zigpy.utils.retryable` so it'll retry initialization three times before giving up, skipping over the parts that worked each time. Should log errors as well.

zigpy/device.py
Adminiuga: Let's have this ready for the next release without app migration and include it after the next release to give more time for testing. 
We don't need to modify db schema for this PR, we could just only save the fully initialized device and hardcore the device status when saving to maintain compatibility in case need to rollback the zigpy version.

zigpy/device.py
```diff
@@ -47,118 +49,191 @@ class Device(zigpy.util.LocalLogMixin, zigpy.util.ListenableMixin):
...
-                self.warning("Requesting Node Descriptor failed: %s", status)
-        except (asyncio.TimeoutError, zigpy.exceptions.ZigbeeException):
-            self.warning("Requesting Node Descriptor failed", exc_info=True)
+
+        self.node_desc = node_desc
+        self.info("Got Node Descriptor: %s", node_desc)
+
+        return node_desc
+
+    async def initialize(self, *, tries=3, delay=0.5):
```
Adminiuga, line 155: Any reason retryable decorator doesn't work here anymore?

zigpy/device.py
```diff
@@ -47,118 +49,191 @@ class Device(zigpy.util.LocalLogMixin, zigpy.util.ListenableMixin):
...
+        """
+        Attempts multiple times to discover all basic information about a device: namely
+        its node descriptor, all endpoints and clusters, and the model and manufacturer
+        attributes from any Basic cluster exposing those attributes.
+        """
+
+        for attempt in range(1, tries + 1):
```
Adminiuga, line 162: IMO 
```
while tries:
    tries--
```
looks better

zigpy/device.py
```diff
@@ -47,118 +49,191 @@ class Device(zigpy.util.LocalLogMixin, zigpy.util.ListenableMixin):
...
+            try:
+                await self._initialize()
+                break
+            except asyncio.CancelledError:
+                raise
+            except Exception:
```
Adminiuga, line 168: Is this too "broad" ? Should retry if it is a timeout or a ZigBee exception, for any other reason it is a BUG and should throw the exception

zigpy/device.py
puddly: Gotta emit `device_init_failure` (which is apparently unused) and call `self.application.remove` when it fails. I guess you could also attach a callback to the `Task` object created during the `schedule_initialize` method.

Do you think it would be better to keep around devices that failed to initialize so that a re-join has a higher chance of succeeding due to some info already being filled out?

Adminiuga: https://github.com/zigpy/zigpy/pull/746 it is an async callback for the retryable decorator

> Do you think it would be better to keep around devices that failed to initialize so that a re-join has a higher chance of succeeding due to some info already being filled out

I was thinking about that too. I'm onboard, just need to modify ZHA so it doesn't try to discover entities for not fully initialized devices

puddly: `Application.remove` utilizes retryable requests with some pretty long timeouts. Should all requests to a device get cancelled if it leaves or re-joins? Or just the removal ones? While somewhat unlikely, it seems possible for a device to fail initialization, the removal taskgets started, and the device re-joins during a removal retry, causing it to be kicked off the network.

Alternatively, we could also *not* send any ZDO remove requests to the device when an initialization failure occurs, since a device that fails to initialize is likely already factory reset again and if the device miraculously starts sending requests again, continuing to initialize it seems like desired behavior?

Adminiuga: Good question. Sounds like `Application.remove` should cancel the `Device.initialize()` and `Device.initialize()` should cancel `device.remove()`. Although the timeouts are mostly waiting for a reply, meaning the request was sent out anyway. But we definitely should cancel any remaining retries.

Adminiuga:
This is a follow up to https://github.com/zigpy/zigpy/pull/746#issuecomment-846705836

I suggest:
1. we don't do device removal on failed initialization, solves problem with task tracking and interaction
2. ZHA ignores not initialized devices
3. At least we do need to save device NWK & IEEE address, because on Silabs it is hard to query device's IEEE based on NWK and we could end up in situation when we receive an incoming message from an unknown device, but we can't do the device discovery at the time when device is responsive, because we don't have device IEEE. In other words, have an persisting listerner event to just save the NWK, IEEE and Status (for now). We could use this event when only just NWK updates.

puddly:
1. Done.
3. Does SiLabs not support the ZDO `IEEE_addr_req`? Can we instead allow the appdb listener to persist incomplete devices in their entirety and just have ZHA ignore ones which are not initialized when zigpy loads (and after join)?

puddly:
There have been some reports of devices changing NWK addresses without notifying the coordinator, so the radio library can never successfully call `self.get_device()`. I'm thinking of maybe adding an API like this and making all of the `handle_` methods async? Or maybe having zigpy deal with this interally and remove `get_device` from radio libraries? It would basically allow for automatic device discovery on completely unknown networks:

```python
async def get_or_discover_device(self, nwk) -> zigpy.device.Device | None:
    try:
        # If a device with this NWK exists, return it
        return self.get_device(nwk=nwk)
    except KeyError:
        pass

    # Otherwise, find its IEEE address
    ieee = await self.ieee_addr_req(nwk=nwk)

    try:
        dev = self.get_device(ieee=ieee)
    except KeyError:
        # Start discovery for a completely unknown device
        self.handle_join(nwk=nwk, ieee=ieee, parent_nwk=None)
        dev = self.get_device(ieee=ieee)
    else:
        # Otherwise, update the existing device's NWK address
        # Maybe have this create a NWK change event?
        dev.nwk = nwk

        # Or something like this?
        await dev.update()

    return dev


async def on_incoming_message(self, nwk, data, ...):
    dev = await self.get_or_discover_device(nwk=nwk)
```

Adminiuga:
> Does SiLabs not support the ZDO IEEE_addr_req?
It does, but feels like it is just querying the internal cache tables and does not send the actual ZDO request. Although, last time I tries the EZSP commands for those. May need to retest with ZDO

> Can we instead allow the appdb listener to persist incomplete devices in their entirety and just have ZHA ignore ones which are not initialized when zigpy loads (and after join)?
Yes, this is what I was trying to ask: are there any downsides of persisting not fully initialized devices with the assumption that ZHA ignores not fully initialized devices?

puddly:
> are there any downsides of persisting not fully initialized devices with the assumption that ZHA ignores not fully initialized devices?

Hmm, there is a big one: half-initialized devices that users are currently using won't load. How about for half-initialized devices that fail to join, we persist only their node descriptor and NWK-IEEE mapping, leaving off the endpoint information? It's a little hacky but there's no way to tell apart a half-joined device from a few months ago from a "cached" device without adding more state to the DB, which I'd like to avoid until we can properly modify the database with migrations.

I think when ZHA starts up, it should (for now) load any device with non-ZDO endpoints, not just devices that are fully initialized. They will be slowly re-initialized at runtime since they likely work. How will ZHA deal with a device getting a quirk after it has loaded?

Adminiuga:
> There have been some reports of devices changing NWK addresses without notifying the coordinator
Most likely the coordinator was not running at the time or they have some networking problems. In case of a NWK conflict, the devices go through a re-join process. 

> so the radio library can never successfully call self.get_device(). I'm thinking of maybe adding an API like this and making all of the handle_ methods async? Or maybe having zigpy deal with this interally and remove get_device from radio libraries?

For now let's keep it this out of the scope of this PR. I'm on the fence whether this decision should be made by Zigpy or if a partiacular radio has better knobs to handle this situation. For now let zigpy raise the KeyError and radio lib can decide how to handle it. Bellows does something like this already. 

> It would allow automatic device discovery on completely unknown networks, in addition to properly dealing with NWK changes
I was thinking just to have a "crawler" to iterate over the topology map. But again, topic for another PR.

Adminiuga:
>Hmm, there is a big one: half-initialized devices that users are currently using won't load. How about for half-initialized devices that fail to join, we persist only their node descriptor and NWK-IEEE mapping, leaving off the endpoint information? 

+1

> It's a little hacky and doesn't persist as much information as it can but there's no way to tell apart an "old" half-joined device from a "new" cached device without adding more state to the DB, which I'd like to avoid until we can properly modify the database with migrations.

I was thinking about ZHA loading any devices which have their status equivalent to ZCL_INIT. There are vendor which do not support model and manufacturer attributes. But other than those, I think it is safe to assume if you have endpoints and clusters, then the device could be considered initialized? and ZCL_INIT assumes we do have the node descriptor. 

>I think when ZHA starts up, it should (for now) load any device with non-ZDO endpoints, not just devices that are fully initialized. They will be slowly re-initialized at runtime since they likely work. How will ZHA deal with a device getting a quirk after it has loaded

By "fully initialized" I do mean the devices with non-ZDO endpoints and endpoint descriptors (Clusters) and assuming we do have model and manufacturer in the cache.

puddly:
Right now there are three implicit initialization states a device can have. `Device.status` isn't always indicative of the presence of info (like you can be at `ZDO_INIT` but earlier zigpy allowed the node descriptor request to fail). I'm just going off what's present in the database:

 1. `has_node_descriptor`: self-explanatory.
 2. `has_non_zdo_endpoints`: the endpoints exist, they may not be necessarily initialized.
 3. `all_endpoints_init`: every endpoint's status is either `ZDO_INIT` or `ENDPOINT_INACTIVE`. A device could be missing its node descriptor and thus not be `is_initialized` but it can still be `all_endpoints_init` and functional.
    To this state I'd like to also add `self.model is not None and self.manufacturer is not None`, which will force re-reading those attributes when they're missing.

The missing/unsupported manufacturer and model info is going to be a little annoying to handle. Currently those attribute reads are allowed to fail due to delivery and timeout errors, not just if the attribute doesn't exist, so we can't assume that missing model info in the database must mean that the device doesn't support it.

How about assuming all devices with missing model info are missing it due to improper initialization and we explicitly store a special "not present" value for devices that explicitly indicate that the attribute is not supported? This will cause one round of unnecessary initialization for the few devices that do not support these attributes but I think the vast majority of devices that are missing these attributes are just improperly initialized.

Something like this:

```Python
# The database can store a special "attribute_really_is_not_present" value
# which can be migrated later
MISSING = object()

class Device:
    def __init__(self):
        self._model = MISSING  # `None` indicates that the value really is `None`
        self._manufacturer = MISSING
```

puddly:
> But other than those, I think it is safe to assume if you have endpoints and clusters, then the device could be considered initialized?

Are you aware of any devices the expose endpoints without any clusters?

Adminiuga:
I'm just talking  in the context of `zigpy,device.Status`:
- NEW -- un-initialized device
- ZDO_INIT -- have node_descriptor (should have) and the list of Endpoints, but not endpoint descriptors (i.e. Node_Desc_req & Active_EP_req completed)
- ENDPOINTS_INIT -- completed Simple_Desc_req for all endpoint and attempted to read model & manufacturer until succeeded at least once for each endpoint

ZHA load devices only which are in zigpy.device.ENDPOINTS_INIT state. 
For backward compatibility, for now we save devices status derived from your state properties:
- `has_non_zdo_endpoints` -- ZDO_INIT (may miss node_desc, but that's fine, cause this PR would query for node_desc on incoming message)
- `all_endpoints_init` -- ENDPOINT_INIT. I say for now we ignore missing model & manufacturer, as worst case scenario users can read those manually, nor I think it is a wide spread problem, because if device is responding to Simple_Desc_req, then very likely to respond to attribute read req. 


> To this state I'd like to also add self.model is not None and self.manufacturer is not None, which will force re-reading those attributes when they're missing.

Skip this for now. In the long term we just need to have a table of unsupported attributes per device

puddly:
> In the long term we just need to have a table of unsupported attributes per device

Just out of curiosity, I tried this change out on my own network after clearing the attribute cache. Mains-powered devices were re-initialized without any issues and so were most Aqara sensors, surprisingly (after a few hours). The persistent retrying-on-message approach seems to work well, even for battery-powered devices.

zigpy/application.py
```diff
@@ -191,37 +205,38 @@ def handle_message(
...
-                dst_ep,
-                cluster,
-            )
-            return
+
+        if not sender.is_initialized:
+            if not sender.initializing:
+                sender.schedule_initialize()
+
+            if not sender.has_node_descriptor and dst_ep != 0:
```
Adminiuga, line 219: Do we need `not sender.has_node_descriptor` here? Doesn't `not sender.is_initialized` satisfies this condition already?

zigpy/application.py
```diff
@@ -231,31 +246,41 @@ def handle_message(
...
-                dev.nwk = nwk
-                dev.schedule_group_membership_scan()
-            elif dev.initializing or dev.status == zigpy.device.Status.ENDPOINTS_INIT:
-                LOGGER.debug("Skip initialization for existing device %s", ieee)
-                dev.schedule_group_membership_scan()
-                return
-        else:
+            LOGGER.debug("Device already exists")
+        except KeyError:
+            LOGGER.debug("Device is unknown, creating a new one")
```
Adminiuga, line 261: let's save a logging line:
```
        try:
            dev = self.get_device(ieee)
            LOGGER.info("Device 0x%04x (%s) re-joined the network", nwk, ieee)
        except KeyError:
            LOGGER.info("New device 0x%04x (%s) joined the network", nwk, ieee)
```

zigpy/application.py
puddly: Devices that have no node descriptor would be unusable until they're initialized if this check isn't there.

puddly: Or not, that's what *will* happen with this change.

Maybe it's better to use `if not sender.all_endpoints_init` instead? In that case, the `elif` branch seems like it can be removed, as the presence of initialized endpoints is really the only proxy we have for "probably functional".

I understand the purpose of the first branch, but what is the purpose of the second one? Is it to prevent errors from being logged about missing endpoints and clusters?

Adminiuga: >  , but what is the purpose of the second one? Is it to prevent errors from being logged about missing endpoints and clusters?

Yes. With the exception of the basic cluster so we could read manufacturer and model. 
But now thinking about it, maybe it is better to allow zdo traffic regardless of the status and ZCL only if all_endpoints_init with the exception of the basic cluster

puddly: Yeah, that sounds reasonable. I've adjusted `handle_message` accordingly.

zigpy/application.py
```diff
@@ -191,71 +205,104 @@ def handle_message(
...
             dev = self.add_device(ieee, nwk)
+            LOGGER.info("New device 0x%04x (%s) joined the network", nwk, ieee)
+
+        if dev.nwk != nwk:
+            dev.nwk = nwk
+            LOGGER.debug("Device %s changed id (0x%04x => 0x%04x)", ieee, dev.nwk, nwk)
 
         self.listener_event("device_joined", dev)
+
         dev.schedule_initialize()
```
puddly, line 292: HA `dev` works without any modifications 🎉  but the only regression right now is this line causing ZHA to do its re-initialization for every device that re-joins the network after a power cycle.

Zigpy didn't previously emit a `device_joined` or `device_initialized` event when an already-initialized device re-joined the network either. Maybe we need to change it to this?

```Python
if not dev.is_initialized:
    dev.schedule_initialize()
```

Also, does ZHA keep no initialization state of its own? As in, once a device joins from zigpy's perspective, is the ZHA initialization progress persisted anywhere?

Adminiuga: ZHA does not keep the initialization state of its own and cannot really tell a re-join from join, so it does ZHA device reconfigure every time it receives the event "device_joined"

Adminiuga: Also, not all radios can tell zigpy if it is a new join or a re-join. zigpy could have device in the DB, but it still is a new join if the device was reset and paired, without removing from zigpy 1st. I think we can assume, any time the device changes NWK it is a new join

Adminiuga: IMO we should do line 290 `self.listener_event("device_joined", dev)` only if NWK was changed or device did not exist before. Need to check ZHA code, as I believe it kicks of ZHA device configuration after the event fired by `Device.initialize()`

puddly: For partially-initialized devices that finish initialization when receiving a message, how about doing something like `schedule_initialize(suppress_event=True)`? Or would it be better for ZHA to redo its initialization??

Adminiuga: Does it do a re-initialization for every device even the ones there were fully initialized before? I.e. have node descriptor, endpoints and clusters?

puddly: It doesn't do anything but `schedule_initialize`/`initialize` will still call `self._application.device_initialized(self)` in the end.

Adminiuga: If it happens only for partially initialized devices, then I'd say that correct behavior. Would be mitigated on zha side by ignoring partially initialized devices

Adminiuga: I think if we `schedule_initialize()` only on:
1. Partially initialized device
2. Nwk change
3. New device created/no existing devices

Which in turn would trigger device re-initialization on zha side, then this would be correct and expected behavior.

If it's a handle_join() but NWK is the same and device is initialized, then we should skip schedule _initialize and just do group_member_scan()

puddly: `handle_join` should now match this behavior.

Adminiuga:
so far looks great! Trying in my prod next

Adminiuga:
so far so good

Adminiuga:
is this still draft? are you planning to add anything else?

puddly:
Nah, feature-wise it's all done. Just gotta look over the changes once more to check for corner cases.

Adminiuga (approved):

PR #745, opened 2021-05-23
Add context to the "Discarded Event" warning

dd32:
The `Discarding %s event` warning is logged with the `%s` left as is, instead of replacing the expected context of which event is being discarded.

### DIFF
--- a/zigpy/appdb.py
+++ b/zigpy/appdb.py
@@ -121,9 +121,7 @@ class PersistingListener(zigpy.util.CatchingTaskMixin):
     def enqueue(self, cb_name: str, *args) -> None:
         """Enqueue an async callback handler action."""
         if not self.running:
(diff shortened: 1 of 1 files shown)


puddly:
Thanks for the PR! It looks like [Black](https://github.com/psf/black) is current formatting the call to `LOGGER.warning` to take up three lines due to the trailing comma. Adding the second missing argument makes it like a normal function call and it can be condensed into a single line:

```Python
            LOGGER.warning("Discarding %s event", cb_name)
```

Adminiuga (approved):

dd32:
Thanks for the quick response and following that through @puddly and @Adminiuga!

I came back to follow up on the contributing guidelines & unit test failure only to find it was already done :)

PR #746, opened 2021-05-23
Add an async_on_failure callback for retryable decorator

Adminiuga:
An async_callback to be awaited if the retryable decorator exhausted all the attempts.

### DIFF
--- a/zigpy/util.py
+++ b/zigpy/util.py
@@ -99,7 +99,7 @@ class LocalLogMixin:
         return self._log(logging.ERROR, msg, *args, **kwargs)
 
 
-async def retry(func, retry_exceptions, tries=3, delay=0.1):
+async def retry(func, retry_exceptions, tries=3, delay=0.1, async_on_failure=None):
     """Retry a function in case of exception
 
     Only exceptions in `retry_exceptions` will be retried.
@@ -111,12 +111,14 @@ async def retry(func, retry_exceptions, tries=3, delay=0.1):
             return r
         except retry_exceptions:
             if tries <= 1:
+                if async_on_failure is not None:
+                    await async_on_failure
                 raise
             tries -= 1
             await asyncio.sleep(delay)
 
 
-def retryable(retry_exceptions, tries=1, delay=0.1):
+def retryable(retry_exceptions, tries=1, delay=0.1, async_on_failure=None):
     """Return a decorator which makes a function able to be retried
 
     This adds "tries" and "delay" keyword arguments to the function. Only
@@ -135,6 +137,7 @@ def retryable(retry_exceptions, tries=1, delay=0.1):
                 retry_exceptions,
                 tries=tries,
                 delay=delay,
+                async_on_failure=async_on_failure,
             )
 
         return wrapper
--- a/zigpy/util.py
+++ b/zigpy/util.py
@@ -112,7 +112,7 @@ async def retry(func, retry_exceptions, tries=3, delay=0.1, async_on_failure=Non
         except retry_exceptions:
             if tries <= 1:
                 if async_on_failure is not None:
-                    await async_on_failure
+                    await async_on_failure()
                 raise
             tries -= 1
             await asyncio.sleep(delay)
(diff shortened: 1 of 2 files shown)


zigpy/util.py
```diff
@@ -111,12 +111,14 @@ async def retry(func, retry_exceptions, tries=3, delay=0.1):
             return r
         except retry_exceptions:
             if tries <= 1:
+                if async_on_failure is not None:
+                    await async_on_failure
```
puddly, line 115: Wouldn't you want `await async_on_failure(e)` or something? Otherwise, if an exception isn't raised, it seems like this would generate a "coroutine not awaited" warning.

zigpy/util.py
```diff
@@ -135,6 +137,7 @@ def wrapper(*args, tries=tries, delay=delay, **kwargs):
                 retry_exceptions,
                 tries=tries,
                 delay=delay,
+                async_on_failure=async_on_failure,
```
puddly, line 140: `retryable` always wraps methods and if the `async_on_failure` function is also a method, you'd have to add `self` as the first argument of `def wrapper(` to get a reference to it and then "bind" `async_on_failure` to it.

I was trying to implement an API like this:

```Python
    async def on_initialize_failure(self, error=None):
        """
        Called when a device fails to initialize.
        """

        self.application.listener_event("device_init_failure", self)
        await self.application.remove(self.ieee)

    @zigpy.util.retryable([
        asyncio.TimeoutError, zigpy.exceptions.ZigbeeException
    ], tries=3, delay=0.5, failure_callback=on_initialize_failure)
    async def initialize(self):
        ...
```

zigpy/util.py
Adminiuga: yes, You are correct, should just pass the callable and await it here.

zigpy/util.py
Adminiuga: yeah, I forgot that with decorators, passing any instance specific detail is not trivial. Open for ideas.

Adminiuga: could do it the hacky way, where it is the "callback name" to be called if the args[0] has that member???

puddly: I'm not seeing a nice way to do this with decorators, tbh. How about using task callbacks?

```Python
        self._initialize_task = asyncio.create_task(self.initialize())

        def done_callback(task):
            try:
                task.result()
            except Exception:
                self.application.listener_event("device_init_failure", self)
                asyncio.create_task(self.application.remove(self.ieee))

        self._initialize_task.add_done_callback(done_callback)

        return self._initialize_task
```

This won't be called if `Device.initialize()` is directly awaited but that seems almost like a feature?

Adminiuga: So, what about this approach: pass the callback by the name, which would be called if there're args and `args[0]` has a member by that name. 
Is this too hacky?

Adminiuga:
So, what about this approach: pass the callback by the name, which would be called if there're args and args[0] has a member by that name.
Is this too hacky?

Adminiuga:
The test is modeled similar to device initialization approach

Adminiuga:
don't see any other way if doing this via decorators and frankly I like the decorators.

puddly:
Passing by name works too. I think you can also "re-bind" the unbound method by checking if `getattr(self, callback.__name__).__func__ is callback`. Or something [more magical](https://stackoverflow.com/a/1015389).

It's a minor detail at this point so if you think this `async_on_failure` will be used elsewhere then this approach works. Otherwise, I would prefer to avoid more complicated code in `retryable` and do something like this:

```Python
    async def initialize(self):
        try:
            await self._initialize()
        except asyncio.CancelledError:
            raise
        except Exception as e:
            if not isinstance(e, (
                asyncio.TimeoutError, zigpy.exceptions.ZigbeeException
            )):
                LOGGER.warning("Device failed to initialize due to unexpected error", exc_info=True)

            self.application.listener_event("device_init_failure", self)
            asyncio.create_task(self.application.remove(self.ieee))

    @zigpy.util.retryable((
        asyncio.TimeoutError, zigpy.exceptions.ZigbeeException
    ), tries=3, delay=0.5)
    async def _initialize(self):
        ...
```

Adminiuga:
Not ATM, so your approach is cleaner. Let's use it.

Adminiuga:
> `asyncio.create_task(self.application.remove(self.ieee))`

may want to do `self.create_catching_task(self.application.remove(self.ieee))` here, but I was thinking if it is better to do an await here, so it still could be cancelled??? 

BUT, do we really have to do this here? Let's say ZHA ignores not fully initialized devices, then could we omit device removal on  failed init and do incremental device initialization???

puddly:
Yeah, I don't see the harm in letting the failed-to-initialize device to stay on the network and for zigpy to keep track of it, at least until the next HA restart (since it won't end up in the database).

Getting device removal and initialization to interact sanely is going to take a little bit of work, especially with all the async cancellation corner cases (i.e. `create_catching_task`). You'd also have to keep track of the device removal task, which would require all the `*remove*` methods to be moved into the `Device` class.

PR #747, opened 2021-05-23
Detect logging with bad format strings during unit tests

puddly:
Causes unit tests to fail if a logging formatting string is invalid (#745).


Adminiuga (approved):
:+1:

PR #759, opened 2021-06-20
Fix direct `Cluster` instances causing errors in `Endpoint.__repr__`

puddly:
Zigpy allows for `Cluster` itself to be initialized (not a subclass!), whose instances don't have a `ep_attribute` attribute, causing this error (only when debug logging is enabled?).


puddly:
Yeah. It's a little too late to fix now that it's merged but you're right, `Optional[str]` is the correct type annotation.

PR #767, opened 2021-07-03
Ignore all GBL image padding

puddly:
GBL images are not supposed to have trailing padding. I was only aware of Hue images that had `\x00` padding but apparently other images exist with `\xFF` padding. If an image is structurally sound and its CRC is valid, it should be accepted.

### DIFF
--- a/tests/test_ota_validators.py
+++ b/tests/test_ota_validators.py
@@ -122,17 +122,10 @@ def test_parse_silabs_gbl():
     assert tag2 == (b"BBBB", b"foo" * 20)
     assert checksum[0] == b"\xFC\x04\x04\xFC" and len(checksum[1]) == 4
 
-    # No padding is allowed
-    with pytest.raises(ValidationError):
-        list(validators.parse_silabs_gbl(image + b"\xFF"))
-
-    # Unless the padding is entirely null bytes
+    # Arbitrary padding is allowed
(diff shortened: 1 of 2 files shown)


MattWestb:
I was baying one [IKEA LEPTITER](https://www.zigbee2mqtt.io/devices/T1820.html) in the recycling hub because i like testing the CT drive in it.
Its using FLOALT firmware but its not possible dumping it then its to good sealed but it was updating nice and easy (ca 15 minutes) and i was putting in the product json then want getting more problems with my controllers than i have getting.
And nearly all firmware request is complaining of trunking data in GBL (i think most in the test feed is EBL files and the prod is GBL) that i have not seeing before.

Good to knowing its being fixed and its working OK.

Great work done !!!

Adminiuga (approved):
Lgtm

MattWestb:
Looks like is still rising warnings in the HA log (My production system):
```
Logger: zigpy.ota.image
Source: /usr/local/lib/python3.9/site-packages/zigpy/ota/image.py:244
First occurred: 12:00:46 (1 occurrences)
Last logged: 12:00:46

Fixing IKEA OTA image with trailing data (1100 bytes)
```
Only cosmetic for my point of view then 2021-7 have braking ESPHome cam that is spamming the HA log :-((

PR #768, opened 2021-07-05
Re-migrate V4 database which was forcibly downgraded to v3

Adminiuga:
Drop existing v4 tables and re-migrate from v3 schema if user downgraded zigpy and then upgraded again.

Fixes https://github.com/home-assistant/core/issues/52394

### DIFF
--- a/tests/test_appdb_migration.py
+++ b/tests/test_appdb_migration.py
@@ -27,6 +27,29 @@ def test_db_v3(tmpdir):
     yield db_path
 
 
+@pytest.fixture
+async def test_db_v4_downgraded_to_v3(test_db_v3):
+    """V4 database forcibly downgraded to v3."""
+
+    app = await make_app(test_db_v3)
+    await app.pre_shutdown()
+
+    with sqlite3.connect(test_db_v3) as conn:
+        # new neighbor
+        conn.execute(
+            """INSERT INTO neighbors VALUES(
+                   'ec:1b:bd:ff:fe:54:4f:40',
+                   '81:b1:12:dc:9f:bd:f4:b6',
+                   '00:0d:6f:ff:fe:a6:11:7b',
+                   48462,
+                   37,2,15,132)
+            """
+        )
+        conn.execute("PRAGMA user_version=3")
+    conn.close()
+    yield test_db_v3
+
+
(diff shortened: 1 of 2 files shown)


puddly (approved):
Looks good.

If we're sticking to dropping tables when re-migrating, lemme undo the last few commits to #711. This was the logic I had as of https://github.com/zigpy/zigpy/pull/711/commits/1657f3de35a748fbe92cfb2d8fa425e0b06d99ca.

Adminiuga:
@puddly doh, i missed your [comment](https://github.com/home-assistant/core/issues/52394#issuecomment-873760626) I think that approach is fine. Whatever approach we take (drop V4 and remigrate or just reuse V4 tables) will have users on both sides. 

Although, one could say if you downgraded and ran V3, then V3 tables become authoritive again and need to be remigrated to V4? 

Anyhow, I'm hesitant to push #711 into 2021.7.x but would like to have a fix for https://github.com/home-assistant/core/issues/52394 prior the release on 7/7

Once we get to Core 2021.7.2 we can push #711 through. 
Thoughts?

puddly:
> Although, one could say if you downgraded and ran V3, then V3 tables become authoritive again and need to be remigrated to V4?

Yeah, there are edge cases for both migration approaches:

 - skipping migrations if the target tables already exist:
   - if you downgrade to v3, add new devices, and upgrade to v4, those "new" devices will be lost.
 - dropping the new tables before migrating:
   - if you add a bunch of devices in a v4 table, downgrade back to v3, and then go back to v4, all of the devices you've added since v3 will be gone.

Not sure what happens more often 🤷‍♂️.

> Anyhow, I'm hesitant to push #711 into 2021.7.x but would like to have a fix for home-assistant/core#52394 prior the release on 7/7

Agreed. Let's just merge this simple PR and deal with whatever form #711 takes by the next release cycle.

Adminiuga:
> if you add a bunch of devices in a v4 table, downgrade back to v3, and then go back to v4, all of the devices you've added since v3 will be gone.

Yes, but the mitigating factors:
1. V3 schema would force you to re-add devices if you really want to use them
2. Changes in V4 would trigger device discovery if some info is missing. 

I think just dropping and re-migrating is just a tad better.

PR #774, opened 2021-07-24
Add ZCL7 start up attributes to `OnOff` and `LevelControl` clusters

puddly:
Not sure if we need a new type for "integers with special values", like `StartUpCurrentLevel`:

 - `0x00` is the minimum
 - `0xFF` is the previous value
 - every other value is handled normally


Adminiuga (approved):
LGTM

PR #782, opened 2021-08-08
Fix `None` node descriptor issues with device removal and the topology scanner

puddly:
I've looked over every reference to `Device.node_desc` in both zigpy and ZHA to hopefully make sure this is the last node descriptor bugfix PR.

Related PR: https://github.com/zigpy/zigpy-xbee/pull/119
HA issue: https://github.com/home-assistant/core/issues/53838#issuecomment-894806418


Adminiuga (approved):
Lgtm

PR #798, opened 2021-08-31
Add a neighbor field in NetworkState object

Adminiuga:
An attempt was made to address https://github.com/zigpy/zigpy-znp/pull/85 comments

@puddly for visibility. Would something like this work for ZNP?

### DIFF
--- a/zigpy/state.py
+++ b/zigpy/state.py
@@ -57,6 +57,7 @@ class NetworkInformation:
     network_key: Key | None = None
     tc_link_key: Key | None = None
     key_table: list[Key] | None = None
+    neighbor_table: list[NodeInfo] | None = None
 
     # Dict to keep track of stack-specific network stuff.
     # Z-Stack, for example, has a TCLK_SEED that should be backed up.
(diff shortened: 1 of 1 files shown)


puddly:
If we're going with a single list, can you add an `is_child` flag or something to the `NodeInfo` class so that we can tell apart children with APS keys?

Adminiuga:
I was thinking that APS Key would go into `NetworkInformation.key_table` and children (or just peer) would go into `neighbor_table` and based on `NodeInfo,logical_type` you can differentiate between children end_devices and routers/coordinators.

puddly:
You're right, that works. This looks good to me.

puddly:
No, just ones with APS link keys. Z-Stack doesn't keep track of logical types at all. Every device with an APS link key needs a corresponding entry in the Z-Stack address manager table, which requires NWK and IEEE addresses.

puddly:
There seems to be a way to query this stuff with EZSP but it didn't return any results with my tiny test network with a HUSBZB-1: https://github.com/puddly/bellows/blob/puddly/open-coordinator-backup/bellows/cli/backup.py#L114

Issue #827, opened 2021-10-18
OTA Support for Salus


davet2001:
I have started work on OTA support for Salus devices, PR #839.

The list of available firmware files is here:
https://eu.salusconnect.io/demo/default/status/firmware?timestamp=0

For comparison here is the one for ikea:
http://fw.ota.homesmart.ikea.net/feed/version_info.json

One thing I find different is that the list of available firmware from salus doesn't have a device ID that matches the numerical device ID coming from the ZHA device like the ikea and ledvance devices.  In the OTA code, those devices seem to be able to reference the FW by a numerical ID to know which one to use.

For Salus, the only way I can see of doing it is matching on the device name, e.g. "SP600".  If this is the only way, I'll have to make some additions to the base OTA functions to allow it to work like that.

Does anyone have experience or ideas on this?  Are there any additional references I should be using?

puddly:
Thanks!

Currently, the only context passed to the main `OTA` class when a device asks for a firmware image is just the manufacturer ID and image type:

https://github.com/zigpy/zigpy/blob/468b533eda9216dfc679acd6d542849206daa87d/zigpy/zcl/clusters/general.py#L1070-L1072

One way to get this to work is to propagate an additional (optional?) `device` argument into here as well, so your image provider has enough context to pick an image:

https://github.com/zigpy/zigpy/blob/468b533eda9216dfc679acd6d542849206daa87d/zigpy/ota/__init__.py#L117

Adminiuga:
Wonder if firmware download module should be outside of zigpy at all. Then just download/refresh all images and put them into a folder for zigpy to pick

davet2001:
Yes, this could work.

But then every HA user would automatically download all FW images from every vendor?  Perhaps this is not kind to the vendor servers.

I think you would still need a way to iterate through the list and download them (and in the case of Salus extract a tar.gz).

MattWestb:
IKEA is using "custom scenes commands" on there remotes and they is being restored to default then doing OTA upgrade and being lost then resetting the device for paring it and ZHA can **sadly** not adding scenes (missing some ZCL 7 commands). 
So downgrading and upgrading IKEA lights you is getting the scene button working  on your lights :-))

davet2001:
@puddly PR #839 created.

Hedda:
> either by maintaining an external index file, like z2m,...

Would it be a bad idea to maintain and reference a such external index file in the zha-quirks / zha-device-handlers repository?

[...]

PR #839, opened 2021-10-30
Add Salus OTA update capability

davet2001:
As discussed in #827, this is a work in progress PR for supporting OTA upgrades for Salus devices (sometimes also called "Computime").

The mechanism is quite similar to the existing Ledvance and Ikea updates with one main exception:

Salus provide a `.json` file listing all of the firmware images, but unfortunately these do not contain an `image_id` that can be used to select the correct image as advertised by the device.  Only the header of the ota update (once it is downloaded and extracted) shows this.

All users downloading and extracting every `.tar.gz` file that Salus publish (each time a check is done for FW updates) would be an unfriendly burden on the download server.  So instead this PR adds the model name as an alternative to be used for the key.

In honesty I don't know if this will work for all devices, but it works for mine and it appears that the Salus official FW updates must either be doing this (or another API or a proprietary lookup table).

Hint: you can use something like this in `zigpy/ota/__init__.py`:line66 in order to trigger a FW update that would normally be skipped:
```
        if ver >= self.version:
            print("IGNORING FW UPGRADE VERSION CHECK!")
            # return False
```

Current status is that it works for my devices.

Please comment!


davet2001:
@Adminiuga Please could you give me a steer whether this approach is acceptable before I go ahead and write test cases?

Adminiuga:
I'm fine with it. 

@puddly  thoughts?

puddly:
Sure.  It's not like we see a new HTTP image provider added every week so while the image key stuff isn't optimal, it works.  When we get around to rewriting OTA entirely to decouple images it from zigpy, we can properly address the underlying zigpy OTA architecture flaws.

Some code changes:

 - Maybe replace the few instances of `4216` with `Salus.MANUFACTURER_ID` for readability, if circular imports don't make it too difficult.
 - Use a context manager to auto-close the `tarfile` object: `with tarfile.open(fileobj=img_tgz) as tar`.

davet2001:
Test coverage should be up to all but 1 line now.

puddly:
You can fix the failing lint stuff by running the pre-commit hook:

```console
$ pip install pre-commit
$ pre-commit install
$ pre-commit run --all-files
```

It'll auto-format your code.

davet2001:
Should be 100% now, with formatting fixed.  Thanks @puddly!

Hedda:
@davet2001 Will you add "Salus/Computime" in the "Zigbee device OTA updates" section of README.md for zigpy?

https://github.com/zigpy/zigpy/blob/dev/README.md#zigbee-device-ota-updates

And maybe update Home Assistant's ZHA docs once a new zigpy release with it is bumped in Home Assistant core?

https://www.home-assistant.io/integrations/zha#ota-firmware-updates

davet2001:
> @davet2001 Will you add "Salus/Computime" in the "Zigbee device OTA updates" section of README.md for zigpy?
> 
> https://github.com/zigpy/zigpy/blob/dev/README.md#zigbee-device-ota-updates
> 
> And maybe update Home Assistant's ZHA docs once a new zigpy release with it is bumped in Home Assistant core?
> 
> https://www.home-assistant.io/integrations/zha#ota-firmware-updates

Both now done.

The [HA docs PR ](https://github.com/home-assistant/home-assistant.io/pull/20638) is a draft since the zigpy version is not bumped yet.

davet2001:
@

Issue #843, opened 2021-11-05
Soil Moisture measurment is missing


majkrzak:
I've recently acquired a plant watering sensor.
https://github.com/smartboxchannel/Plant-Watering-Sensor-Zigbee
https://www.zigbee2mqtt.io/devices/EFEKTA_PWS.html

Sadly it only reports battery and temperature, but not the most important: Soil Moisture

What I see is that the cluster 0x408 is in the "TODO"
https://github.com/zigpy/zigpy/blob/fc8e923e2f2482810b9643c0b2bde149a56a1930/zigpy/zcl/clusters/measurement.py#L122

Should it be just added in the similar way as the other there?

puddly:
You can find the appropriate cluster in [the ZCL v7 spec](https://github.com/Koenkk/zigbee-herdsman/blob/master/docs/Zigbee%20Cluster%20Library%20Specification%20v7.pdf), it'll show the cluster's attributes. I believe it's similar in structure to the surrounding clusters (min/max/tolerance and value attributes), if not identical.

Issue #826, opened 2021-10-16
Thermostat remote temperature


mdeweerd:
As far as I understand it a HVAC device such as a Thermostat can receive the temperature from a remote sensors if attribute 0x201/0x001a RemoteSensing is correctly set.

I believe that the remote sensor has to be bound to the thermostat.  [TI's sample implementation show that it accepts incoming temperature reports and uses them as the "LocalTemperature"](http://ala.isti.cnr.it/svn/wnlab/MONSTER/it.cnr.isti.monster.firmwares/ZStack%202.6.2%20for%20IAR%208.30/Z-Stack/Projects/zstack/HomeAutomation/SampleThermostat/Source/zcl_samplethermostat.c) in zclSampleThermostat_ProcessInReportCmd .
And I believe that this is possible if the Thermostat has a "TemperatureMeasurement" output cluster.

I am trying to set this up in Home Assistant, but I think it all boils down to zigpy functionnality.  I've been able to bind a remote control to a light bulb (through a group) for instance, but I can't bind the temperature sensor to the Thermostat.

I have a Thermostat that has the appropriate output cluster:
```json
{
  "node_descriptor": "NodeDescriptor(logical_type=<LogicalType.EndDevice: 2>, complex_descriptor_available=0, user_descriptor_available=0, reserved=0, aps_flags=0, frequency_band=<FrequencyBand.Freq2400MHz: 8>, mac_capability_flags=<MACCapabilityFlags.AllocateAddress: 128>, manufacturer_code=0, maximum_buffer_size=80, maximum_incoming_transfer_size=160, server_mask=10752, maximum_outgoing_transfer_size=160, descriptor_capability_field=<DescriptorCapability.NONE: 0>, *allocate_address=True, *is_alternate_pan_coordinator=False, *is_coordinator=False, *is_end_device=True, *is_full_function_device=False, *is_mains_powered=False, *is_receiver_on_when_idle=False, *is_router=False, *is_security_capable=False)",
  "endpoints": {
    "1": {
      "profile_id": 260,
      "device_type": "0x0301",
      "in_clusters": [
        "0x0000",
        "0x0001",
        "0x0003",
        "0x0201"
      ],
      "out_clusters": [
        "0x0003",
        "0x0402"
      ]
    }
  },
  "manufacturer": "Private",
  "model": "Private",
  "class": "zigpy.device.Device"
}
```

and a temperature sensor with input cluster:
```json
{
[...]

PR #844, opened 2021-11-05
Implement missing clusters

majkrzak:
In #843 I noticed that couple of cluster definitions do not exist, and are just a `#TODO` comments. Here is the simple blind implementation. Not yet tested, as I'm not sure how to easily do it on the live HA instance, and will probably need to set up testing one, which I don't have time now.

### DIFF
--- a/zigpy/zcl/clusters/measurement.py
+++ b/zigpy/zcl/clusters/measurement.py
@@ -118,11 +118,79 @@ class OccupancySensing(Cluster):
     client_commands = {}
 
 
-# TODO: 0x0407: Leaf Wetness
-# TODO: 0x0408: Soil Moisture
-# TODO: 0x0409: pH Measurement
-# TODO: 0x040A: Electrical Conductivity
-# TODO: 0x040B: Wind Speed Measurement
+class LeafWetness(Cluster):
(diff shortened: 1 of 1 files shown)


puddly:
Thanks!

> Not yet tested, as I'm not sure how to easily do it on the live HA instance

In addition to these changes to zigpy, ZHA would need to be updated to [support the new clusters](https://github.com/home-assistant/core/blob/8a1d80c60933f07c8325bcc37ac51a3c6dd37d5b/homeassistant/components/zha/sensor.py#L330).

You may find this comment thread useful, it describes how to setup an editable copy of ZHA: https://github.com/zigpy/zigpy/issues/826#issuecomment-947850313. You should be able to also change the zigpy dependency in that same `manifest.json` from `zigpy==0.39.0` to `git+https://github.com/majkrzak/zigpy@patch-1` so that HA auto-installs your zigpy branch on startup.

Adminiuga (approved):
Lgtm

majkrzak:
> In addition to these changes to zigpy, ZHA would need to be updated to support the new clusters.

I knew that this change can not be that easy 😅

PR #847, opened 2021-11-08
Added description attribute to OnOff cluster

jacekk015:
Added description attribute to OnOff cluster to use it later in ZHA and zha-device-handlers quirks

### DIFF
--- a/zigpy/zcl/clusters/general.py
+++ b/zigpy/zcl/clusters/general.py
@@ -396,6 +396,7 @@ class OnOff(Cluster):
     name = "On/Off"
     ep_attribute = "on_off"
     attributes = {
+        0x001C: ("description", t.CharacterString),
         0x0000: ("on_off", t.Bool),
         0x4000: ("global_scene_control", t.Bool),
         0x4001: ("on_time", t.uint16_t),


puddly:
Is `0x001C` a valid attribute for the OnOff cluster? I'm not seeing it in the [ZCL v7 spec](https://github.com/Koenkk/zigbee-herdsman/blob/master/docs/Zigbee%20Cluster%20Library%20Specification%20v7.pdf).

Adminiuga:
Where did you get this attribute? I don't see it in ZCL specs v7

MattWestb:
I have one different stand point **if** the attribute is sent with tuya:s manufacture code it shall being OK by ZCL V7 2.3.3.  _Add manufacturer specific attributes to a standard cluster_.

Adminiuga:
Yeah, we are not adding manufacturer specific attributes to ZCL clusters in zigpy. Those should be added in quirks for that specific device.

MattWestb:
May it be possible implanting one exception with manufacture "ZHA-Quirk" for making some things possible without braking other things and not doing large code in the quirks and ZHA for work around ?

I dont like doing one "Z2M" that have braking all things that is possible for getting easy functions in there system only one way getting extended information from quirks to ZHA without braking  the Zigbee spec.

The maintainers is still have the last word and responsibility what is OK putting in or not for not braking Zigbee specifications (not only ZCL 7) and HA and what is best for the system in the long run.

And this method of using is one work around for luck of manufacture attributes and command in ZHA (from sensor.py):
```
@MULTI_MATCH(
    channel_names=CHANNEL_THERMOSTAT,
    manufacturers="Zen Within",
    stop_on_match=True,
)
```
Is little limited and can making things not possible implanting or making use of large code for work around the limited its having.

By open for moderate moderated user of "ZHA-Quirk" as manufacture open many possibility implanting things that is not possible to day but it shall being used with care.

Way not implanting manufacture code for attributes and commands like the Zigbee spec is saying or using for not standard things but moderating use of it ?

In the end the strengths and weakness of ZHA is the tight Zigbee spec to following but may not closing all doors for getting better functionality in the system.

jacekk015:
> Yeah, we are not adding manufacturer specific attributes to ZCL clusters in zigpy. Those should be added in quirks for that specific device.

Any example?

MattWestb:
Sorry for spamming this PR more !!

I think implanting "small cosmetic" things with the method used by ZHA is very OK then most normal zigbee devices is having one good zigbee ground functions that is not needing so much "plastic operations" for working good in ZHA.

The problem is tuya DP devices that is being one simple on/off plug (with no function and only tunneling tuya MQTT commands to cloud) and is the other end one TRV with very many complex functions and commands that is not compatible with Zigbee standard.

Perhaps shall Zigpy only tunneling the tuya DP protocol to ZHA and ZHA is doing the magic ?

I have asking one devs that is implanting tuya API in one other system but the new one is still not supporting adding Zigbee device.

Having the tuya DP decoding and making complex devices in the level over zigpy and not squeezing them thru Zigbee standard and loosing 75% of there function is more logic in my eyes.

But in the end im only one user and is not steering this ship.

Hedda:
What about ZCL v7 specification section 3.10.1.3 and 3.10.2.1 as examples of derived cluster identifiers and dependencies?

That is, could this attribute be "derived" from another base cluster if it is dependent on that other cluster?

server requirements and dependencies are derived from the base cluster.

https://github.com/Koenkk/zigbee-herdsman/raw/master/docs/Zigbee%20Cluster%20Library%20Specification%20v7.pdf

### 3.10.1.3 Cluster Identifiers

Derived cluster specifications are defined elsewhere. This base cluster specification MAY be used for generic level control; however, it is recommended to derive another cluster to better define the application and domain requirements. If one of more derived cluster identifiers and the base identifier exists on a device endpoint, then they SHALL all represent a single instance of the device level control. See Chapter 2 – Instance Model for more information.

Identifier | Hierarchy | Name
-- | -- | -- 
0x0008 | Base | Level (this cluster specificiation)
0x0008 | Derived | Level Control for Lighting (3.19)
0x001c | Derived | Pulse Width Modulation (3.20)

## 3.10.2 Server

###  Dependencies

For many applications, a close relationship between this cluster and the On/Off cluster is needed. This section describes the dependencies that are required when an endpoint that implements this server cluster and also implements the On/Off server cluster. 

The OnOff attribute of the On/Off cluster and the CurrentLevel attribute of the Level Control cluster are intrinsically independent variables, as they are on different clusters. However, when both clusters are implemented on the same endpoint, dependencies MAY be introduced between them. Facilities are provided to introduce dependencies if required.

Hedda:
FYI; guess this discussion was started when @rforro posted in https://github.com/zigpy/zha-device-handlers/issues/1061 with a proposed solution to use a derived attribute from the second device into the name of a first device that was added before without repairing it:

https://github.com/zigpy/zha-device-handlers/issues/1061#issuecomment-962846745

rforro also referred to that in related https://github.com/home-assistant/core/pull/59347 with the comment:

"The name gets updated automatically, you don't need to repair the device."

Adminiuga:
> Any example?

see https://github.com/zigpy/zha-device-handlers/blob/dcb30aaf9409e394de773a3f6a004976c3c0249a/zhaquirks/philips/__init__.py#L75-L82

Issue #842, opened 2021-11-01
[RFC] Standard radio settings API


puddly:
I propose changing the radio library API to separate high-level network operations from the low-level radio details:

 - `connect()`: connects to the serial port and verifies that it's compatible with the radio library. Does not start the network. This would also allow `probe` to be moved into zigpy (or removed entirely), since it's basically `connect()` in a `try:` block.
 - `disconnect()`: shut down the radio (if it's running) and release the serial port. Not all radios have watchdog timers and thus will continue to run after the serial port loses connection.
 - `load_network_info(load_devices=False)`: populates `self.state.network_information` and `self.state.node_information`. The `load_devices` keyword argument allows for "slow" operations that provide supplemental network information (like scanning TCLK tables) to be done only when necessary.
 - `write_network_info(network_info, node_info)`: Writes the provided network and node information to the radio hardware. Not all available information can be used by every radio so warnings should be logged where appropriate when information is ignored.
 - `start_network()`: start the network with settings currently stored in the stick.  Zigpy's `startup` would then essentially be something like this:
    ```Python
    await self.connect()

    try:
        network_info, node_info = await self.load_network_info(False)
    except NetworkNotFormed:
        if auto_form:
            await self.form_network()
        else:
            raise RuntimeError("Network is not formed")

    if info != info_in_db:
        if strict:
            raise RuntimeError("Network settings don't match what's in the database")
        else:
            LOGGER.warning("Network settings don't match what's in the database")

    await self.start_network()
    await self.permit(0)
    ```

This provides a standard interface for downstream applications (e.g. ZHA and zigpy-cli) to manage network settings independent of the radio hardware. It also removes the need to separately maintain identical network formation code in each radio library.

Below is an example network formation API for zigpy that maintains compatibility with the current YAML configuration options:

```Python

# Implemented by the radio library
def stack_specific_formation_settings(self):
    return {"zstack": {"tclk_seed": os.urandom(16).hex()}}

async def form_network(self, *, start_after_formation=False):
    # First, make the settings consistent and randomly generate missing values
    channel = self.config[conf.CONF_NWK][conf.CONF_NWK_CHANNEL]
    channels = self.config[conf.CONF_NWK][conf.CONF_NWK_CHANNELS]
    pan_id = self.config[conf.CONF_NWK][conf.CONF_NWK_PAN_ID]
    extended_pan_id = self.config[conf.CONF_NWK][conf.CONF_NWK_EXTENDED_PAN_ID]
[...]

PR #848, opened 2021-11-10
New radio API

puddly:
This is a WIP implementation of #842. Database changes will come in a future PR.

Since we're going to have to edit every radio library to accommodate these changes anyways, I've renamed `pre_shutdown` to `shutdown` and also shortened `NetworkInformation` to `NetworkInfo` (and the associated state attributes). The rest of the changes more or less match what's in the linked RFC issue.


Adminiuga:
Should we have an API for setting the manufacturer id ib the node descriptor? EZSP protocol has a command for it. ConBee serial protocol just added a command for it and ZNP would release (unless it already happened) a modification to firmware to allow setting the mfg id.

Right now the only unknown if it should be set permanently or dynamically. 

Thoughts?  

@dmulcahey  for visibility

puddly:
Sure, I don't see why not.  The deCONZ serial spec doesn't say what part of the node descriptor response is configurable so maybe it would allow zigpy to respond with the whole thing?

> and ZNP would release (unless it already happened) a modification to firmware to allow setting the mfg id.

It's baked into the firmware, unfortunately: https://github.com/Koenkk/Z-Stack-firmware/blob/62692f2c3d06054eaa3a263ecda0b573fe60223e/coordinator/Z-Stack_3.x.0/firmware.patch#L525

dmulcahey:
> Should we have an API for setting the manufacturer id ib the node descriptor? EZSP protocol has a command for it. ConBee serial protocol just added a command for it and ZNP would release (unless it already happened) a modification to firmware to allow setting the mfg id.
> 
> Right now the only unknown if it should be set permanently or dynamically.
> 
> Thoughts?
> 
> @dmulcahey for visibility

should we make this an option in the UI in HA? Documented Flow could be something like:

1. User sets mfg code to whatever it needs to be
2. User pairs device
3. User sets mfg code back to 0x0000

Or should we look for a way to make this dynamic similar to what Z2M did In the TI firmware so that users don't have to deal with it at all...

Adminiuga:
Need to try what does setting the mfg id to 0x0000 on ezsp do. Does it actually sets to all zeroes or resets to the standard one.

puddly:
> should we make this an option in the UI in HA? 

Are there any other devices like these Aqara sensors?  Seems like an awfully specialized user-facing setting.

> Or should we look for a way to make this dynamic similar to what Z2M did In the TI firmware so that users don't have to deal with it at all...

What happens if we globally set an Aqara manufacturer ID on startup and never change it back to the stack default?

dmulcahey:
> > should we make this an option in the UI in HA?
> 
> Are there any other devices like these Aqara sensors? Seems like an awfully specialized user-facing setting.
> 
> > Or should we look for a way to make this dynamic similar to what Z2M did In the TI firmware so that users don't have to deal with it at all...
> 
> What happens if we globally set an Aqara manufacturer ID on startup and never change it back to the stack default?

I think it’s all E series Aqara sensors. I’m not sure what happens if we leave it…

MattWestb:
I think i have reading that some other device need having replay with there manufacture number for some requests for working OK but i dont remember winch device it was (I think it was in deCONZ form).

Adminiuga:
This is what Deconz does. I guess they leave it up to the application to send the response?
https://github.com/dresden-elektronik/deconz-rest-plugin/pull/5080 requires new [firmware](https://github.com/dresden-elektronik/deconz-rest-plugin/wiki/Firmware-Changelog#conbee-ii-and-raspbee-ii-r21)

puddly:
How do you feel about moving probing and the bellows baudrate override into ZHA?

```Python
RADIO_PROBE_CONFIGS = {
    bellows.zigbee.ControllerApplication: [{}, {zigpy.config.CONF_DEVICE_BAUDRATE: 115200}],
    zigpy_deconz.zigbee.ControllerApplication: [{}],
    zigpy_znp.zigbee.ControllerApplication: [{}],
    zigpy_zigate.zigbee.ControllerApplication: [{}],
    zigpy_xbee.zigbee.ControllerApplication: [{}],
}


async def find_compatible_radio(path: pathlib.Path | str) -> tuple[
    zigpy.application.ControllerApplication, ConfigType
]:
    for radio, overrides in RADIO_PROBE_CONFIGS.items():
        for override in overrides:
            config = {zigpy.config.CONF_DEVICE_PATH: str(path), **override}
            radio = ControllerApplication(config)

            try:
                await radio.connect()
            except Exception:
                LOGGER.debug("Failed to probe %s with %s: %s", config, radio, exc_info=True)
            else:
                return radio, config
            finally:
                await radio.disconnect()
    else:
        raise ValueError("Failed to find compatible radio module for %s", path)
```

Adminiuga:
> How do you feel about moving probing and the bellows baudrate override into ZHA?

I don't have strong preference, the idea was that the "radio lib" would have specific knowledge to auto detect settings were applicable and then in this case should return the config back to ZHA. in other words ZHA is not supposed to know all possible radio configs, current of the future. The radio lib may poses that knowledge.

puddly:
That makes sense.  My main goal was to remove unnecessary duplication between the radio libraries (all of them just connect/disconnect) so I guess something like this would work too as an override in the radio library:

```Python
@classmethod
async def probe(cls, device_config):
    result = await super().probe(device_config)

    if result:
        return result

    return await super().probe({**device_config, CONF_DEVICE_BAUDRATE: 115200})
```

On the other hand, each radio library may have different probe timeouts (e.g. ZNP's startup time is a lot longer than bellows) so maybe it's best to just leave it alone.

pipiche38:
> This is what Deconz does. I guess they leave it up to the application to send the response? [dresden-elektronik/deconz-rest-plugin#5080](https://github.com/dresden-elektronik/deconz-rest-plugin/pull/5080) requires new [firmware](https://github.com/dresden-elektronik/deconz-rest-plugin/wiki/Firmware-Changelog#conbee-ii-and-raspbee-ii-r21)

If I can share what Zigate is doing, and that was a surprise when I saw 0000 when using the first ZNP one. Zigate is set to 0x1147.
The fix which has been developed is the same as the one from @koenk.

The best would be indeed to be able to set the manufacture code on the fly via an API. I would imagine to do it when receiving an handle_join() then I would set the manufacturer to whatever is needed so at the time the device query it, it will get what I put.

Hedda:
> > This is what Deconz does. I guess they leave it up to the application to send the response? [dresden-elektronik/deconz-rest-plugin#5080](https://github.com/dresden-elektronik/deconz-rest-plugin/pull/5080) requires new [firmware](https://github.com/dresden-elektronik/deconz-rest-plugin/wiki/Firmware-Changelog#conbee-ii-and-raspbee-ii-r21)
> 
> If I can share what Zigate is doing, and that was a surprise when I saw 0000 when using the first ZNP one. Zigate is set to 0x1147. The fix which has been developed is the same as the one from @koenk.
> 
> The best would be indeed to be able to set the manufacture code on the fly via an API. I would imagine to do it when receiving an handle_join() then I would set the manufacturer to whatever is needed so at the time the device query it, it will get what I put.

FYI, [koenk](https://github.com/koenk) is the wrong Dutch guy if you meant @Koenkk from https://github.com/Koenkk/Z-Stack-firmware and [Zigbee2MQTT](https://github.com/Koenkk/zigbee2mqtt) ;)

SylvainPer:
@puddly gentle reminder on connect() missing function in startup as we discussed it on zigpy-znp repo

pipiche38:
Any plan or targeted date when it will be released ?

pipiche38 (commented):
@puddly,  while I understand Endpoint 0x01, what is the purpose of Ep: 0x02 ?

```
    await self.add_endpoint(
            zdo_types.SimpleDescriptor(
                endpoint=2,
                profile=zigpy.profiles.zll.PROFILE_ID,
                device_type=zigpy.profiles.zll.DeviceType.CONTROLLER,
                device_version=0b0000,
                input_clusters=[zigpy.zcl.clusters.general.Basic.cluster_id],
                output_clusters=[],
            )
```

puddly:
Alright, this PR is about as ready as it will ever be. Any final thoughts before I merge this and the corresponding PRs in all of the other radio libraries?

SylvainPer:
Let's go!

dumpfheimer:
For what its worth: I just instelled zigpy, zigpy-znp and bellows from the new radio settings branches:
Everything seems to be **working perfectly fine**.

1. I previously had issues with pairing new devices using a specific router (possible related to https://github.com/zigpy/zigpy-znp/issues/148) which seem to have disappeared.
2. Many/Most battery driven devices turned up "unavailable" until I pressed a button or they checked in by themselves. Also, I had to re-joined 3 or 4 remotes that were not used in a long time - not sure when they were lost. I believe those were the devices related to error messages I got ( `Status.NWK_INVALID_REQUEST` ). Not an issue, just wanted to mention it.
3. One more random error message of a device that is behaving well: `[zigpy.appdb] Error handling '_save_attribute' event with (a4:c1:38:8d:f0:52:aa:f9, 1, 0, 5, 'TS011F') params: FOREIGN KEY constraint failed`
 
This might be a psychogical thing or maybe it is due to the fact that the network was offline for 30m but I feel like the network is more **responsive and reliable**

Issue #535, opened 2020-10-29
[REQUEST] grab OTA files from https://github.com/Koenkk/zigbee-OTA/blob/master/index.json


grischard:
@Koenkk maintains a nice convenient list of OTA firmwares at https://github.com/Koenkk/zigbee-OTA

Would it be possible to automatically support grabbing OTA updates from https://github.com/Koenkk/zigbee-OTA/blob/master/index.json ?

Hedda:
FYI, there is an existing discussion about OTA updates in https://github.com/zigpy/zigpy/issues/154 which might be relative or even superseed this?

grischard:
Definitely related, but not superseded. It's good that the tickets are separate but linked.

almico:
I tried to download images from https://github.com/Koenkk/zigbee-OTA/tree/master/images/Xiaomi and all of them turned to be invalid OTA files according to zigpy:

```
DEBUG (SyncWorker_2) [zigpy.ota.provider] File '/config/zigpy_ota/OTA_lumi.flood.agl02_V1.0.18_20190814.20191008104903.ota' doesn't appear to be a OTA image
```

Did I miss something?

puddly:
They parse fine for me. What is the full error traceback?

almico:
It should be this:

```
2021-06-29 05:24:11 DEBUG (SyncWorker_13) [zigpy.ota.provider] File '/config/zigpy_ota/1600167449-si32_zg_uart_connect_sleep_ZS5_ty_OTA_1.1.5.bin' doesn't appear to be a OTA image
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/zigpy/ota/image.py", line 262, in parse_ota_image
    return HueSBLOTAImage.deserialize(data)
  File "/usr/local/lib/python3.8/site-packages/zigpy/ota/image.py", line 201, in deserialize
    header, remaining_data = OTAImageHeader.deserialize(data)
  File "/usr/local/lib/python3.8/site-packages/zigpy/ota/image.py", line 112, in deserialize
    raise ValueError(
ValueError: Wrong magic number for OTA Image: 168430090

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/zigpy/ota/provider.py", line 302, in scan_image
    parsed_image, _ = parse_ota_image(f.read())
[...]

PR #849, opened 2021-11-10
Central HTTP client with global cache and per-device OTA handling

puddly:
This is more of a test than a real PR at the moment but I've implemented a centralized wrapper around aiohttp (with associated unit testing infrastructure) to allow for all explicit locks and concurrency controls to be removed from OTA handlers.

 - Concurrent requests for the same resource are combined and only a single one will ever be sent.
 - All responses are cached.

My idea is to implement OTA handling per-device within quirks and to make the OTA image downloaders very straightforward:

```Python
async def download_ota_image(
    self,
    *,
    image_type,
    manufacturer_id,
    current_file_version,
    hw_version,
):
    rsp = await CACHED_HTTP_CLIENT.get(
        url="http://fw.ota.homesmart.ikea.net/feed/version_info.json",
        headers={"accept": "application/json;q=0.9,*/*;q=0.8"},
        cache_for=60 * 60 * 24,
    )

    # IKEA does not always respond with an appropriate Content-Type but the response is
    # always JSON
    fw_lst = await rsp.json(content_type=None)

    for fw in fw_lst:
        if "fw_file_version_MSB" not in fw:
            continue

        if fw["fw_manufacturer_id"] != manufacturer_id:
            continue

        if fw["fw_image_type"] != image_type_id:
            continue

        file_version = (fw["fw_file_version_MSB"] << 16) | fw["fw_file_version_LSB"]

        if fw["current_file_version"] <= file_version:
            continue

        break
    else:
        LOGGER.debug("No compatible firmware found")
        return

    fw_rsp = await CACHED_HTTP_CLIENT.get(fw["fw_binary_url"], cache_for=60 * 60 * 24)
    fw_data = await fw_rsp.read()

    assert len(fw_data) == fw["fw_filesize"]

    ota_image, _ = parse_ota_image(fw_data)

    return ota_image
```

In the above handler, the only "slowdown" with a filled cache would be parsing the cached JSON responses each time, which is fast enough to not be an issue (92 µs on my relatively slow hardware).

Related issues and PRs: #827, #839


puddly:
On the other hand, so few manufacturers even provide a way to download updates over HTTP that it's probably wasted bandwidth to poll them every day for their bi-yearly updates. Perhaps it would be better to externally download and re-host all of these OTA images, even for IKEA and the others with HTTP APIs, and just serve the OTA files themselves with a central index.

We could even provide a "beta" channel to prevent supposedly-stable updates from breaking existing devices, like the recent IKEA remote issue with group binding.

MattWestb:
If not possible hosting the files of legal grounds then making one index with firmware links and classes stable production, pre production and beta and the clients can pulling the index file some time a day and only the firmware then needing them (i hope is not one copy of how Z2M is doing).

Some manufactures is having more version online if knowing there locations perhaps it can being interesting have the links to them saved in the index if user like using one old Light Link firmware on one IKEA light.

If saving device current version of the device firmware version in Zigbee.db (with image number the devices is requesting) then can also having one record for one absolute version or stable production / pre production or latest beta firmware and disabled and setting it from "custom attribute" on the OTA cluster.
The OTA only need looking for the device current version and the wonted version and if its need updating using the firmware from the index or local and updating the device.

Interesting but i spamming in the wrong place as normal !!

Adminiuga:
> Perhaps it would be better to externally download and re-host all of these

Are you allowed.to re-hostnota images? We could host an external index file though.

What if we decouple OTA updates from zigpy completely? And instead use some form of inter process communication: ws or rest where zigpy just asks a local OTA server for an update and provides all relevant information, like: IEEE, image type, version, model, manufacturer, manufacturer id, etc and get the image back?

Then this external ota server could do all caching, UI and even do downgrades for specific devices based on IEEE?

puddly:
> If not possible hosting the files of legal grounds

As far as I can tell, Z2M has not had any issues hosting images. I'm not even sure if there's really a legal restriction to begin with: any manufacturer worried about users peeking at their firmware is free to encrypt the OTA images.

> What if we decouple OTA updates from zigpy completely?

What do you think about middlewares or plugins instead of RPC?  That way, we could independently have a `zigpy-ota` package that has full access to the zigpy API and `Device` object internals, without dealing with headaches stemming from serialization and managing a second process.  Might even be a way to spin off a simple `zigpy-webui` package or something...

Adminiuga:
I'm not so keen on the plugins and giving access to internal structures, as then users would run all the crap and keep opening zigpy issues

Adminiuga:
Really just trying to avoid adding more and more OTA providers to zigpy. Otherwise probably could go with keeping and external index, same as z2m

davet2001:
I like the idea of having a big json file defining all the manufacturer IDs, image_id and image urls.  Then the effort would be how to add a particular vendor's upgrade table to the list.  In the case of salus, that would be by extracting every image to get the data out of the headers.

Perhaps a hash of each ota image could be added to the table to reduce the chance of malicious firmware updates.

I don't think there is any need to store firmware images either in the git repo or to host them.  Just point to the vendor download server.

Should I put my salus work on hold in the mean time?  At the moment it requires changes to zigpy that wouldn't be needed if the above was implemented.

Hedda:
One additional related question raised https://github.com/zigpy/zigpy/issues/535 and https://github.com/zigpy/zigpy/issues/154 was whether or not OTA firmware image files without URL source could be hosted in its own repo under the zigpy org similar to the https://github.com/Koenkk/zigbee-OTA repo? E.g. maybe begin by forking theirs to https://github.com/zigpy/zigpy-zigbee-OTA if that would make it easier to maintain a copy of the same OTA images?

> I don't think there is any need to store firmware images either in the git repo or to host them. Just point to the vendor download server.

FYI, the reason why copies of some images are stored in  https://github.com/Koenkk/zigbee-OTA is that those specific OTA images are not downloadable via a public URL. Instead, those images are either manual backups via JTAG or OTA images that have been lifted on-the-fly by Zigbee sniffing during OTA updates from a branded hub/gateway/bridge. Those it does not contain copies of OTA images that are available via public URLs, as noticed by the lack of OTA images from example Philips, IKEA, and Salus-> https://github.com/Koenkk/zigbee-OTA/tree/master/images (i.e. there is simply no public vendor download server to point to).

Hedda:
> One additional related question raised #535 and #154 was whether or not OTA firmware image files without URL source could be hosted in its own repo under the zigpy org similar to the https://github.com/Koenkk/zigbee-OTA repo? E.g. maybe begin by forking theirs to https://github.com/zigpy/zigpy-zigbee-OTA if that would make it easier to maintain a copy of the same OTA images?
> 
> > I don't think there is any need to store firmware images either in the git repo or to host them. Just point to the vendor download server.
> 
> FYI, the reason why copies of some images are stored in https://github.com/Koenkk/zigbee-OTA is that those specific OTA images are not downloadable via a public URL. Instead, those images are either manual backups via JTAG or OTA images that have been lifted on-the-fly by Zigbee sniffing during OTA updates from a branded hub/gateway/bridge. Those it does not contain copies of OTA images that are available via public URLs, as noticed by the lack of OTA images from example Philips, IKEA, and Salus-> https://github.com/Koenkk/zigbee-OTA/tree/master/images (i.e. there is simply no public vendor download server to point to).

Should add to zigbee-herdsman-converters library by Koenkk which is used in Zigbee2MQTT has a OTA library similar to zigpy's OTA providers -> https://github.com/Koenkk/zigbee-herdsman-converters/tree/master/lib/ota but otherwise I understand that zigbee-herdsman-converters itself is otherwise more of the equivalent to zigpy's [zha-device-handlers](https://github.com/zigpy/zha-device-handlers)

MattWestb:
Z2M have little to much "cooked" firmware like for Xiaomi that is not fabric ones and have problem that the manufacture and model is not unique and user have flashing them and braking there devices also with some tuya TRVs upgrade for the Zigbee module without matching the device was braking many devices for users. 
So its not only IKEA test feed that braking devices is not using verified recommended firmware from the manufacture (IKEA was more or less self made problem then using test feed that is not stable production ones).

davet2001:
I think this could be a problem.  

There are probably several levels of users:

A) only wanting to do FW upgrades that are explicitly triggered by the vendor's proprietary hardware/hub

B) wanting to do FW upgrades that are available from standard public vendor download pages

C) wanting to do FW upgrades that are from the vendor but have been 'found' by non standard methods such as sniffing traffic or dumping flash.

D) wanting to use the latest community/experimental/test FW that is newest but not recommended/endorsed by the vendor.

Then with reports of non-unique device IDs, products being bricked and vendors not sticking to rules this all seems v risky.  The vendor can be expected to keep A) and B) in order, but not C) or D).

If a vendor finds a bug they might make that image unavailable for download.

How are community supplied images vetted?

Perhaps a multi tier approach would work, where anything above B has to be explicitly turned on.

Hedda:
As you say, the questions about grey areas only comes when the provider source is not using official OTA image URLs servers provided by the manufacturers/vendors/brands. (With the exception when the user themselves manually copy the OTA images to a local directory path to enable ZHA to use those). 

I think that for an OTA provider where the OTA images are hosted/cached on sites there are not officially provided by the manufacturer/vendor/brand (like an unofficial repository GitHub maintain by ZHA/zigpy or zigbee2MQTT community) there would be good to have at least give a warning in ZHA to the user about the fact that there is a risk of corrupting firmware and bricking their device when flashing OTA images from any unofficial providers/sources.

I can think of two approaches that would still make upgrading via OTA user-friendly for such an unofficial provider/source:

1. Always force the users to manually active/start the OTA upgrade process per device via the ZHA UI when the OTA images are from unofficial providers/sources. You could still make it very user-friendly have a pop-up notification that let the user know that there is a firmware upgrade available per device and give a warning why firmware on that device was not automatically upgraded. 

2. Alternatively, make OTA download and upgrade process automatic for OTA images from unofficial providers/sources as well, but make the users explicitly have to enable "unofficial OTA firmware upgrades" for that manufacturer/vendor/brand in ZHA and post a clear warning about the greater risk of using automatic OTA upgrades for unofficial providers/sources.

Even auto OTA firmware upgrades do have to be explicitly enabled per manufacturer/vendor/brand in ZHA and I think that is a good approach to start with (though GUI settings/configurations would be nicer from a user experience perspective rather than having to edit the configuration in YAML). That has not really been an issue before since ZHA/zigpy currently only have OTA providers for IKEA and Ledvance via the manufacturers/vendors/brands official OTA image URLs servers, thus those should at least in theory be perfectly safe sources.

https://www.home-assistant.io/integrations/zha#ota-firmware-updates

Hence adding further OTA providers that use official OTA image URLs servers provided by the manufacturers/vendors/brands should presumably continue to be perfectly safe to use as sources, which then could also easily be explicitly enabled per manufacturer/vendor/brand in ZHA.

Again, having UI settings for all and any such OTA configurations in ZHA would be very nice and make it much more user-friendly, which in turn should hopefully bring more awareness and perhaps in the end even indirectly pressure manufacturers/vendors/brands to provide official sources and servers for OTA images.

davet2001:
Agreed.

If a user copies an OTA file into a directory they are assumed to be a certain level of advanced user and aware of risks.

I don't think a FW upgrade should be an automatic background process.  In my opinion it should be prompted in the UI when a new one is available and the user should do something to start the upgrade (even if just so that people understand why their lights just flashed etc, it also eliminates accidental continuous upgrade cycles in a fault scenario).

So how about this:
- Vendor provided official OTA downloads should be implemented per vendor as per ledvance, ikea.  
  - Ultimately these should prompt on the UI e.g. 'a FW upgrade is available for this device, click here to upgrade now.'
- A community provided list of OTA images should be made available by the method in this PR, 
  - users should explicitly have to turn this on (not enabled by default)
  - it should have warnings
  - It should still prompt e.g. 'click here to upgrade now'

So the community list is like an 'extra' vendor that is off by default.

PR #861, opened 2021-12-04
Enforce minimum SQLite version and use `UPSERT` syntax

### LINKED ISSUE #860: Enforcing minimum SQLite version or using another DB
The issue fixed by #515 seems to be present in other portions of the appdb code (https://github.com/home-assistant/core/issues/57200).

Instead of being held back by older SQLite versions not supporting `UPSERT` and other useful features, perhaps we should enforce a minimum version. Some possibilities:

 1. [`pysqlite3-binary`](https://github.com/coleifer/pysqlite3) is *mostly* `sqlite3`-compatible and wheels exist for Linux.  Windows Python builds are already fine, as is macOS Homebrew Python, so it should fix most problems immediately.
    - The major problem with it is that `aiosqlite` doesn't work with alternative drivers yet (https://github.com/omnilib/aiosqlite/pull/106) and pysqlite3 isn't 100% compatible with sqlite3's API.
 2. Switch to another single file database, like [DuckDB](https://duckdb.org/).  It's more like a [typical SQL databases](https://duckdb.org/docs/sql/introduction) in terms of types and has a first-party Python module that's simply `pip install`-able.
    - It hasn't had a 1.0.0 release yet.
    - Has no async wrapper. On the other hand, we already have a separate synchronous loop to process database events so it's not like we're really using `aiosqlite` to begin with.
 3. Just throw a runtime error and leave it to the user to figure out how to upgrade SQLite.
    - Realistically, who would this affect?

puddly:
Fixes #860

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9, '3.10']
+        python-version: ['3.7.7', 3.8, 3.9, '3.10']
     steps:
       - name: Check out code from GitHub
         uses: actions/checkout@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ on:
 
 env:
   CACHE_VERSION: 1
-  DEFAULT_PYTHON: 3.7
+  DEFAULT_PYTHON: 3.9
   PRE_COMMIT_HOME: ~/.cache/pre-commit
 
 jobs:
@@ -271,7 +271,7 @@ jobs:
     needs: prepare-base
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9, '3.10']
+        python-version: ['3.7.7', 3.8, 3.9, '3.10']
     name: >-
       Run tests Python ${{ matrix.python-version }}
     steps:
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -6,3 +6,5 @@ pytest
 pytest-aiohttp
 pytest-cov
 pytest-timeout
+
+pysqlite3-binary; platform_system=="Linux"
--- a/tests/test_appdb.py
+++ b/tests/test_appdb.py
@@ -769,14 +769,13 @@ async def test_load_unsupp_attr_wrong_cluster(tmpdir):
         ((4, 0, 0), True),
     ],
 )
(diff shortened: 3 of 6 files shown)


puddly:
@TheJulianJES want to test this one out?

```yaml
pypi:
 - git+https://github.com/puddly/zigpy.git@puddly/minimum-sqlite-version
```

I'm not really sure what specific circumstances were required for a quirk-less device to lose its group membership after a re-join, but all of the `INSERT OR REPLACE` queries are gone so I'm hoping this will fix it.

TheJulianJES:
> @TheJulianJES want to test this one out?

Sure, I updated my Home Assistant instance to use your branch now. I'll monitor and see if any members are lost in the database/after a restart. (It might take a week or so to see if anything changed, as a bulb in a group needs to "rejoin".)

dmulcahey:
@TheJulianJES now that this is installed correctly on your end are we convinced it corrects the issue?

TheJulianJES:
@dmulcahey Pretty sure it fixed my issue. I just restarted my Home Assistant instance after another week of uptime and I didn't loose any lights again.

dmulcahey:
@TheJulianJES thanks for confirming / testing. @puddly this may be good to go.

Adminiuga (approved):
Lgtm. Just needs to pass the tests.

puddly:
The new unit tests are unfortunately blocked by https://github.com/Martiusweb/asynctest/issues/152 and won't work with new releases of Python 3.7. The `asynctest` module [is dead](https://github.com/Martiusweb/asynctest/issues/158#issuecomment-785872568) so this won't be fixed.

Possible solutions:

 - Pin Python 3.7.7 in CI instead of installing the latest 3.7 (currently 3.7.12).
 - Delete the new unit tests and just pretend pysqlite3 will work, since no code specific to pysqlite3 is ever used and they pass on all other Python versions. The new tests also increase the runtime for me from 42s to 57s (35% increase).

dmulcahey:
"Rewrite with blocks to work with Python 3.6" - Is there a reason for this? 3.6 went full EOL on 12/23. as for the note above about 3.7... I'd actually be in favor of dropping 3.6 and 3.7 support all together. Thoughts?

Adminiuga:
I'm in favor of dropping 3.6 and 3.7

puddly:
Works for me. I'll pin the version of 3.7 in the GitHub Actions config to make tests pass and then we can remove 3.7 (and its associated `CancelledError` code) in a future PR.

PR #874, opened 2021-12-26
add attributes used by Tuya TS011F

jlayec:
ability to change the behavior of the led on the plug and the power on state on the plug Tuya TS011F also known as LIDL Silvercrest HG06337 / SAPZ 1 A1

for information jeedom format parameters
  "config" : [
    {"endpoint":11,"cluster":6,"attribute":32769,"name":"Light mode","type":"select","values":[
      {"value":0,"name":"Led disabled"},
      {"value":1,"name":"Led enabled"},
      {"value":2,"name":"Led enabled but inverted"}
    ]},
    {"endpoint":11,"cluster":6,"attribute":32770,"name":"Power on state","type":"select","values":[
      {"value":0,"name":"Off"},
      {"value":1,"name":"On"},
      {"value":2,"name":"Last state"}
    ]}
  ],

### DIFF
--- a/zigpy/zcl/clusters/general.py
+++ b/zigpy/zcl/clusters/general.py
@@ -401,6 +401,8 @@ class OnOff(Cluster):
         0x4001: ("on_time", t.uint16_t),
         0x4002: ("off_wait_time", t.uint16_t),
         0x4003: ("start_up_on_off", StartUpOnOff),
+        0x8001: ("light_mode", t.enum8),
+        0x8002: ("power_on_state", t.enum8),
     }
     server_commands = {
         0x0000: ("off", (), False),


puddly:
Thanks for the PR. I'm looking at the ZCL R8 spec and I can't see `0x8001` and `0x8002` as defined attributes for the OnOff cluster. If these are Tuya-specific, this change belongs in https://github.com/zigpy/zha-device-handlers in the appropriate quirk for your device. The `zigpy` package only implements what's in the specification.

MattWestb:
Both is implanted in tuya init https://github.com/zigpy/zha-device-handlers/pull/1105 and later switchmode and switch  configuration.

jlayec:
Thanks for feedbacks. New PR submitted https://github.com/zigpy/zha-device-handlers/pull/1237

MattWestb:
@puddly I have starting looking for ZCL R9 and hoping tuya have getting there "new" attribute in there then tuya is one of the reale heavy in CSA then producing Zigbee device in the world (IKEA and Philips HUE is little smaller if looking on global market) ;-)))

PR #882, opened 2022-01-11
Use correct datatype for `start_up_current_level` attribute

### LINKED ISSUE #881: StartUpCurrentLevel under ZCL LevelControl cluster is wrong data type
Right now, [zcl/clusters/general.py](https://github.com/zigpy/zigpy/blob/dev/zigpy/zcl/clusters/general.py) gives the StartUpCurrentLevel attribute as an enum8 type, but according to the
[Zigbee Lighting and Occupancy Device Specification](https://zigbeealliance.org/wp-content/uploads/2019/11/docs-15-0014-05-0plo-Lighting-OccupancyDevice-Specification-V1.0.pdf) line 1390, it should be an uint8_t.

When trying to write the cluster attribute, I got the following error in my debug log:
> command_id=Command.Write_Attributes_rsp>
> [0x8717](LCD002): set: 253 for attr: 16384 to cluster: 8 for ept: 11 - res: [[WriteAttributesStatusRecord(status=<Status.INVALID_DATA_TYPE: 141>, attrid=16384)]]
> Set attribute for: cluster_id: [8] cluster_type: [in] endpoint_id: [11] attribute: [16384] value: [253] manufacturer: [None] response: [[[WriteAttributesStatusRecord(status=<Status.INVALID_DATA_TYPE: 141>, attrid=16384)]]]

As expected from the error, the modification did not successfully apply to the light.

However, when I changed the zigpy code for zcl/clusters/general.py line 437 to be
`class StartUpCurrentLevel(t.uint8_t): `
instead of the current code
`class StartUpCurrentLevel(t.enum8):`
I am able to modify the cluster attribute with no errors, and the setting is successfully applied to the light.

(This is my first ever GitHub issue for any project, so I'm sorry if I am missing any information! I'm happy to add/clarify anything to help the process.)

puddly:
Fixes #881.

There's currently no nice way to represent a type with "special" values, even though many types actually have special "invalid" values already:

 - `0xFF` for `bool`
 - `0xFF` for `uint8` and `enum8` (!)
   - same for uint16, 32, etc.
 - Strings of length `0xFF`
 - Arrays with length `0xFFFF`

### DIFF
--- a/zigpy/zcl/clusters/general.py
+++ b/zigpy/zcl/clusters/general.py
@@ -434,10 +434,6 @@ class LevelControl(Cluster):
     """Attributes and commands for controlling devices that
     can be set to a level between fully ‘On’ and fully ‘Off’."""
 
-    class StartUpCurrentLevel(t.enum8):
-        Minimum = 0x00
-        PreviousValue = 0xFF
-
     cluster_id = 0x0008
     name = "Level control"
     ep_attribute = "level"
@@ -449,7 +445,7 @@ class LevelControl(Cluster):
         0x0012: ("on_transition_time", t.uint16_t),
         0x0013: ("off_transition_time", t.uint16_t),
         0x0014: ("default_move_rate", t.uint8_t),
-        0x4000: ("start_up_current_level", StartUpCurrentLevel),
+        0x4000: ("start_up_current_level", t.uint8_t),
     }
     server_commands = {
         0x0000: ("move_to_level", (t.uint8_t, t.uint16_t), False),


pfak:
According to the Zigbee specification, should this not hexadecimal? 

image

https://zigbeealliance.org/wp-content/uploads/2019/11/docs-15-0014-05-0plo-Lighting-OccupancyDevice-Specification-V1.0.pdf pg120

puddly:
Hexadecimal is a presentation of the number, internally it's just a `uint8_t`.  Are you having some specific problem with `start_up_current_level`?

MattWestb:
Shall the "imprecation" of the special function being made in the upper system (ZHA) and zigpy only "piping" the data to the upper system ?

I think the upper system shall having the responsibility making it right.
Its the same with light level that  have reserved values in the under and upper part of the range (Z2M is "converting" all to light level also reserved functions).

PR #883, opened 2022-01-12
Update link of zha_custom to active fork

Hedda:
Update link of zha_custom to active fork by @mdeweerd (the original zha_custom project by Adminiuga is still linked in parentes)


mdeweerd:
Note: I created a new project from the zha_custom fork to zha-toolkit avoid confusing with the original sample project and because the 'zha_custom' name is not really representive of what it is now.

Adminiuga:
I'm not sure this even should be in zigpy readme.

Hedda:
> Note: I created a new project from the zha_custom fork to zha-toolkit avoid confusing with the original sample project and because the 'zha_custom' name is not really representive of what it is now.

Closing this then as fork no longer has same name and scope has gone beyond the original.

Issue #873, opened 2021-12-21
Add write_attributes_raw to cluster methods


mdeweerd:
There is a "read_attributes_raw" method that is helpful as it allows reading attributes by using their numerical values.

There is no "write_attributes_raw" method.

It guess this could essentially be:

```python
    async def write_attributes(
        self, attributes: Dict[Union[str, int], Any], manufacturer: Optional[int] = None
    ) -> List:
        args = self._write_attr_records(attributes)
        return self.write_attributes_raw(args,manufacturer)       

    async def write_attributes_raw(
        self,  List[foundation.Attribute],, Any], manufacturer: Optional[int] = None
    ) -> List:
        result = await self._write_attributes(args, manufacturer=manufacturer)
        if not isinstance(result[0], list):
```

I was able to write to the manufacturer attribute like this:
```python
      # cluster is a Cluster instance.
      try:
         attr_value=foundation.TypeValue(0x41,t.LVBytes([0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]))
         attr=foundation.Attribute(0x0000, value=attr_value)
         result=awaitc luster._write_attributes(attr_list,manufacturer=0xF000)
      except asyncio.exceptions.TimeoutError:
         print("Exception on write manf attribute")
```

puddly:
Can you turn this change into a pull request?

mdeweerd:
Ok, I will fork and make & PR.  Leave it open, it will be on my issues & notificatons lists in Github.

Adminiuga:
Perhaps better is to expose the foundation commands to be allowed to called directly

mdeweerd:
Closing this because of the integration of the PR.

PR #887, opened 2022-01-20
#873 Add write_attributes_raw method

mdeweerd:
Implement #873 

Enables writing attributes that are not predefined.
Similar to read_attributes_raw

### DIFF
--- a/tests/test_zcl.py
+++ b/tests/test_zcl.py
@@ -389,8 +389,16 @@ async def test_write_wrong_attribute(cluster):
 
 
 async def test_write_unknown_attribute(cluster):
+    with patch.object(cluster, "_write_attributes", new=AsyncMock()):
+        with pytest.raises(KeyError):
+            # Using an invalid attribute name, the call should fail
+            await cluster.write_attributes({"dummy_attribute": 5})
+        assert cluster._write_attributes.call_count == 0
+
+
+async def test_write_unknown_attribute_dual_mocktest(cluster):
     with patch.object(cluster, "_write_attributes", new=AsyncMock()), patch.object(
-        cluster, "write_attributes", new_AsyncMock()
+        cluster, "write_attributes", new=AsyncMock()
     ):
         # Using an invalid attribute name, the call should fail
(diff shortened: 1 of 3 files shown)


zigpy/zcl/__init__.py
```diff
@@ -374,7 +374,14 @@ def _write_attr_records(
     async def write_attributes(
         self, attributes: Dict[Union[str, int], Any], manufacturer: Optional[int] = None
     ) -> List:
+        """Write attributes to device with internal 'attributes' validation"""
         args = self._write_attr_records(attributes)
+        return self.write_attributes_raw(args, manufacturer)
+
+    async def write_attributes_raw(
+        self, args: List[foundation.Attribute], manufacturer: Optional[int] = None
```
puddly, line 382: `args` was an internal variable name but making it a parameter necessitates a more useful name.  Could you rename it to `attrs`?

mdeweerd: I renamed it, but I have to review the test cases I added.

tests/test_zcl.py
```diff
@@ -388,12 +388,28 @@ async def test_write_wrong_attribute(cluster):
         assert cluster._write_attributes.call_count == 1
 
 
+async def test_write_unknown_attribute(cluster):
+    with patch.object(cluster, "_write_attributes", new=AsyncMock()):
+        # Using an invalid attribute name, the call should fail
+        res = await cluster.write_attributes({"dummy_attribute": 5})
+        assert isinstance(res.exception(), KeyError)
```
puddly, line 395:
```
with pytest.raises(KeyError):
    await cluster.write_attributes({"dummy_attribute": 5})
```

mdeweerd: Thanks, this worked (tried another method as well that didn't).

mdeweerd:
I've updated the branch.  The tests pass in my fork, but some executions fail because of uploads to https://coveralls.io/ .
There is also a minor change to the README.md because I wanted to see if that would trigger the workflow in my fork.

PR #892, opened 2022-01-24
Uniform cluster cmd

mdeweerd:
The function signatures for Cluster.command and CustomCluster.command were different so I ran into a problem where 'tries' was present in the function call, but not for CustomCluster.command.

A quick search for tests of CustomCluster.command did not reveal anything, so I did not add any specific test for this.


mdeweerd:
I am  a bit surprised that all the changes already integrated in #887 are listed here as well.  The only update is the last one e3de7c7 ;

PR #896, opened 2022-01-29
Use correct values for ZDO leave request flags

### LINKED ISSUE #831: [Request] Command for sending Leave request with rejoin flag.
Some "interesting" devices (tuya) is start using EFR32MG21 and have finding out that they is having much hardware to playing with and have implanting "mode changing" by custom attribute is written to the device.
The problem device (TS004F) is one pity Zigbee dimmer switch with one normal EP 1 and sending commands from its on/off and level cluster with broadcast.
Then the working mode is changed its having 4 end points (1-4) and only using on/off cluster for sending 0, 1 and 2 (Short, double and hold ala tuya).

So fare so good and if sending the "mode change " command the device is partly changing mode but is not initiating the extra endpoints.

From sniffing tuya ZBGW doing the same thing but they is sending one leave command with the flag rejoin. and then setting the mode then the device have rejoining and is in "init mode".
After receiving the command the device is start sending command in unicast from cluster 0x0006 on all 4 endpoints.

I have reading that some Philips HUE lights can also changing some parameters but only then them is in INIT mode and its looks like the "secret" is sending one leave with rejoin flag for getting the device in init mode.

The problem is we have not finding any way sending the leave with rejoin that tuya is doing.
Zigpy can sending leave command https://github.com/zigpy/zigpy/blob/a19a2a063d5edacb46722aeb56d538badd37fcc0/zigpy/zdo/types.py#L456
Also by basic cluster but all is **without** the rejoin flag so its not working.

So that the short long story for the request.

Im very sure its possible sending the command in raw form on some coordinators and can being OK for testing but in very near future its needed be done in one quirk for doing the mode changing in "user mode".

And also the problem that the device is changing its endpoints configuration can making problem for matching quirks (or only doing 2 different for the EP config) and also reading the new endpoints without rejoining the device the device then is always starting in dimmer mode after being rested and rejoined and need doing it after the change mode procedure.

Some of the findings and posted sniffs is in https://github.com/zigpy/zha-device-handlers/pull/969.

I think its good if some of our core devs is getting one TS004F for testing and is have one very nice EFR32MG21 module that can being flashed with better firmware :-)) 
(warning the "old" TS0044 is not having the same module and its firmware cant changing mode)

The user from Hubitat and Z2M is the brains that have sniffing the device and tuya TBGW and also trying getting the mode change working in our community systems.

Thanks in advance for help and feed back !!

Adminiuga:
Fixes #831 

Thanks to @mattwest and @puddly  for identifying the issue and fix.

### DIFF
--- a/tests/test_application.py
+++ b/tests/test_application.py
@@ -280,7 +280,7 @@ async def test_known_device_left(app, ieee):
 async def _remove(
     app, ieee, retval, zdo_reply=True, delivery_failure=True, has_node_desc=True
 ):
-    async def leave():
+    async def leave(*args, **kwargs):
         if zdo_reply:
             return retval
         elif delivery_failure:
--- a/tests/test_zdo.py
+++ b/tests/test_zdo.py
@@ -85,9 +85,9 @@ async def test_unbind(zdo_f):
     "remove_children, rejoin, flags",
     (
         (False, False, 0),
-        (False, True, 1),
-        (True, False, 2),
-        (True, True, 3),
+        (False, True, 0x80),
+        (True, False, 0x40),
+        (True, True, 0xC0),
     ),
 )
 async def test_leave(zdo_f, remove_children, rejoin, flags):
--- a/zigpy/application.py
+++ b/zigpy/application.py
@@ -141,7 +141,9 @@ class ControllerApplication(zigpy.util.ListenableMixin, abc.ABC):
             device.neighbors.add_context_listener(self._dblistener)
         self.listener_event("device_initialized", device)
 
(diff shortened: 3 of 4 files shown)


MattWestb:
Thanks **A**
@mdeweerd was pinpointing it was working on IT but we was not getting it working on EZSP.

MattWestb:
By the way wrong matt addressed ;-)

Adminiuga:
> By the way wrong matt addressed ;-)

Oops. Sorry "wrong" Matt :)

MattWestb:
That's the life S . . . happens sometime  ;-)

PR #897, opened 2022-02-01
Do not remove children when the parent device is removed

puddly:
I think #896 defaulting to `remove_children=True` will cause unintended behavior:

> This field has a value of 1 if the device being asked to leave the network is also being asked to remove its child devices, if any. Otherwise, it has a value of 0.

### DIFF
--- a/zigpy/application.py
+++ b/zigpy/application.py
@@ -142,7 +142,7 @@ class ControllerApplication(zigpy.util.ListenableMixin, abc.ABC):
         self.listener_event("device_initialized", device)
 
     async def remove(
-        self, ieee: t.EUI64, remove_children: bool = True, rejoin: bool = False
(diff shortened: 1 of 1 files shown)


Adminiuga:
Have you tried to capture the traffic by any chance? Does the parent actually sends a "disassociate" request to children? Or does it just "clear" the parent's table of all children?

puddly:
Hmmm. I tried it and the removed router's end device re-joined the network. No traffic was sent by the router to the end device. What exactly does this option do?

MattWestb:
Is you sure that  the router was not sending one rejoin to the end devicees / children (if it / they was awake and receiving it) ?

Adminiuga:
Need to read the specs. Is it applicable only with rejoin=True to tell the router to kick out it's children then rejoin?

PR #716, opened 2021-04-23
Update ZCL with changes from v7 spec

puddly:
Notable additions:

 - `enhanced` color operations
 - global `attr_reporting_status` (0xFFFE) attribute. Not sure if it's actually used yet but it can now appear in attribute reports.
 - `start_up_on_off` and `start_up_current_level` attributes, which control the power-on behavior of devices.
 - new optional parameters are now documented, which should also fix the various IKEA remote deserialization warnings
 - I've named a bunch of enums and bitmaps in various clusters

PR #905, opened 2022-02-07
Review cluster attribute names and types

mdeweerd:
I've reviewed cluster attribute names as per https://github.com/zigpy/zigpy/discussions/902, added missing attributes and ajusted some attribute types.

I found less typos that I think I saw when using HA/ZHA.


mdeweerd:
The following command definition I added to general.py failed the test, but I do not see immediately why.

```python
0x0008: ("move_to_closest_frequency", (t.uint16_t), False),
```

zigpy/zcl/clusters/closures.py
```diff
@@ -434,11 +434,12 @@ class WindowCovering(Cluster):
     attributes = {
         # Window Covering Information
         0x0000: ("window_covering_type", t.enum8),
-        0x0001: ("physical_close_limit_lift_cm", t.uint16_t),
-        0x0002: ("physical_close_limit_tilt_ddegree", t.uint16_t),
+        0x0001: ("physical_closed_limit_lift_cm", t.uint16_t),
+        0x0002: ("physical_closed_limit_tilt_ddegree", t.uint16_t),
```
Adminiuga, line 438: Why "ddegree" here and in a few other attributes next to it?

zigpy/zcl/clusters/general.py
```diff
@@ -448,14 +462,15 @@ class LevelControl(Cluster):
...
-        0x0000: ("move_to_level", (t.uint8_t, t.uint16_t), False),
+        0x0000: ("move_to_level", (t.uint8_t, t.uint16_t), False),  # + 2 optionals
         0x0001: ("move", (t.uint8_t, t.uint8_t), False),
         0x0002: ("step", (t.uint8_t, t.uint8_t, t.uint16_t), False),
         0x0003: ("stop", (), False),
         0x0004: ("move_to_level_with_on_off", (t.uint8_t, t.uint16_t), False),
         0x0005: ("move_with_on_off", (t.uint8_t, t.uint8_t), False),
         0x0006: ("step_with_on_off", (t.uint8_t, t.uint8_t, t.uint16_t), False),
         0x0007: ("stop", (), False),
+        # 0x0008: ("move_to_closest_frequency", (t.uint16_t), False),
```
Adminiuga, line 473: Do we need to add the commented out attribute?

zigpy/zcl/clusters/closures.py
mdeweerd: It was already there, and there is also "_cm".
"_cm" is for centimeters and I figured that "ddegree" is for decimal degrees as opposed to radians.

The ZCL spec does not add "Cm" nor "DDegree" in the attribute label.

Personally I would remove it as it's not in the original label, but as it was already there I left it in place.

If you all agree, I can remove those.

zigpy/zcl/clusters/general.py
mdeweerd: The commented line is a command, but when I added it the tests didn't pass and I did not understand why not.

So I commented it out and made [a note about it](https://github.com/zigpy/zigpy/pull/905#issuecomment-1030983815)

puddly: A one-tuple containing just `t.uint16_t` would be `(t.uint16_t,)`.

I've just merged https://github.com/zigpy/zigpy/pull/716, which rewrites just about every single ZCL command definition and likely fixes many of the problems you pointed out.

mdeweerd: Thank you for explaining the tuple subtlety.  

Ok, so this PR is useless and I can close it and remove the branch?  (just adding a last change to test the tuple fix ).

puddly:
Make sure to set up pre-commit in the Git repository:

```bash
$ pip install pre-commit
$ pre-commit install
```

That way, it will re-format your code when you make a Git commit and notify you of any problems.

zigpy/zcl/clusters/general.py
Adminiuga: that was a good catch. @mdeweerd are you running pre-commit hooks? i thought those were supposed to catch and correct this, or at least modify it.

mdeweerd: I have not set up pre-commit hooks, I have just set up "tox" that I run manually to debug cases that pop up.

Now I installed: 
- `pip install pre-commit`
- and, in the repository:  `pre-commit install`
- and, to test: `pre-commit run --all-files`

So that should avoid some commits with failing tests in the future.

zigpy/zcl/clusters/closures.py
puddly: They're called `PhysicalClosedLimit – Lift` and `PhysicalClosedLimit – Tilt` in the spec. The two subsequent attributes are named similarly.  I think `physical_closed_limit_lift` is fine, we don't provide units on any other attributes.

zigpy/types/basic.py
```diff
@@ -489,6 +489,33 @@ def deserialize(cls, data):
         return cls(s), data[cls._prefix_length + num_bytes :]
 
 
+def LimitedLVBytes(max_len):  # noqa: N802
```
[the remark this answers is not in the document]
dmulcahey: +1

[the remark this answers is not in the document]
mdeweerd: It crossed my mind, but it seems to met that at this moment there is only one case with a specific length and it's also extra work.
We could also say that Character strings are just a specific case of LVBytes.

I set out to review names, and ended up adjusting type names and even added type checking.
Maybe I'll be in a mood later to add this, but it shouldn't stall the PR.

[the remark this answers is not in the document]
puddly: No problem, we need to re-write all of the dynamic class generation "types" anyways.

zigpy/zcl/clusters/closures.py
mdeweerd: I removed the _cm and _ddegree post fixes on the name.  I agree that the base names are sufficiently different.

puddly:
Is this PR ready to merge?  The changes look good to me.

mdeweerd:
For me it is.

PR #909, opened 2022-02-09
Rename IasZone.ZoneType enum member

sergey-suloev:
According to specification rename IasZone.ZoneType.Carbon_Monoxide_Sensor to  IasZone.ZoneType.Gas_Sensor in   zcl/clusters/security.py

### DIFF
--- a/zigpy/zcl/clusters/security.py
+++ b/zigpy/zcl/clusters/security.py
@@ -26,7 +26,7 @@ class IasZone(Cluster):
         Contact_Switch = 0x0015
         Fire_Sensor = 0x0028
         Water_Sensor = 0x002A
-        Carbon_Monoxide_Sensor = 0x002B
+        Gas_Sensor = 0x002B
         Personal_Emergency_Device = 0x002C
         Vibration_Movement_Sensor = 0x002D
         Remote_Control = 0x010F


puddly:
Thanks for the PR! Which revision of the spec are you referencing? I'm not seeing "gas sensor" in ZCL R8:

image

sergey-suloev:
@puddly I think I incorrectly said "spec". I named it according to zstack documentation. For example, here https://e2e.ti.com/cfs-file/__key/communityserver-discussions-components-files/158/Z_2D00_Stack-Linux-Gateway-_2D00_-API-document-v1.1.pdf.

I am 100% sure that the value is not only intended for CO2 but for every gas.

MattWestb:
From https://docs.silabs.com/d/zigbee-af-api/7.0/enums#ember-af-ias-zone-type
`EMBER_ZCL_IAS_ZONE_TYPE_CARBON_MONOXIDE_SENSOR` that is the lastest EZSP 7.0 = ZCL R8.

puddly:
For consistency, the `zigpy` package tries to follow the ZCL and ZDO specs to the letter.

If a specific device uses the zone type attribute differently (or the spec is wrong...), you can create a `CustomDevice` subclass that renames or adds a synonym for the attribute (not recommended) or just change what's displayed in your end application if it's built on top of zigpy.

dmulcahey:
> It was possible changing the device class in HA but its taken away for 2 releases ago the the customizing was deleted.
> 
> I have one Aqara magnet (door / window sensor) that is in one fire detector and is triggered of the the LED is getting 9V (normal open collector out put) and its diverted by resistors to 3V and is then sending one open to ZHA and HA is making one smoke alarm off it.

You can still do this in HA. It’s in the entity settings UI although your options are now limited

Adminiuga:
Next time please provide the exact specification you are referencing too. There are a few things which don't make sense in the specs, yet i think it is better to follow the official spec than ZStack one. Less confusion down the road.

MattWestb:
@dmulcahey Its looks its only possible using the same class from the device and not changing it then its in ground on open / close sensor so cant being overrides to fire alarm with the GUI its only possible in customize.yaml
```
binary_sensor.openclose_23:
  device_class: smoke
```
LUMM01

dmulcahey:
Right it’s very limited. If you want to do something like that customize.yaml is the way to go.

PR #916, opened 2022-02-13
Add mypy, autoflake8, pyupgrade to pre-commit

mdeweerd:
i've added mypy to the flow and inhibited several notifications to those that seem most relevant.

They can be listed using:
```
 pre-commit run --hook-stage manual  mypy -a
```

Some highlights:
- `zigpy/state.py:20: error: Name "t.uin32_t" is not defined  [name-defined]`
- `zigpy/device.py:428: error: Name "model" already defined on line 413  [no-redef]`  (model defined two times, different implementations)
- `zigpy/quirks/__init__.py:192: error: Name "foundation.Command" is not defined  [name-defined]` - was renamed to foundation.GeneralCommand .


puddly:
Good catch! Completely annotating zigpy will definitely help https://github.com/zigpy/zha-websocket-server/.

Regarding the issues you noted:

1. The annotation isn't used by `dataclasses` as far as I can tell but it should be corrected.
2. The second definition is intentional, it's a property setter: https://docs.python.org/3/library/functions.html#property. I believe this is a mypy bug and can be "fixed" by moving the definition of the property setter directly below the definition of the property.
3. `foundation.Command` can still be accessed but it will generate a `DeprecationWarning`. However, this definitely shouldn't happen in zigpy itself. This should be fixed too.

puddly:
Could you also move the mypy options into [`setup.cfg`](https://github.com/zigpy/zigpy/blob/dev/setup.cfg)?  This would let you run `mypy` locally and use the same config as CI.

mdeweerd:
I moved the options to setup.cfg .  Mypy accepts it even if that feature does not seem to be mentioned in its documentation.

I fixed the examples I mentionned.  Moving the setters worked.

I haven't addec override to the exclusions nor no-redef .   I think the remaining notfications can either be "fixed" or skipped using an annotation like ` # type:ignore[no-redef]` .

puddly:
Thanks!  How do you feel about disabling as much of MyPy as necessary to just get CI passing? We can incrementally enable more options in future PRs, that way we don't have a perpetual PR that constantly needs to be rebased.

mdeweerd:
These issues remain.
You're more an expert in Python than I am, so I am awaiting your feedback.
 

These seem to be real errors, should we return an iter() ?  I added an ignore to see if the functionnal test pass in CI.
- zigpy/state.py: note: In member "__iter__" of class "CounterGroups":
zigpy/state.py:217: error: Return type "Iterable[CounterGroup]" of "__iter__" incompatible with return type "Iterator[Any]" in supertype "dict"  [override]
- zigpy/state.py:217: error: Return type "Iterable[CounterGroup]" of "__iter__" incompatible with return type "Iterator[Any]" in supertype "Iterable"  [override]

These are related to functools.partial.  I did not find a way to define them in a way mypy would accept them.  If you don ot know of a way, I propose that we just ignore those (I added the ignore tag to the source):
zigpy/quirks/__init__.py: note: In member "_configure_reporting" of class "CustomCluster":
zigpy/quirks/__init__.py:249: error: Signature of "_configure_reporting" incompatible with supertype "Cluster"  [override]
zigpy/quirks/__init__.py: note: In member "_read_attributes" of class "CustomCluster":
zigpy/quirks/__init__.py:268: error: Signature of "_read_attributes" incompatible with supertype "Cluster"  [override]
zigpy/quirks/__init__.py: note: In member "_write_attributes" of class "CustomCluster":
zigpy/quirks/__init__.py:282: error: Signature of "_write_attributes" incompatible with supertype "Cluster"  [override]
zigpy/quirks/__init__.py: note: In member "_write_attributes_undivided" of class "CustomCluster":
zigpy/quirks/__init__.py:298: error: Signature of "_write_attributes_undivided" incompatible with supertype "Cluster"  [override]

This is probably not important, so I ignored it, but just in case you think otherwise:
- tests/test_appdb_pysqlite.py:36: error: Name "sqlite3" already defined (by an import)  [no-redef]

puddly:
As long as unit tests pass, everything should work fine.

 - The `iter` stuff is probably correct but since we only ever just do `for elem in iterable`, it doesn't matter. Maybe try `return iter(self.values())` and change the return type annotation to `-> Iterator[CounterGroup]`?

 - The `partial` stuff I think is a real bug.  The quirks functions are all missing kwargs present in `Cluster.general_command`:

   ```Python
   expect_reply: bool = True,
   tries: int = 1,
   tsn: Optional[Union[int, t.uint8_t]] = None,
   **kwargs,
   ```

 - I'd omit `tests/` from mypy completely.  All of the `*` imports in `tests/test_appdb_pysqlite.py` are intentional, since the goal is to re-run the entire `tests/test_appdb.py` suite but with a different SQLite module.

zigpy/quirks/__init__.py
```diff
@@ -273,7 +273,7 @@ def _read_attributes(  # type:ignore[override]
         **kwargs,
     ):
         """Read attributes ZCL foundation command."""
-        if manufacturer is None and self._has_manuf_attr(args):
+        if manufacturer is None and self._has_manuf_attr(attribute_ids):
```
puddly, line 276: Good catch!  This should have been caught with a unit test...

puddly: You should be able to run them with `pytest -sv` after installing testing deps: `pip install -r requirements_test.txt`.

zigpy/zcl/clusters/__init__.py
```diff
@@ -15,8 +16,8 @@
 )
 from .. import Cluster
 
-CLUSTERS_BY_ID: dict[int, Cluster] = {}
-CLUSTERS_BY_NAME: dict[str, Cluster] = {}
+CLUSTERS_BY_ID: typing.Dict[int, Cluster] = {}
```
puddly, line 19: Add `from __future__ import annotations` at the very top of the file, it will allow built-in types like `dict` to be used in annotations on older versions of Python.

zigpy/device.py
```diff
@@ -79,7 +79,11 @@ def name(self) -> str:
 
     @property
     def non_zdo_endpoints(self) -> list[zigpy.endpoint.Endpoint]:
-        return [ep for epid, ep in self.endpoints.items() if epid != 0]
+        return [
+            ep
+            for epid, ep in self.endpoints.items()
+            if epid != 0 and not (isinstance(ep, zdo.ZDO))
```
Adminiuga, line 85: isn't this redundant? endpoint id is reserved for ZDO endpoints

zigpy/ota/provider.py
```diff
@@ -324,7 +324,13 @@ async def fetch_image(self) -> Optional[OTAImage]:
             with tarfile.open(fileobj=img_tgz) as tar:  # Unpack tar
                 for item in tar:
                     if item.name.endswith(".ota"):
-                        file_bytes = tar.extractfile(item).read()
```
puddly, line 327: I think this would work better (if you don't just want to ignore this error):

```Python
tar_f = tar.extractfile(item)
assert tar_f is not None
file_bytes = tar_f.read()
```

The conditions for `None` being returned are pretty specific and probably will never be encountered:

```python
# If there's no data associated with the member (directory, chrdev,
# blkdev, etc.), return None instead of a file object.
```

Adminiuga (commented):
Looks good

zigpy/device.py
mdeweerd: I think so, but I do not know all "secrets" ... so I left the original condition in place.
'mypy' does not understand that 'epid!=0' means its an Endpoint instance.

Ok if I remove the !=0 test?

zigpy/ota/provider.py
mdeweerd: I've seen a static code checker complain about assertions - they're not compiled in optimized code.
I agree it's likely never encountered - a side effect of static code checks, but I prefer fixing it in the code rather than adding an exception for the tool.

zigpy/zcl/clusters/__init__.py
mdeweerd: Good, I was wondering what the trick was to get 'dict' working.

zigpy/ota/provider.py
puddly: Sure. In that case, `raise ValueError(f"Issue ...` should be enough. It will automatically be logged as an exception here: https://github.com/zigpy/zigpy/blob/e91ac25d968403fd8e26c828e2177cfa2797973a/zigpy/util.py#L68

zigpy/quirks/__init__.py
mdeweerd: - The iter stuff: changed & tests pass.
- The partial stuff: I had already added *args, and **kwargs, but mypy still does not understand it.  Anyway, it should be functionnaly fixed.

- "I'd omit tests/ from mypy completely."  
   I would not.   There are not too many complaints from mypy in the test code, so fixing or ignoring them would help identify real issues in future code during development.

Quite a few mypy rules have been fixed.
You may look into enabling type-arg, and return-value - there's only a few cases left that you may fix or ignore.
The 'operator' rule only breaks in CI (zigpy.types.basic), but in my local setup.
The other rules generate a lot of occurences (close to 1000 each) that may for some need only a few code changes.

You'll need to comment them in setup.cfg .

zigpy/ota/provider.py
mdeweerd: Ok, changed.

zigpy/device.py
mdeweerd: I removed the epid!=0, tests pass.

mdeweerd:
I reviewed my changes as well and delete a local work file that got into version control.

I invite you to remove the last violations of the type-arg, return-value and operator rules of mypy.

mdeweerd:
I've added pyupgrade and autoflake8 as well.

mdeweerd:
I pulled the dev branch and to make the mypy tests 
pass I had make some annotations to the update AttributeReportingConfig class ad2f002f76dc6435b93147ad894f1ad94c1b2093 .

Issue #918, opened 2022-02-15
Improve AttributeReportingConfig (de)serialisation


mdeweerd:
I managed to make a Read Report Configuration request.

I got "Unsupported attributes" back, but zigpy did not understand the message:

2022-02-15 01:01:43 ERROR (MainThread) [zigpy.device] Failed to parse message (b'18530986000000') on cluster 1794, because Data is too short to contain 2 bytes

tshark's decoding of the request:
```
ZigBee Cluster Library Frame, Command: Read Reporting Configuration, Seq: 83
    Frame Control Field: Profile-wide (0x00)
        .... ..00 = Frame Type: Profile-wide (0x0)
        .... .0.. = Manufacturer Specific: False
        .... 0... = Direction: Client to Server
        ...0 .... = Disable Default Response: False
    Sequence Number: 83
    Command: Read Reporting Configuration (0x08)
    Attribute Status Record
        Direction: Reported (0x00)
        Attribute: Client: Functional Notification Flags / Server: Current Summation Delivered (0x0000)
```

tshark's decoding of the response:
```
ZigBee Cluster Library Frame, Command: Read Reporting Configuration Response, Seq: 83
    Frame Control Field: Profile-wide (0x18)
        .... ..00 = Frame Type: Profile-wide (0x0)
        .... .0.. = Manufacturer Specific: False
        .... 1... = Direction: Server to Client
        ...1 .... = Disable Default Response: True
    Sequence Number: 83
    Command: Read Reporting Configuration Response (0x09)
    Reporting Configuration Record
        Status: Unsupported Attribute (0x86)
        Direction: Reported (0x00)
        Attribute: Client: Functional Notification Flags / Server: Current Summation Delivered (0x0000)
```

A longer example (3 attributes):
2022-02-15 01:05:13 ERROR (MainThread) [zigpy.device] Failed to parse message (b'185b09860000008600120086000004') on cluster 1794, because Data is too short to contain 2 bytes

Request (as decoded by tshark):
```
ZigBee Cluster Library Frame, Command: Read Reporting Configuration, Seq: 91
    Frame Control Field: Profile-wide (0x00)
        .... ..00 = Frame Type: Profile-wide (0x0)
        .... .0.. = Manufacturer Specific: False
        .... 0... = Direction: Client to Server
        ...0 .... = Disable Default Response: False
    Sequence Number: 91
    Command: Read Reporting Configuration (0x08)
    Attribute Status Record
        Direction: Reported (0x00)
[...]

PR #920, opened 2022-02-16
#918 Test case for report configuration

mdeweerd:
Fix will be pushed next.

First pushing test case to validate it fails on zigpy with bug.


mdeweerd:
To many commits reported for this merge, canceling and redoing that from dev.

PR #921, opened 2022-02-16
#918 Test case & Fix for Read Reporting Configuration

mdeweerd:
Commit test case first to show it fails.


mdeweerd:
Fixes #918 .

mdeweerd:
@puddly This does not fix all the cases, only where there is exactly one record.  Each Read Reporting Configuration Attribute Record has a status byte and only dir and attribute id as mandatory fields.

How do you want to support this: https://github.com/zigpy/zigpy/blob/ae5d86e8f8e93213c7461f00f9f9e789858763f4/tests/test_device.py#L189-L199 ?

We can't do `t.List[t.List[Success, Union[UnsupportedAttribute, AttributeReportConfig]]` I think.

zigpy/zcl/foundation.py
```diff
@@ -322,9 +322,9 @@ def deserialize(cls, data):
         else:
             # Notifying that I will report things to you
             self.datatype, data = t.uint8_t.deserialize(data)
+            datatype = DATA_TYPES[self.datatype]
```
Adminiuga, line 325: Why moving the datatype away from the line were it is being used?

mdeweerd: I moved that line up during debug close to where self.datatype is assigned and had an instruction in place to log its value.
I initially examined the possibility that something was wrong with the datatype and as bytes are missing at some point, that exception occurred before I could look at datatype.

Do you know why datatype is named very much like self.datatype while they are not the same variable?  
Personnally I'ld avoid that.  It could be datatypePropse, typeProps, typeInfo, ... .

puddly:
I've submitted a PR to your repo with my implementation: https://github.com/mdeweerd/zigpy/pull/1

The `status` field dictating parsing but not being included in the attribute report config makes this an annoying problem if you want to avoid copy/pasting the entire class to add in a new field.  `t.Struct` doesn't support fields with types that change during parsing so all of that has to be implemented by hand.

PR #928, opened 2022-02-27
Allow disabling the automatic manufacturer ID override in quirks

### LINKED ISSUE #924: Ensure that manufacturer id can be forced to "None" from high level call
It appeared in https://github.com/zigpy/zha-device-handlers/discussions/1321#discussion-3849987 that even if the "manufacturer" was set to None in the high level read_attributes call, the Zigbee command actually sent did include the manufacturer code.

In this specific case that was because a quirk with a cluster id  with automatically decided that a manufacturer cluster was active.

I think that setting "manufacturer" to `b""` avoids that it is replaced in the current code and not sent in the request.

There may however be caveats in the current implementation, and a method should continue to exist in future implementations, the typing hints need to be updated as well (manufacturer can only be None or uint16 .

Another (additional) method would be to recognize 0 as No Manufacturer.

Wireshark indicates that Manufacturer Code "0" stands for No Manufacturerer.

In the Metering Cluster and for Tunneling it is said that "A value of 0xFFFF indicates that the Manufacturer Code is not used"  but in OTA it is said to be a "wild card".  So 0xFFFF does not seem suitable.

So I propose:
- To continue to accept None as a non-enforced manufacturer code that can be set in lower levels;
- To use b"", "" and 0 as codes to indicate "No Manufacturer" and ensure that this does not result in the manf bit being set;
- Alternatively define a constant (string) allowing to detect the "No manufacturer code" parameter.

puddly:
Fixes #924

### DIFF
--- a/tests/test_quirks.py
+++ b/tests/test_quirks.py
@@ -906,3 +906,40 @@ def test_quirk_deprecated_manufacturer_prefixes():
             manufacturer_server_commands = {
                 0x1234: ("foo3", {}, False),
             }
+
+
+async def test_manuf_id_disable(real_device):
+    class TestCluster(ManufacturerSpecificCluster):
+        cluster_id = 0xFF00
+
+    real_device.manufacturer_id_override = 0x1234
+
+    ep = real_device.endpoints[1]
+    ep.add_input_cluster(TestCluster.cluster_id, TestCluster(ep))
+    assert isinstance(ep.just_a_cluster, TestCluster)
+
+    assert ep.manufacturer_id == 0x1234
+
+    # Works normally
+    with patch.object(ep, "request", AsyncMock()) as request_mock:
+        request_mock.return_value = (zcl.foundation.Status.SUCCESS, "done")
+        await ep.just_a_cluster.command(
+            ep.just_a_cluster.commands_by_name["server_cmd0"].id,
+            manufacturer=None,
+        )
+
+    data = request_mock.mock_calls[0][1][2]
+    hdr, _ = zcl.foundation.ZCLHeader.deserialize(data)
+    assert hdr.manufacturer == 0x1234
+
+    # Can be disabled by passing NO_MANUFACTURER_ID
+    with patch.object(ep, "request", AsyncMock()) as request_mock:
+        request_mock.return_value = (zcl.foundation.Status.SUCCESS, "done")
+        await ep.just_a_cluster.command(
+            ep.just_a_cluster.commands_by_name["server_cmd0"].id,
+            manufacturer=zcl.foundation.ZCLHeader.NO_MANUFACTURER_ID,
+        )
+
+    data = request_mock.mock_calls[0][1][2]
+    hdr, _ = zcl.foundation.ZCLHeader.deserialize(data)
+    assert hdr.manufacturer is None
--- a/tests/test_quirks.py
+++ b/tests/test_quirks.py
@@ -920,7 +920,8 @@ async def test_manuf_id_disable(real_device):
 
     assert ep.manufacturer_id == 0x1234
 
-    # The default is to include the manufacturer ID
+    # The default behavior for a manufacturer-specific cluster, command, or attribute is
+    # to include the manufacturer ID in the request
     with patch.object(ep, "request", AsyncMock()) as request_mock:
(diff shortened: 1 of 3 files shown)


mdeweerd:
As far as DISABLE_AUTO_MANUFACTURER_ID, I would call it "NO_MANUFACTURER_ID" (similar to what we find in Wireshark btw) and define it in zcl or foundation, not specific to the quirk feature.

There are several approaches to implement this, and I was thinking a conversion/normalisation method would be cleaner, and only as close to where the manufacturer is actually encoded.

I would have left the quirks code as is, where "is None" would allow automatic encoding and resolve the manufacturer code where the actual encoding thereof is needed.


That would allow an extra feature as well.

Something like:
```
manf_code = NormalizeManufacturer(manufacturer)
```

NormalieManufacturer would then either return the bytes for the manufacturer: either empty or 2 bytes.
The following conversions apply:
- manufacturer is None -> b""
- manufacturer is 0 -> b""
- manufacturer is int in range 1..0xFFFF -> 2 bytes encoding
- manufacturer is 2 bytes encoding -> same value
- manufacturer is 1 byte encoding -> prefix byte 0x00.
- future: parameter manufacturer is string listed in manufacturer dictionnary -> 2 bytes encoding of the listed code
- manufacturer is invalid (out of range, number of bytes > 2, not found in dictionary) -> raise ValueError .

mdeweerd:
Ok, I suppose that using "-1" will work as well.

I know this is WIP, but the constant should not be defined in the quirks only - it has to work when no quirk is set for the device as well.

puddly:
> it has to work when no quirk is set for the device as well.

From what I see, only `quirks.py` ever modifies the provided `manufacturer`.  Everywhere else, the attribute is passed unmodified down into `ZCLHeader` so this problem only affects devices with quirks, no? 

To maintain API compatibility, `None` should disable the manufacturer ID and every other value should be passed through as-is. The quirks clusters are an exception because they set it by default but since the same code should work with and without a quirk, `None` is the only value that makes sense.

Does that sound right? Am I missing something?

mdeweerd:
When code uses the device/ep/cluster/... instance to send a command, for instance an attribute_write command, and it set the manufacturer code to -1 while the device does not hava a quirk.  What happens then?  The -1 is not changed and will be encoded to 0xFFFF I guess.

The caller should not know if the destination has a quirk or not - this can change over time.

Therefore, the -1 needs to be taken into account in the zcl/foundation layers so that -1 gets encoded to "No manufacturer" if this value makes it through all the layers.

puddly:
`-1` was meant only as an internal default parameter value for `quirks.py` to keep MyPy happy, since I couldn't figure out how to get typing working for a sentinel value like `_NOT_SET = object()`.  It's not meant to be used outside of zigpy internals and I will rename the global to reflect this.

Both with and without a quirk, I'm trying to make `command(..., manufacturer=None)` disable the manufacturer ID, `command(..., manufacturer=0x1234)` set it to `0x1234`, and `command(...)` (what 99% of code uses) do whatever the default is: either disable the manufacturer ID (when no quirk is loaded) or set it to the device manufacturer ID (with a quirk, in some cases).

mdeweerd:
- "command(...)" === Automatic manufacturer selection
- "command(...., manufacturer=None) means no manufacturer,
- ""command(...., manufacturer=<int>0>) means with manufacturer (I would still exclude the 0).

Regarding the requirement to change the quirks: that's why I was proposing "None" as the automatic value and b"" or 0 as the explicit "No manufacturer value" moving the change entirely to zcl/foundation rather than changing the quirk setup.

`_NOT_SET = object()` - I do not know the exact situation but if _NOT_SET is a proporty in a class, you can type it without setting it ( _NOT_SET: int ) or even provide a type hint in the comment when it's set to None upon initialisation ( `## type: int` ) .

mdeweerd:
I am more comfortable with the updated implementation and I hope that you are too.

A test case writing/reading an attribute that is labeled as the manufacturer attribute due to the quirk would be appropriate I think.
Testing both with "manufacturer=None" in the higher layer (resulting in the manufacturer being set) and "manufacturer=NO_MANUFACTURER_ID".

mdeweerd (approved):

Adminiuga (approved):
Lgtm

PR #933, opened 2022-03-13
Support devices with temporarily unknown IEEE or NWK addresses

Adminiuga:
Add support for devices with temporarily unknown IEEE or NWK addresses. If `get_device()` does not find the device, then create an "ephemeral" device, so the deserialize/serialize ZDO method would work to allow discovery of the missing address information via ZDO requests.

Previously, each radio lib had to handle "unknown" devices on its own, this allows handling unknown device discovery in zigpy. 

Tested with EZSP and deConz radios. For NWK address discovery, radio lib should support IEEE addressing mode.

### DIFF
--- a/tests/test_application.py
+++ b/tests/test_application.py
@@ -209,7 +209,7 @@ async def test_permit_ncp():
 
 async def test_permit(app, ieee):
     app.devices[ieee] = MagicMock()
-    app.devices[ieee].zdo.permit = MagicMock(side_effect=asyncio.coroutine(MagicMock()))
+    app.devices[ieee].zdo.permit = AsyncMock()
     app.permit_ncp = AsyncMock()
     await app.permit(node=(1, 1, 1, 1, 1, 1, 1, 1))
     assert app.devices[ieee].zdo.permit.call_count == 0
@@ -366,11 +366,6 @@ def test_get_device_ieee(app, ieee):
 def test_get_device_both(app, ieee):
     dev = app.add_device(ieee, 8)
     assert app.get_device(ieee=ieee, nwk=8) is dev
-
-
-def test_get_device_missing(app, ieee):
-    with pytest.raises(KeyError):
-        app.get_device(nwk=8)
 
 
 def test_ieee(app):
--- a/tests/test_application.py
+++ b/tests/test_application.py
@@ -708,3 +708,29 @@ async def test_get_device(app):
     app.add_device(t.EUI64.convert("11:11:11:11:22:22:22:33"), 0x0000)
 
     assert app.get_device(nwk=0x0000) is dev_2
+
+
+async def test_get_unknown_device_by_ieee(app):
+    """"Test getting unknown devices."""
+
+    await app.startup()
+
+    dev_unk_nwk = app.get_device(ieee=t.EUI64([0x01] * 8))
+    assert dev_unk_nwk.nwk == t.NWK.unknown()
+    dev_unk_nwk.nwk = t.NWK(0x1234)
+    dev_with_nwk = app.get_device(ieee=t.EUI64([0x01] * 8))
+    assert dev_with_nwk.nwk == t.NWK(0x1234)
+
+
+async def test_get_unknown_device_by_nwk(app):
+    """"Test getting unknown devices."""
+
+    await app.startup()
+
+    nwk = t.NWK(0x1122)
+
+    dev_unk_eui = app.get_device(nwk=nwk)
+    assert dev_unk_eui.ieee == t.EUI64([0x00] * 8)
+    dev_unk_eui.ieee = t.EUI64([0x11] * 8)
+    dev_with_eui = app.get_device(nwk=nwk)
+    assert dev_with_eui.ieee == t.EUI64([0x11] * 8)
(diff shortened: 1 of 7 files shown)


tests/test_application.py
```diff
@@ -708,3 +703,30 @@ async def test_get_device(app):
...
+async def test_get_unknown_device_by_ieee(app):
+    """"Test getting unknown devices."""
+
+    await app.startup()
+
+    dev_unk_nwk = app.get_device(ieee=t.EUI64([0x01] * 8))
+    assert dev_unk_nwk.nwk == t.NWK.unknown()
+    dev_unk_nwk.nwk = t.NWK(0x1234)
+    dev_with_nwk = app.get_device(ieee=t.EUI64([0x01] * 8))
+    assert dev_with_nwk.nwk == t.NWK(0x1234)
```
puddly, line 717: I think it would also be useful to check `assert dev_with_nwk is dev_unk_nwk`.

zigpy/application.py
```diff
@@ -462,6 +477,32 @@ def get_device(self, ieee=None, nwk=None):
...
+            new_ephemeral_dev.schedule_initialize()
+            return new_ephemeral_dev
+
+        # search by IEEE
+        for d in self._ephemeral_devices:
+            if d.ieee == ieee:
+                return d
+
+        new_ephemeral_dev = zigpy.device.Device(self, ieee, t.NWK.unknown())
+        self._ephemeral_devices.append(new_ephemeral_dev)
```
puddly, line 503: `new_ephemeral_dev.schedule_initialize()`?

zigpy/application.py
```diff
@@ -462,6 +477,32 @@ def get_device(self, ieee=None, nwk=None):
 
         raise KeyError
 
+    def get_ephemeral_device(self, ieee, nwk) -> zigpy.device.Device:
+        """Get ephemeral device with currently unknown ieee or nwk."""
+
+        clear_expired = [d for d in self._ephemeral_devices if not d.expired]
```
puddly, line 483: If an application holds a reference to an ephemeral device and the device "expires", this seems like it will cause unexpected behavior when zigpy creates a new clone.

What do you think about storing ephemeral devices in a [`weakref.WeakValueDictionary`](https://docs.python.org/3/library/weakref.html#weakref.WeakValueDictionary) (since there is no `WeakList`, maybe with just an incrementing key)?  That way, we can hold strong references to the devices in `_ephemeral_devices_refs` and expire them the way you're doing it here (or is `list.remove()` required?) but zigpy won't replace an ephemeral device with an identical clone when it's still in use.

puddly (commented):
This looks good!

The only potential issue I can think of is keeping active references to ephemeral devices from being prematurely expired. I think a `weakref` dict may be able to solve this problem.

zigpy/device.py
```diff
@@ -136,6 +153,39 @@ def schedule_initialize(self) -> asyncio.Task | None:
...
+            self.application, zdo.types.ZDOCmd.NWK_addr_req, 0, 0, self.ieee, 0x00, 0x00
+        )
+        status, ieee, nwk, *_ = await self.zdo.NWK_addr_req(
+            self.ieee, 0x00, 0x00, tries=2, delay=0.1, use_ieee=True
+        )
+
+        if status != zdo.types.Status.SUCCESS:
+            raise zigpy.exceptions.InvalidResponse(f"Requesting NWK failed: {status}")
+
+        self.nwk = nwk
```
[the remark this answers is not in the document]
Adminiuga: May need to check. The idea that this is part of the `Device.initialize()` method and once discover is complete it would be save into DB by `application.device_initialized()` method

zigpy/device.py
```diff
@@ -136,6 +153,39 @@ def schedule_initialize(self) -> asyncio.Task | None:
 
         return self._initialize_task
 
+    async def discover_ieee(self) -> EUI64:
+        self.info("Requesting IEEE address")
+
+        status, ieee, nwk, *_ = await self.zdo.IEEE_addr_req(
+            self.nwk, 0x00, 0x00, tries=2, delay=0.1
```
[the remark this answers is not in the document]
Adminiuga: the idea for these devices that only one address is unknown: either IEEE or NWK, but not both. And most of the times I'm expecting it  to be the IEEE address. If device is on the Network, then coordinator/Network most likely to know the network address and this would be triggered, when radio libarry fails to find a device in `get_device(nwk=received_frame_source_address)`

zigpy/application.py
Adminiuga: To trigger the discovery of the NWK addresses. NWK and IEEE discovery part of `Device.device_initialize()`

zigpy/application.py
Adminiuga: By application, do you mean ZHA or `ControllerApplication` ?  There could be trouble indeed if an ephemeral devices goes up all the way to ZHA. May be guard the origination of `device_initialized` event only for devices with fully discovered addresses? I haven't tested, but I would not expect it get there though, not until the device is fully initialized :thinking: 

Other than that, the `ControllerApplication` wouldn't store references to those devices anywhere, would it? except it could in topology ? https://github.com/zigpy/zigpy/blob/0c4b9a8a25c5eef5b008d66c2e56aae3f8d5e769/zigpy/neighbor.py#L31

Adminiuga: I need to play with the `WeakValueDictionary`, this sounds like a better approach here indeed.

zigpy/application.py
puddly: Sorry, I meant to say that `schedule_initialize` isn't called for the "unknown NWK" branch but is called for the "unknown IEEE". Is intentional?

PR #962, opened 2022-04-13
Persist attribute writes in the database

### LINKED ISSUE #958: Successful attribute writes are not written to the database
Attribute reports are cached in the database when they are received. Successful attribute writes should also be cached to ensure that the attribute cache doesn't get out of sync.

puddly:
Fixes #958


dmulcahey:
whoops didn't realize @puddly did this already

dmulcahey (approved):

PR #966, opened 2022-04-16
Store attributes in db after successful write

dmulcahey:
This PR adds attribute updated events after attribute writes which will cause attributes to be written to the database after writing their value successfully. This is needed to correctly support configuration entities for things such as StartupOnOff, StartupLevel, OnOffTransitionTime etc.


dmulcahey:
whoops didn't realize @puddly did this already

PR #969, opened 2022-04-20
Do not return the radio library status code when `expect_reply` is false

puddly:
Related to https://github.com/home-assistant/core/issues/69893


Adminiuga (approved):
Lgtm

PR #971, opened 2022-04-21
Persist `Device.last_seen` in the database

puddly:
`last_seen` will now be persisted in the database and loaded on startup.

I've also added a runtime check to `_migrate_tables` to ensure that all "old" tables are explicitly specified in the migration. This will protect against migrations that accidentally leave out a new table introduced in an earlier migration (like I almost did when making this PR 😃).


puddly:
Unit tests need to be fixed for 3.7, there seems to be some issue with `AsyncMock.__aenter__`.

PR #997, opened 2022-06-15
0.47.0 Release

puddly:
0.47.0 Release


pipiche38:
@puddly is the zigpy (dev) which has now the new-radio-api merged compatible with zigpy-znp, zigpy-deconz, bellows and zigpy-zigate ? I was under the impression that all radio libraries should be also aligned.

PR #1006, opened 2022-07-06
Automatic radio backups

puddly:
Backups are performed asynchronously on startup and then once every 24 hours (configurable).


pipiche38:
@puddly would it be possible to get it via an API ?
My question is related to the fact that we would be interested to integrate, but then we would store the backup into a JSON format into our own database and not the zigpy which is not used

puddly:
@pipiche38 There is a simple (provisional) API:

 - https://github.com/puddly/zigpy/blob/7ef57c4158e15d8cc279303b72618048e4a413c1/zigpy/backups.py#L33-L124
 - https://github.com/puddly/zigpy/blob/7ef57c4158e15d8cc279303b72618048e4a413c1/zigpy/application.py#L93-L95

If you choose to use the zigpy startup code instead of implementing your own, backups will be periodically broadcast with a `"network_backup_created"` event that can be consumed by your own database handler.

pipiche38:
Short questions, 
is the coordinator backup done in one single operation ? (just would like to know if we are preventing to corrupt something when restore the backup) ?
Is the backup stop all activities on the coordinator ? especially those touching to the trust center ?

puddly:
@pipiche38 Backups are completely asynchronous on all the coordinators I have tested. You shouldn't notice them happening.

pipiche38:
Thanks, my question was more on the risk to have a backup inconsistency if we have zigbee network changes during the backup time . My worries are more based on zigate which many tables than the other coordinators.
But if the backup is starting and you have a pairing in progress pit anything else touching key tables 

Envoyé de mon iPhone

PR #1008, opened 2022-07-11
Drop support for Python 3.7

puddly:
Python 3.7 will be end-of-life in 11 months and I don't know of any project depending on Zigpy's Python 3.7 support.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.7.7', 3.8, 3.9, '3.10']
+        python-version: [3.8, 3.9, '3.10']
     steps:
       - name: Check out code from GitHub
         uses: actions/checkout@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -315,7 +315,7 @@ jobs:
     needs: prepare-base
     strategy:
       matrix:
-        python-version: ['3.7.7', 3.8, 3.9, '3.10']
+        python-version: [3.8, 3.9, '3.10']
     name: >-
       Run tests Python ${{ matrix.python-version }}
     steps:
--- a/setup.py
+++ b/setup.py
@@ -20,6 +20,6 @@ setup(
     license="GPL-3.0",
     packages=find_packages(exclude=["tests", "tests.*"]),
     install_requires=REQUIRES,
-    python_requires=">=3.7",
+    python_requires=">=3.8",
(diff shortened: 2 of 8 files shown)


Adminiuga (approved):

pipiche38:
This is a very bad news. Why not waiting that it is not supported any more ?

Envoyé de mon iPhone

puddly:
What will change in 11 months? If you're still using 3.7, update now and save everyone the effort of having to maintain compatibility with both 3.7 and 3.8+

All new zigpy features need their unit tests rewritten multiple times to be compatible with both Python 3.7 and 3.8+. This is wasted effort. Zigpy should be compatible with Python 3.7 for the foreseeable future, this just will not be a tested configuration.

zigbeefordomoticz:
This is not about me, this is about users in production. Don’t forget this is linked to Domoticz which run an embedded python framework.

I’m know owning zippy and I’m just passing my concerns, dropping a compatibility 11 months simply before the end of the official support is quiet premature from my point of view. I understand the effort for tests, 
`just passing my feedback, and I found quiet strange that you started the PR "Python 3.7 will be end-of-life in 11 months and I don't know of any project depending on Zigpy's Python 3.7 support.” And just implement it!

Don’t forget that you are providing a libraries not an application

pipiche38:
I didn't want to be rude (I'm not native english) that might be the reason,  but on the other side the decision is also brutal.
I do understand your point, buut please acknowledge now that zigpy is not only use by HA

dumpfheimer:
Glad to hear it was not meant as I interpreted it. Apologies for calling it rude.

This is the way I see it:
The developers have decided that their time on supporting python 3.7 is no longer well spent. Which, in my opinion, is very reasonable. Python 3.7 will not receive security updates anymore in a few months. That is the absolute LATEST point in time where python 3.7 should not be used anymore, because it is potentially unsafe.

My second argument:
The "people in production" you mentioned do not have to turn off their zigbee network. They just might be "stuck" on the latest zigpy release that supports 3.7 for a few months (until they update their python installation). Without having data to back this up: I very strongly feel like "a few"(?) people having no zigpy updates for a few months does not outweigh overall progress on zigpy.

If python 3.7 is a thing in production for users I would rather spend time getting them to python 3.8 or higher than holding back development of zigpy (again, where development time is the most valuable resource)

Hope this makes sence

puddly:
Dropping 3.7 support was announced and discussed in January: https://github.com/zigbeefordomoticz/Domoticz-Zigbee/issues/1008

That being said, you're more than welcome to maintain Python 3.7 compatibility by submitting compatible pull requests to zigpy.

PR #1009, opened 2022-07-11
Remove Python 3.7 from CI

puddly:
Required for https://github.com/zigpy/zigpy/pull/1008 to pass CI


pipiche38:
Very sad news to stop the support of 3.7 before the en of life.


Envoyé de mon iPhone

PR #1014, opened 2022-07-20
Make the call to self.permit() at startup an option

pipiche38:
By default the self.permit(0) will call, but if the "disable_permit_to_join_at_startup" parameter is set to False this won't be the case any more


puddly:
Thanks. You can set up a development environment by using `pre-commit` within the zigpy repository. This will fix formatting issues automatically:

```
pip install pre-commit
pre-commit install .
pre-commit run --all-files  # to fix current problems
```

`CONF_DISABLE_PERMIT_TO_JOIN_AT_STARTUP` makes it sound like joins are permitted on startup. Can you think of a different name? Perhaps call it `CONF_DISABLE_JOINS_ON_STARTUP` and set the default to `True`?

pipiche38:
Ok, this seems too complex for me.

PR #1021, opened 2022-07-31
Set the request header's `direction` bit according to the cluster type

### LINKED ISSUE #1018: General commands on client cluster
Hi,

I need to  discover and read attributes on device's client cluster.
I manage to send general command to a server cluster and even to send general command to a client cluster with `cluster.reply(True, 0x0C, foundation.GENERAL_COMMANDS[0x0C].schema, 0, 200, manufacturer=manufacturer_code)`. However, I don't have any response on zigpy even if the device does respond (verified with a capture).
Is there an "easy" way to perform those general commands on client clusters ?

Thanks!

puddly:
Fixes #1018

I'm a little wary of merging this change without more runtime testing, especially with quirks. Are there any known devices that rely on the current behavior to work properly?

### DIFF
--- a/tests/test_zcl.py
+++ b/tests/test_zcl.py
@@ -957,3 +957,26 @@ def test_zcl_response_type_tuple_like():
     assert req == (0, 1, 2)
     assert req == req
     assert req == req.replace()
+
+
+async def test_zcl_request_direction():
+    """Test that the request header's `direction` field is properly set."""
+    dev = MagicMock()
+
+    ep = zigpy.endpoint.Endpoint(dev, 1)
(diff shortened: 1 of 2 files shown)


Adminiuga:
I think this particular change should be fine. 

In hindsight, `is_reply` name was poorly chosen, as it is really a direction: client -> server or server -> client.

MattWestb:
One little different problem with in and out cluster that i having problem with (tuya devices).
tuya have messing up in and out cluster on 3 endpoints of 4 and need the default response sent Client to Server and ZHA is sending it Server to Client.
Discussion https://github.com/zigpy/zha-device-handlers/issues/1448.


Is it possible fixing that  2 in some similar way ?

PR #1022, opened 2022-08-01
Replace `is_reply` with `direction` in ZCL headers

puddly:
This better matches the spec and also brings all of the ZCL header bitfield creation into a single function, making it a little easier to reason about:

https://github.com/puddly/zigpy/blob/8d620e97caf4375472bf1f76c9ed20081d886909/zigpy/zcl/__init__.py#L276-L286

Can we improve the way we set the `direction` and `disable_default_response` bits?

### DIFF
--- a/tests/test_device.py
+++ b/tests/test_device.py
@@ -10,6 +10,7 @@ import zigpy.exceptions
 from zigpy.profiles import zha
 import zigpy.state
 import zigpy.types as t
+import zigpy.zcl.foundation as foundation
 from zigpy.zdo import types as zdo_t
 
 from .async_mock import ANY, AsyncMock, MagicMock, int_sentinel, patch, sentinel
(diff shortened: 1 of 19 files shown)


puddly:
#1021 causes Tuya quirk unit tests to fail. Here is a reversion of it with this slightly more verbose format:

```diff
diff --git a/zigpy/zcl/__init__.py b/zigpy/zcl/__init__.py
index 6b57156..24840cc 100644
--- a/zigpy/zcl/__init__.py
+++ b/zigpy/zcl/__init__.py
@@ -315,11 +315,7 @@ class Cluster(util.ListenableMixin, util.CatchingTaskMixin):
                 manufacturer=manufacturer,
                 tsn=tsn,
                 disable_default_response=self.is_client,
-                direction=(
-                    foundation.Direction.Server_to_Client
-                    if self.is_client
-                    else foundation.Direction.Client_to_Server
-                ),
+                direction=foundation.Direction.Server_to_Client,
                 **kwargs,
             )
         except (ValueError, TypeError) as e:
```

Was Tuya not validating the `direction` bit beforehand? Or is #1021 going to break Tuya?

PR #1023, opened 2022-08-02
Automatically restore the most recent network backup instead of forming a new network

puddly:
If an adapter breaks and loses network information (or is re-flashed), restore the most recent network backup instead of forming a brand new network.

This seems like a more sensible default than randomly generating settings.

### DIFF
--- a/tests/test_application.py
+++ b/tests/test_application.py
@@ -14,7 +14,7 @@ import zigpy.state as app_state
 import zigpy.types as t
 import zigpy.zdo.types as zdo_t
 
-from .async_mock import AsyncMock, MagicMock, patch
+from .async_mock import AsyncMock, MagicMock, patch, sentinel
 from .conftest import App
 
 NCP_IEEE = t.EUI64.convert("aa:11:22:bb:33:44:be:ef")
@@ -607,6 +607,9 @@ async def test_startup_not_formed(app):
     )
     app.permit = AsyncMock()
 
+    app.backups.backups = []
+    app.backups.restore_backup = AsyncMock()
+
     with pytest.raises(NetworkNotFormed):
         await app.startup(auto_form=False)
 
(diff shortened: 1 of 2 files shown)


pipiche38:
I'm not sure that is a good idea because

- I think the restore should be a conscient action from the user. May be via question , where the end user is queried about the fact that we will restore a particular backup
- If you run several coordinator on the same application (several instance zigpy), then what is going to happen if you add a second instance when the first one has already started to do some automatic backup ?
-

puddly:
> If you run several coordinator on the same application (several instance zigpy)

Sharing a database between multiple zigpy instances isn't a supported configuration.

> What about restore on Elelabs ,where I think we can only overwrite the MAC@ once

It won't be overridden unless that's explicitly done by the user. It'll just log a warning. The current behavior is to form a brand new network, which is worse.

> Last I think the restore should be a conscient action from the user. 

Of course. All of this logic lives in `startup` so feel free to override it within your application by passing `auto_form=False` and then handling the `NetworkNotFormed` error.

pipiche38:
Ok sound good to me. 
It confirms that we have to run our on startup and not rely on zigpy and we have to run our own database

Hedda:
So this will make zigpy applications work similar to Zigbee2MQTT automatic recovery as well as migration? See thisZ2M FAQ:

https://www.zigbee2mqtt.io/guide/faq/#how-do-i-migrate-from-one-adapter-to-another

Also see these other related Z2M FAQs:

https://www.zigbee2mqtt.io/guide/faq/#what-does-and-does-not-require-repairing-of-all-devices

And

https://www.zigbee2mqtt.io/guide/faq/#how-do-i-move-my-zigbee2mqtt-instance-to-a-different-environment

puddly:
More or less. A few differences:

1. It won't overwrite any network settings if they're present on the new stick. This needs user feedback, which will be done with ZHA.

2. It supports all adapters, so you can unplug a Conbee and plug in a freshly flashed CC2652.

Right now ZHA has no way to reconfigure the serial port to change it. This currently restricts the usefulness of this change to just automatic failure recovery (i.e. Conbee lost its network settings), and restoring settings after flashing new firmware.

pipiche38:
@puddly, you have made zigate backup/restore as well ? that is a very good news

puddly:
ZiGate is not supported. I could find no way to do the backup over serial during operation, nor a way to dissect the binary NVRAM format.

pipiche38:
I understand and NXP is really not open minded ...

PR #1029, opened 2022-08-16
Move the core initialization logic out of `startup`, into a new method

puddly:
This allows for radio libraries to re-connect and re-initialize zigpy without having `disconnect` called during startup, breaking their reconnect loop.

@pipiche38 since you override `startup` in your `ControllerApplication` subclass, make sure to adjust your code accordingly.

https://github.com/home-assistant/core/issues/75292

### DIFF
--- a/zigpy/application.py
+++ b/zigpy/application.py
@@ -102,9 +102,8 @@ class ControllerApplication(zigpy.util.ListenableMixin, abc.ABC):
         Starts a network, optionally forming one with random settings if necessary.
         """
 
-        await self.connect()
-
         try:
+            await self.connect()
             await self.initialize(auto_form=auto_form)
         except Exception as e:
             LOGGER.error("Couldn't start application", exc_info=e)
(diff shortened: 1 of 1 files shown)


pipiche38:
@puddly, first thanks a lot for tagging me, really appreciated. Then I also understood that you have remove the permit(0) which them will allow us to use directly the ControllerApplication subclass without overriding (as it was the only reason).

Thanks for all the work you do.

puddly:
> Then I also understood that you have remove the permit(0) 

Not quite. The method was renamed, that's more or less it.

pipiche38:
ok that is bad, and a wrong understanding from  https://github.com/zigpy/zigpy-znp/pull/162

puddly:
No difference, just copy the new code. Or finish your PR to make it configurable.

pipiche38:
Ok understood. As the PR is concerned, not possible for me, too much things asked/expected that I don't know. better to drop

PR #1034, opened 2022-08-23
Reduce DB filesystem writes

puddly:
- switches the journal to WAL mode (big change)
 - optimizes a few common queries to not perform writes unnecessarily (minor change)

Should address https://github.com/home-assistant/core/issues/77049

I've also reduced the frequency with which `last_seen` timestamps are updated per-device to once every 30s. In theory this can cause `last_seen` to roll backwards if the value is cached somewhere and then zigpy is restarted.

### DIFF
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -6,5 +6,6 @@ pytest
 pytest-asyncio
 pytest-cov
 pytest-timeout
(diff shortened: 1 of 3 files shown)


dmulcahey (approved):
LGTM

PR #1037, opened 2022-08-29
Compare network state to last backup on startup

puddly:
This will prevent zigpy from silently allowing adapter swapping, specifically for devices that have non-unique serial ports (i.e. HA Yellow).  I've also added a config option to disable this feature.

https://github.com/home-assistant/core/pull/77044


puddly:
Looks like I've made some false assumptions about network frame counters: EmberZNet can roll back the frame counter by a few thousand after resetting. It seems like the runtime value is persisted only periodically.

With old Conbees existing that do not have a value for the frame counter and it being possible for users to migrate back and forth between the Conbee and other coordinators, I'm unfortunately going to have to relax the comparison to ignore the frame counter.

Instead, the current network state will always be saved, regardless of counter rollbacks or not. Only compatible past backups whose frame counters are lesser will be removed.

PR #1043, opened 2022-09-08
[RFC] Use a single structure and mirrored functions for transmitting/receiving Zigbee packets

puddly:
This change removes all of the specialized `ControllerApplication` methods responsible for sending requests (`request` for unicast, `mrequest` for group, and `broadcast` for broadcast) and replaces them with a single `send_packet` function. Similarly, `handle_message` is replaced with `receive_packet`.  Both of these functions pass around [`ZigbeePacket` objects](https://github.com/puddly/zigpy/blob/b3647dc487620560887071fb6f7293449a53cea4/zigpy/types/named.py#L551-L580) with addressing information instead of exposing `Device` objects to the radio library.

The aim of this change is remove functionality from individual radio libraries that instead can be handled by zigpy. Current plans:

 - Implement ZNP's `get_or_discover_device`, which sends a ZDO IEEE address request to discover a device if it does not exist in the database.
 - Possibly implement automatic retrying using the IEEE address to deal with devices whose NWK address has changed.
 - Support APS encryption (unlikely but it will be possible).
 - Allow sending packets directly to NWK and IEEE addresses, without an associated `Device` object.
   - To do this, I'm planning on replacing the hierarchical `Device`/`Endpoint`/`Cluster` `serialize()` and `deserialize()` methods with objects for ZCL and ZDO packets that will be passed as part of the `ZigbeePacket` object. They will only be serialized by the radio library when the request is actually about to be sent, which allows for easy modification of any layer of a packet by quirks, without having convert to/from bytes.

I'm currently testing with a patched zigpy-znp and everything works. This PR will likely change significantly as I patch the various radio libraries.


pipiche38:
This looks quiet interessting

puddly:
> shall we expect less adherence to the zigpy database for receiving and sending packet ?

Yes. The radio libraries will still attempt to look for `Device` objects when they need more information (i.e. zigpy-znp retrying using IEEE address and `relays`) but none will rely on it. You could just send/receive `ZigbeePacket` objects directly and never use any of zigpy's `Device`, `Endpoint`, or `Cluster` classes:

```Python
2022-09-09 15:55:57.028 DEBUG (MainThread) [zigpy.application] Received a packet: ZigbeePacket(src=AddrModeAddress(addr_mode=<AddrMode.NWK: 2>, address=0x1f72), src_ep=1, dst=AddrModeAddress(addr_mode=<AddrMode.NWK: 2>, address=0x0000), dst_ep=1, source_route=None, extended_timeout=True, tsn=50, profile=260, cluster_id=25, data=b'\x01\x03\x01\x01|\x11\x01!1\x86\x06#\x01\x00', tx_options=<TransmitOptions.NONE: 0>, radius=0, non_member_radius=0, lqi=255, rssi=-77)

...

2022-09-09 16:02:43.667 DEBUG (MainThread) [bellows.zigbee.application] Sending packet ZigbeePacket(src=None, src_ep=1, dst=AddrModeAddress(addr_mode=<AddrMode.NWK: 2>, address=0x1075), dst_ep=1, source_route=None, extended_timeout=True, tsn=17, profile=260, cluster_id=768, data=b'\x00\x11\x00\x00@\x01\x00\x07\x00', tx_options=<TransmitOptions.ACK: 1>, radius=0, non_member_radius=0, lqi=None, rssi=None)
```

However, this would remove your ability to use the somewhat-consistent ZDO interface for changing channels and energy scans. Perhaps it would be better to move it into a dedicated API as opposed to hacking zigpy's `ZDO` object.

pipiche38:
https://github.com/zigpy/zigpy/pull/1043/commits/b59858f7fa0ea6475b64356d200f3840084afb71 do you mean that we will still be able to use the old fashion ? with request() and handle_message() ?

puddly:
Yes, I'm hoping to make this change backwards compatible if you're using the public zigpy API.

pipiche38:
As source_routing is concerned, I wonder how does it works on ZNP, as I understood that is native with @koenkk firmware?

puddly:
> I wonder how does it works on ZNP

Z-Stack does its own routing but you can override it. This change is mainly for the Conbee.

The current implementation is a stub, as it just sends packets backwards over the last-seen route coming from the device.  In the future I'm hoping to implement something more interesting using the data from the topology scanner.

zigpy/device.py
```diff
@@ -284,11 +286,15 @@ async def request(
         timeout=APS_REPLY_TIMEOUT,
         use_ieee=False,
     ):
+        extended_timeout = False
+
         if expect_reply and (self.node_desc is None or self.node_desc.is_end_device):
             self.debug("Extending timeout for 0x%02x request", sequence)
             timeout = APS_REPLY_TIMEOUT_EXTENDED
+            extended_timeout = True
```
dmulcahey, line 294: should we change this to look at the rx on when idle stuff?

PR #1048, opened 2022-09-16
Add additional endpoint configuration option

prairiesnpr:
This PR adds a configuration option to allow adding additional endpoints.

This is required for xbee end devices as they require 0xE6 and 0xE8 endpoints and by default zigpy ignores those.

Example configuration.

```
zha:
  zigpy_config:
    additional_endpoints:
      - endpoint: 0xE6
        profile: 0xC105
        device_type: 0x0000
        device_version: 0b0000
        input_clusters: []
        output_clusters: [0x21]
      - endpoint: 0xE8
        profile: 0xC105
        device_type: 0x0000
        device_version: 0b0000
        input_clusters: []
        output_clusters: [0x21]
```

### DIFF
--- a/zigpy/application.py
+++ b/zigpy/application.py
@@ -23,6 +23,7 @@ import zigpy.util
 import zigpy.zcl
 import zigpy.zdo
 import zigpy.zdo.types as zdo_types
+from zigpy.config import CONF_ADDITIONAL_ENDPOINTS
 
 DEFAULT_ENDPOINT_ID = 1
 LOGGER = logging.getLogger(__name__)
@@ -538,6 +539,9 @@ class ControllerApplication(zigpy.util.ListenableMixin, abc.ABC):
                 output_clusters=[],
             )
         )
+
+        for add_endpoint in self.config.get(CONF_ADDITIONAL_ENDPOINTS):
+            await self.add_endpoint(zdo_types.SimpleDescriptor(**add_endpoint))
 
     async def mrequest(
         self,
--- a/zigpy/application.py
+++ b/zigpy/application.py
@@ -10,6 +10,7 @@ from typing import Any
 import zigpy.appdb
 import zigpy.backups
 import zigpy.config as conf
+from zigpy.config import CONF_ADDITIONAL_ENDPOINTS
 import zigpy.device
 import zigpy.exceptions
 import zigpy.group
@@ -23,7 +24,6 @@ import zigpy.util
 import zigpy.zcl
 import zigpy.zdo
 import zigpy.zdo.types as zdo_types
-from zigpy.config import CONF_ADDITIONAL_ENDPOINTS
 
 DEFAULT_ENDPOINT_ID = 1
 LOGGER = logging.getLogger(__name__)
--- a/zigpy/application.py
+++ b/zigpy/application.py
@@ -10,7 +10,6 @@ from typing import Any
 import zigpy.appdb
 import zigpy.backups
 import zigpy.config as conf
-from zigpy.config import CONF_ADDITIONAL_ENDPOINTS
 import zigpy.device
 import zigpy.exceptions
 import zigpy.group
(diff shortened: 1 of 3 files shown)


zigpy/config/__init__.py
```diff
@@ -124,3 +126,7 @@
 CONFIG_SCHEMA = ZIGPY_SCHEMA.extend(
     {vol.Required(CONF_DEVICE): SCHEMA_DEVICE}, extra=vol.ALLOW_EXTRA
 )
+
+ENDPOINT_SCHEMA = vol.Schema(
+    {vol.Optional(CONF_ADDITIONAL_ENDPOINTS): SimpleDescriptor}
+)
```
puddly, line 132: This schema isn't actually utilized. Instead of parsing the descriptor objects within `application.py`, move the parsing in here:

```python
        vol.Optional(
            CONF_NWK_VALIDATE_SETTINGS, default=CONF_NWK_VALIDATE_SETTINGS_DEFAULT
        ): cv_boolean,
        vol.Optional(CONF_ADDITIONAL_ENDPOINTS, default=[]): [
            lambda v: SimpleDescriptor(**v)
        ],

...
```

That way, `self.config[conf.CONF_ADDITIONAL_ENDPOINTS]` will always either be an empty list, or a list of `SimpleDescriptor` objects. The application will never have to parse them and you will receive a validation error if the config is invalid, before zigpy has a chance to start up.

zigpy/application.py
```diff
@@ -539,6 +540,9 @@ async def register_endpoints(self):
             )
         )
 
+        for add_endpoint in self.config.get(CONF_ADDITIONAL_ENDPOINTS):
```
puddly, line 543: Once the endpoint dictionary is parsed in `zigpy/config/__init__.py`, this code can just become:

```Python
for endpoint in self.config[conf.CONF_ADDITIONAL_ENDPOINTS]:
    await self.add_endpoint(endpoint)
```

zigpy/application.py
```diff
@@ -10,6 +10,7 @@
 import zigpy.appdb
 import zigpy.backups
 import zigpy.config as conf
+from zigpy.config import CONF_ADDITIONAL_ENDPOINTS
```
puddly, line 13: `zigpy.config` is imported as `conf` on the line above, you don't need to add another one.

puddly (changes requested):
Thanks for the PR! I've added a few comments regarding the config schema parsing.

Hedda:
@pipiche38 could this also be related to https://github.com/zigbeefordomoticz/Domoticz-Zigbee/issues/1145 or is that issue only on the coordinator side?

@prairiesnpr, for reference, exactly which Zigbee USB radio adapter are you using to test and exactly which firmware does it have?

pipiche38:
> @pipiche38 could this also be related to [zigbeefordomoticz/Domoticz-Zigbee#1145](https://github.com/zigbeefordomoticz/Domoticz-Zigbee/issues/1145) or is that issue only on the coordinator side?
> 
> @prairiesnpr, for reference, exactly which Zigbee USB radio adapter are you using to test and exactly which firmware does it have?

My understanding is that it should be done at the radio lib level ( ZNP, deConz or EZSP) because every coordinator as it is way to handle it.
For exemple, ZNP and deConz we had to add a number of Ep to get our Wiser , Livolo devices working, while on EZSP it looks like it is using an other mecanism and we didn't have to do it.
So the PR made at zigpy level, might be against the EZSP part

prairiesnpr:
> @pipiche38 could this also be related to [zigbeefordomoticz/Domoticz-Zigbee#1145](https://github.com/zigbeefordomoticz/Domoticz-Zigbee/issues/1145) or is that issue only on the coordinator side?
> 
> @prairiesnpr, for reference, exactly which Zigbee USB radio adapter are you using to test and exactly which firmware does it have?

It's one of Tube's Poe models.  cc2652p2, firmware 20220219.  Will also need https://github.com/zigpy/zigpy-znp/pull/175 for this to work.

puddly (approved):

PR #1068, opened 2022-10-07
Perform complete backups periodically

puddly:
Currently only a fast backup is performed periodically. This doesn't include APS keys or NWK address tables.


pipiche38:
So does that means a full backup should be done for proper restore ?

puddly:
Nobody has complained about backup/restore/migration not working thus far in ZHA so I'm not entirely sure if it's necessary.

pipiche38:
Ok. So the principle I have taken is to do full backup when shutdown, and then the fast backup during operation

puddly:
I've found no performance degradation doing continuous, complete backups with ZNP (i.e. zero second delay) so I don't expect it to even be noticeable once every 24 hours. Perhaps with coordinators that have unstable serial connections.

PR #1085, opened 2022-10-28
Only enable APS ACKs for packets without implicit delivery notification

puddly:
I think the current logic may be inverted. Instead, how about:

 - If a packet will be receiving a reply, we know when it was not received because no response was sent by the recipient.
 - If a packet will *not* be receiving a reply, we will have no way to know when it sends successfully, so we should enable APS ACKs to receive a low-level response from the device.

Thoughts?

Probably will fix https://github.com/home-assistant/core/issues/81057, since the Conbee struggles with APS ACKs it seems.

### DIFF
--- a/tests/test_application.py
+++ b/tests/test_application.py
@@ -807,7 +807,7 @@ def packet(app, device):
         data=t.SerializableBytes(b"test data"),
         source_route=None,
         extended_timeout=False,
-        tx_options=t.TransmitOptions.ACK,
+        tx_options=t.TransmitOptions.NONE,
     )
 
 
@@ -861,6 +861,14 @@ async def test_request(app, device, packet):
     app.build_source_route_to.assert_called_once_with(dest=device)
     app.send_packet.assert_called_once_with(
         packet=packet.replace(source_route=[0x000A, 0x000B])
+    )
+    app.send_packet.reset_mock()
+
+    # Test sending without waiting for a reply
+    status, msg = await send_request(app, expect_reply=False)
+
+    app.send_packet.assert_called_once_with(
+        packet=packet.replace(tx_options=t.TransmitOptions.ACK)
     )
     app.send_packet.reset_mock()
 
--- a/zigpy/application.py
+++ b/zigpy/application.py
@@ -651,6 +651,11 @@ class ControllerApplication(zigpy.util.ListenableMixin, abc.ABC):
         else:
             source_route = None
 
+        tx_options = t.TransmitOptions.NONE
+
+        if not expect_reply:
+            tx_options |= t.TransmitOptions.ACK
+
         await self.send_packet(
             t.ZigbeePacket(
                 src=src,
@@ -663,9 +668,7 @@ class ControllerApplication(zigpy.util.ListenableMixin, abc.ABC):
                 data=t.SerializableBytes(data),
                 extended_timeout=extended_timeout,
                 source_route=source_route,
-                tx_options=(
-                    t.TransmitOptions.ACK if expect_reply else t.TransmitOptions.NONE
-                ),
+                tx_options=tx_options,
             )
         )
 


MattWestb:
Testing and sniffing how its working in real you see how its working or not?
Do you knowing commands or devices that is having problem if sniffing ?

puddly:
I've unfortunately not been able to replicate this problem with my Conbee. My network behaves identically, regardless of whether or not APS ACKs are enabled.

From the above issue though, a user specifically mentions:

> This appears to only affect some of my Philips Hue Bulbs (LWB015). Where they will not respond to any commands.

Perhaps they do not support APS ACKs?

MattWestb:
If its the first gen its have the same chip ad Rap/CornBee I is having and have many strange thing implanted in the firmware like not deleting jumping / leaving children.

I only have one HUE light and its in the box of bad Zigbee devices but i dont knowing if its having the same undocumented futures.

My RaspBee I is broken (was not liking my new NanoPi Neo pin headers and got in fire) so cant testing that combination.

Make one test build and letting the user testing it if its possible.

dumpfheimer:
I have multiple Hue 1st gen lights, later gen lights and lightstrips, a whole bunch of Tradfri lights, quite a few Xiaomi devices and a few random devices like Tuya and gladopto.

All seem to be responsive after updating to this PR (has been only a few minutes now). Can I test anything in specific?

MattWestb:
Sevus @dumpfheimer and great thanks for testing then the devs dont have all hardware for testing and some armature (like my) is braking there hardware so cant helping.

Hopefully HUE first gen devices is working OK with deCONZ coordinators now.

Great work @puddly !!!

zigbeefordomoticz:
I'm not sure I understand this.
It looks like you are forcing Ack on all requests
But how can I send a request without Ack ?

puddly:
Use the packet API (`send_packet` and `packet_received`) if you want more control.

pipiche38:
Sure I can use send_packet and packet_received(). But my question is more on the `expect_reply` attribute, as what ever its value you'll always set it to ACK.
I'm trying to understand the logic from a zigpy standpoint , like you have as part of the Zigbee norm some request to be send without ACK

dmulcahey:
They aren’t… not sure why you keep saying they are always on. I’ve attempted to walk you through the 2 lines of code and you have no evidence of them being on other that stating it over and over. Get a packet capture if you want to continue this discussion

pipiche38:
@puddly, @dmulcahey 

I did further investigations as you recommended and instrument the zigpy code by adding 2 debug lines.

- 1 when receiving the request() , 1 prior sending send_packet


```
2022-11-03 20:43:42,630 DEBUG   :receive request dst: <Device model=None manuf=None nwk=0x4E23 ieee=00:15:bc:00:2f:00:cc:66 is_initialized=False> dst_ep: 2 expect_reply: False use_ieee: False extended_timeout: False
2022-11-03 20:43:42,634 DEBUG   :ready to send packet dst: AddrModeAddress(addr_mode=<AddrMode.NWK: 2>, address=0x4E23) dst_ep: 2 extended_timeout: False source_route: [] tx_options: TransmitOptions.ACK
```

From that log, when the expect_reply is not set, then TransmitOption get the ACK added
Confirmed from the sniff

Screenshot 2022-11-03 at 20 51 21

- Then 1 send the same command with expected_reply set to True

```
2022-11-03 20:46:49,609 DEBUG   :receive request dst: <Device model=None manuf=None nwk=0x4E23 ieee=00:15:bc:00:2f:00:cc:66 is_initialized=False> dst_ep: 2 expect_reply: True use_ieee: False extended_timeout: False
2022-11-03 20:46:49,612 DEBUG   :ready to send packet dst: AddrModeAddress(addr_mode=<AddrMode.NWK: 2>, address=0x4E23) dst_ep: 2 extended_timeout: False source_route: [] tx_options: TransmitOptions.NONE
```

Result is that the TransmitOptions is not set to Ack. 
Confirmed with the sniff
Screenshot 2022-11-03 at 20 50 44


So it looks like the expect_reply behaviour is inverses from before, where when expect_reply is set to True then teh Ack is set and vice-versa.


the code is here

```
        LOGGER.debug(
            "receive request dst: %s dst_ep: %s expect_reply: %s use_ieee: %s extended_timeout: %s", device, dst_ep, expect_reply, use_ieee, extended_timeout
        )

        if use_ieee:
            src = t.AddrModeAddress(
                addr_mode=t.AddrMode.IEEE, address=self.state.node_info.ieee
            )
            dst = t.AddrModeAddress(addr_mode=t.AddrMode.IEEE, address=device.ieee)
        else:
            src = t.AddrModeAddress(
                addr_mode=t.AddrMode.NWK, address=self.state.node_info.nwk
            )
            dst = t.AddrModeAddress(addr_mode=t.AddrMode.NWK, address=device.nwk)

        if self.config[conf.CONF_SOURCE_ROUTING]:
            source_route = self.build_source_route_to(dest=device)
        else:
            source_route = None

        tx_options = t.TransmitOptions.NONE

        if not expect_reply:
            tx_options |= t.TransmitOptions.ACK

        LOGGER.debug(
            "ready to send packet dst: %s dst_ep: %s extended_timeout: %s source_route: %s tx_options: %s", dst, dst_ep, extended_timeout, source_route, tx_options
        )

        await self.send_packet(
            t.ZigbeePacket(
                src=src,
                src_ep=src_ep,
                dst=dst,
                dst_ep=dst_ep,
                tsn=sequence,
                profile_id=profile,
                cluster_id=cluster,
                data=t.SerializableBytes(data),
                extended_timeout=extended_timeout,
                source_route=source_route,
                tx_options=tx_options,
            )
        )

```

dmulcahey:
Right. THis is exactly what was changed.

pipiche38:
Sound ok, but would be very interesting to see how it behave with End Devices, where the Ack is a must to get the end device having time to do the pooling.

PR #1091, opened 2022-11-05
Remove unsupported attributes when they become supported

dmulcahey:
This PR removes unsupported attributes if they become supported in the future. This can be from a firmware update or a quirk update etc. 

~~@puddly let me know if you want to do this differently. I think I covered everything. I have an outstanding note in one of the tests for something that is a bit confusing.~~

### DIFF
--- a/tests/test_appdb.py
+++ b/tests/test_appdb.py
@@ -716,7 +716,6 @@ async def test_unsupported_attribute(tmp_path, dev_init):
     assert "physical_env" in dev.endpoints[3].in_clusters[0].unsupported_attributes
     await app2.shutdown()
 
-
     async def mockrequest(
         is_general_req, command, schema, args, manufacturer=None, **kwargs
     ):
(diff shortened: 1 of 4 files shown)


tests/test_appdb.py
```diff
@@ -705,9 +714,51 @@ async def test_unsupported_attribute(tmp_path, dev_init):
...
+    assert 0x0011 in dev.endpoints[3].in_clusters[0].unsupported_attributes
+    assert "physical_env" in dev.endpoints[3].in_clusters[0].unsupported_attributes
+    await app3.shutdown()
+
+    # Everything should've been saved - check that it re-loads
+    app4 = await make_app(db)
+    dev = app4.get_device(ieee)
+    assert dev.is_initialized == dev_init
+    assert dev.endpoints[3].device_type == profiles.zha.DeviceType.PUMP
+    if dev_init:
```
dmulcahey, line 751: not sure why this is needed but isn't for the existing test? I copied the existing structures to add the functionality and both listeners have this guard:

```
if not cluster.endpoint.device.is_initialized:
            return
```
but only the parts of the test that I added seem to trip up when initialized == False

I must be missing something stupid

dmulcahey: removed the guards and fixed the all success branch as discussed

zigpy/zcl/__init__.py
```diff
@@ -684,7 +684,9 @@ async def configure_reporting_multiple(
         elif isinstance(records, list) and (
             len(records) == 1 and records[0].status == foundation.Status.SUCCESS
         ):
-            self.remove_unsupported_attribute(records[0].attrid)
+            # we get a single success when all are supported
+            for attr in attributes.keys():
+                self.remove_unsupported_attribute(records[0].attrid)
```
puddly, line 689: `records[0].attrid` should be `attr`.

dmulcahey: Good catch 🤦🏻‍♂️

PR #1102, opened 2022-11-15
Application-level request/response matching

puddly:
I'm hoping to move request/response matching into the main ControllerApplication, allowing for more granular control over requests/responses within async code and for dynamically attaching objects that expect responses. In the future, this will allow for sending requests to NWK addresses directly and to wait for a response, without the need for intermediate `Device` objects.

As a proof of concept, below is a synchronous OTA manager that can send a specific OTA image to a specific device, without modifying the global ZCL cluster definitions:

```Python
from __future__ import annotations

import typing
import asyncio
import logging
import contextlib

LOGGER = logging.getLogger(__name__)


from zigpy.zcl import foundation
from zigpy.zcl.clusters.general import Ota


class OTAManager:
    def __init__(
        self,
        device: zigpy.device.Device,
        image: zigpy.ota.image.BaseOTAImage,
        progress_callback=None,
    ) -> None:
        self.device = device
        self.ota_cluster = None

        for ep in device.non_zdo_endpoints:
            try:
                self.ota_cluster = ep.out_clusters[0x0019]
                break
            except KeyError:
                pass
        else:
            raise ValueError("Device has no OTA cluster")

        self.image = image
        self._image_data = image.serialize()
        self.progress_callback = progress_callback

        self._upgrade_end_future = asyncio.get_running_loop().create_future()
        self.stack = contextlib.ExitStack()

    def __enter__(self) -> OTAManager:
        self.stack.enter_context(
            self.device._application.callback_for_response(
                src=self.device,
                filters=[
                    self.ota_cluster.commands_by_name["query_next_image"].schema(),
                ],
                callback=self._image_query_req,
            )
        )

        self.stack.enter_context(
            self.device._application.callback_for_response(
                src=self.device,
                filters=[
                    self.ota_cluster.commands_by_name["image_block"].schema(),
                ],
                callback=self._image_block_req,
            )
        )

        self.stack.enter_context(
            self.device._application.callback_for_response(
                src=self.device,
                filters=[
                    self.ota_cluster.commands_by_name["upgrade_end"].schema(),
                ],
                callback=self._upgrade_end,
            )
        )

        return self

    def __exit__(self, *exc_details) -> None:
        self.stack.close()

    async def _image_query_req(self, hdr, command):
        await self.ota_cluster.query_next_image_response(
            status=foundation.Status.SUCCESS,
            manufacturer_code=self.image.header.manufacturer_id,
            image_type=self.image.header.image_type,
            file_version=self.image.header.file_version,
            image_size=self.image.header.image_size,
            tsn=hdr.tsn,
        )

    async def _image_block_req(self, hdr, command):
        block = self._image_data[
            command.file_offset : command.file_offset + command.maximum_data_size
        ]

        if not block:
            await self.ota_cluster.image_block_response(
                status=foundation.Status.MALFORMED_COMMAND,
                tsn=hdr.tsn,
            )
            return

        await self.ota_cluster.image_block_response(
            status=foundation.Status.SUCCESS,
            manufacturer_code=self.image.header.manufacturer_id,
            image_type=self.image.header.image_type,
            file_version=self.image.header.file_version,
            file_offset=command.file_offset,
            image_data=block,
            tsn=hdr.tsn,
        )

        if self.progress_callback is not None:
            self.progress_callback(
                command.file_offset + len(block), len(self._image_data)
            )

    async def _upgrade_end(self, hdr, command):
        await self.ota_cluster.upgrade_end_response(
            manufacturer_code=self.image.header.manufacturer_id,
            image_type=self.image.header.image_type,
            file_version=self.image.header.file_version,
            current_time=0x00000000,
            upgrade_time=0x00000000,
            tsn=hdr.tsn,
        )

        self._upgrade_end_future.set_result(command.status)

    async def notify(self):
        await self.ota_cluster.image_notify(
            payload_type=(self.ota_cluster.ImageNotifyCommand.PayloadType.QueryJitter),
            query_jitter=100,
        )

    async def wait(self):
        return await self._upgrade_end_future


async def send_firmware(
    device: zigpy.device.Device,
    image: zigpy.ota.image.BaseOTAImage,
) -> typing.Any:
    # Force it to send
    image.header.file_version = 0xFFFFFFFF - 1

    def progress(current, total):
        LOGGER.info(
            "OTA progress for %s: (%d / %d): %0.4f%%",
            device,
            current,
            total,
            (100 * current) / total,
        )

    with OTAManager(device, image, progress_callback=progress) as ota:
        await ota.notify()
        await ota.wait()
```

### DIFF
--- a/zigpy/application.py
+++ b/zigpy/application.py
@@ -900,7 +900,7 @@ class ControllerApplication(zigpy.util.ListenableMixin, abc.ABC):
         filters: list[zigpy.typing.MatcherType],
         callback: typing.Callable[
             [
-                zigpy.zcl.foundation.ZCLHeader | zdo_types.ZDOHeader,
+                zigpy.zcl.foundation.ZCLHeader,
                 zigpy.zcl.foundation.CommandSchema,
             ],
             typing.Any,
--- a/zigpy/application.py
+++ b/zigpy/application.py
@@ -894,7 +894,7 @@ class ControllerApplication(zigpy.util.ListenableMixin, abc.ABC):
             raise ValueError(f"Invalid address: {address!r}")
 
     @contextlib.contextmanager
-    def callback_for_response(
+    def _callback_for_response(
         self,
         src: zigpy.device.Device,
         filters: list[zigpy.typing.MatcherType],
@@ -924,7 +924,7 @@ class ControllerApplication(zigpy.util.ListenableMixin, abc.ABC):
             self._req_listeners[src].remove(listener)
 
     @contextlib.contextmanager
-    def wait_for_response(
+    def _wait_for_response(
         self,
         src: zigpy.device.Device,
         filters: list[zigpy.typing.MatcherType],
--- a/zigpy/application.py
+++ b/zigpy/application.py
@@ -18,6 +18,7 @@ import zigpy.config as conf
 import zigpy.device
 import zigpy.exceptions
 import zigpy.group
+import zigpy.listeners
 import zigpy.ota
 import zigpy.profiles
 import zigpy.quirks
@@ -62,7 +63,7 @@ class ControllerApplication(zigpy.util.ListenableMixin, abc.ABC):
 
         self._req_listeners: collections.defaultdict[
             zigpy.device.Device,
-            collections.deque[zigpy.util.BaseRequestListener],
+            collections.deque[zigpy.listeners.BaseRequestListener],
         ] = collections.defaultdict(lambda: collections.deque([]))
 
     async def _load_db(self) -> None:
@@ -897,7 +898,7 @@ class ControllerApplication(zigpy.util.ListenableMixin, abc.ABC):
     def _callback_for_response(
         self,
         src: zigpy.device.Device,
-        filters: list[zigpy.typing.MatcherType],
+        filters: list[zigpy.listeners.MatcherType],
         callback: typing.Callable[
             [
                 zigpy.zcl.foundation.ZCLHeader,
@@ -910,7 +911,7 @@ class ControllerApplication(zigpy.util.ListenableMixin, abc.ABC):
         Context manager to create a callback that is passed Zigbee responses.
         """
 
-        listener = zigpy.util.CallbackListener(
+        listener = zigpy.listeners.CallbackListener(
             device=src,
             matchers=tuple(filters),
             callback=callback,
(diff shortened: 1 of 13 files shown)


zigpy/application.py
```diff
@@ -863,6 +871,59 @@ def get_device_with_address(
...
+        filters: list[zigpy.typing.MatcherType],
+    ) -> typing.Any:
+        """
+        Context manager to wait for a Zigbee response.
+        """
+
+        listener = zigpy.util.FutureListener(
+            device=src,
+            matchers=tuple(filters),
+            future=asyncio.get_running_loop().create_future(),
```
dmulcahey, line 917: why not just `asyncio.Future()` as opposed to `asyncio.get_running_loop().create_future()`

zigpy/util.py
```diff
@@ -479,3 +485,109 @@ def __getattr__(name: str) -> typing.Any:
         return replacement
 
     return __getattr__
+
+
+@dataclasses.dataclass(frozen=True)
+class BaseRequestListener:
+    device: zigpy.device.Device
+    matchers: tuple[zigpy.typing.MatcherType]
```
dmulcahey, line 493: any reason this is a tuple and not a list? the other signatures are all accepting lists

zigpy/application.py
```diff
@@ -863,6 +871,59 @@ def get_device_with_address(
...
+        listener = zigpy.util.CallbackListener(
+            device=src,
+            matchers=tuple(filters),
+            callback=callback,
+        )
+
+        self._req_listeners[src].append(listener)
+
+        try:
+            yield
```
dmulcahey, line 900: what is this for?

zigpy/application.py
puddly: It's recommended by [the docs](https://docs.python.org/3/library/asyncio-future.html#asyncio.Future):

> The rule of thumb is to never expose Future objects in user-facing APIs, and the recommended way to create a Future object is to call [loop.create_future()](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.create_future). This way alternative event loop implementations can inject their own optimized implementations of a Future object.

Probably makes no difference.

dmulcahey: TIL… cool thx :)

zigpy/util.py
puddly: The listeners are supposed to be immutable, this should be converted into a tuple within `__post_init__`.

zigpy/application.py
puddly: Do you mean the purpose of the method or of the `yield` specifically?

 - It's the `@contextlib.contextmanager` code pattern that turns everything before the `yield` into `__enter__`, the object you yield into `with func() as obj`, and everything after the `yield` into `__exit__`.
 - The purpose of the callback listener is so you can temporarily attach a callback to specific packet types coming from a device, and then detach once you leave the context manager. It's used by the OTA example in the PR.

dmulcahey: I meant the yield itself. I need to look more at context mgrs apparently :) thx for the explanation!

PR #1104, opened 2022-11-17
Update OTA_URLs.md

Hedda:
Update OTA_URLs.md


MattWestb (changes requested):
IKEA test feed is killed (= reversed to on very old version and is not updated) for over one year ago and is better deleting then its only making problems if user is trying downgrading there lights and the device is not liking it.

MattWestb:
@puddly Can you pleas deleting this lines in the IKEA section then its outdated for over one and a half year:
```
Download-URL (Test/Beta-Version): 
http://fw.test.ota.homesmart.ikea.net/feed/version_info.json
```

PR #1109, opened 2022-11-20
Update supported Python versions

blazewicz:
Python versions are inconsistent and outdated, a cleanup is required.

What I did:

1. Updated `setup.py` with a constraint on Python `>=3.8,<3.11`, current version is **not** compatible with 3.7 and 3.11.
2. Updated list of Python versions for test suite in GH Actions to match all supported versions from `setup.py`.
3. Same as 2. but in `tox.ini`.
4. Updated Python minor releases to latest available versions.
5. Removed  `.travis.yml` as I don't see it used anywhere.
6. Updated Github Actions scripts.


dmulcahey:
https://github.com/zigpy/zigpy/pulls?q=is%3Apr+is%3Aclosed+python+

I think the intent was to drop official usage of 3.7 but still allow it to install in 3.7 environments. I don’t think we should be adding 3.7 back to CI

setup.py
```diff
@@ -29,6 +29,6 @@
     license="GPL-3.0",
     packages=find_packages(exclude=["tests", "tests.*"]),
     install_requires=REQUIRES,
-    python_requires=">=3.7",
+    python_requires=">=3.7,<3.11",
```
dmulcahey: Is this necessary?

blazewicz:
Actually current version is not compatible with Python 3.7 either.

```
ERROR tests/test_serial.py - AttributeError: module 'unittest.mock' has no attribute 'AsyncMock'
```

setup.py
blazewicz: I think so, yes. Without this user may install zigpy either directly or as a dependency with Python 3.11 and scripts will fail. With this flag the installation won't succeed.

dmulcahey:
> Actually current version is not compatible with Python 3.7 either.
> 
> ```
> ERROR tests/test_serial.py - AttributeError: module 'unittest.mock' has no attribute 'AsyncMock'
> ```

It may be just the tests but I’m not positive

blazewicz:
> It may be just the tests but I’m not positive

Probably yes, even if so, it should be fixed in a separate PR if support for 3.7 is still expected.

setup.py
dmulcahey: I guess I should have been more clear. I haven’t actually tried to run this on 3.11 yet have you? I was curious if it does in fact have compatibility issues

blazewicz:
Oh, and by "support" I mean that code is ensured to be working on given version by running CI on that version. I don't think 3.7 should be considered supported if CI can't run on it.

If it's only matter of update to the test suite it should be easy to reinstate the support in another PR.

setup.py
blazewicz: Oh, of course, I could have guessed that, sorry.

This is the error while running on Python 3.11:

```
ImportError while loading conftest '/Users/kblazewicz/git/github.com/zigpy/zigpy/tests/conftest.py'.
tests/conftest.py:7: in <module>
    import zigpy.application
zigpy/application.py:12: in <module>
    import zigpy.appdb
zigpy/appdb.py:14: in <module>
    import zigpy.backups
zigpy/backups.py:11: in <module>
    import zigpy.config as conf
zigpy/config/__init__.py:5: in <module>
    from zigpy.config.defaults import (
zigpy/config/defaults.py:1: in <module>
    import zigpy.types as t
zigpy/types/__init__.py:1: in <module>
    from .basic import *  # noqa: F401,F403
zigpy/types/basic.py:391: in <module>
    class bitmap2(bitmap_factory(uint2_t)):
zigpy/types/basic.py:305: in bitmap_factory
    class _NewEnum(int_type, enum.Flag):
zigpy/types/basic.py:309: in _NewEnum
    enum.IntFlag._create_pseudo_member_.__func__
../../../../.pyenv/versions/3.11.0/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/enum.py:767: in __getattr__
    raise AttributeError(name) from None
E   AttributeError: _create_pseudo_member_
```

dmulcahey:
> Oh, and by "support" I mean that code is ensured to be working on given version by running CI on that version. I don't think 3.7 should be considered supported if CI can't run on it.
> 
> If it's only matter of update to the test suite it should be easy to reinstate the support in another PR.

I don’t think it is ensured… IIRC that was done to allow another smart home system to install at their own risk.

blazewicz:
Ok, so you suggest that I set `python_requires=">=3.7,<3.11",` in `setup.py` and leave everything else as is?

Probably we could also update Github Actions scipts, `setup-python` already has version [`4.2.0`](https://github.com/actions/setup-python/releases), while zigpy uses `2.1.4`.

dmulcahey:
> Ok, so you suggest that I set `python_requires=">=3.7,<3.11",` in `setup.py` and leave everything else as is?
> 
> Probably we could also update Github Actions scipts, `setup-python` already has version [`4.2.0`](https://github.com/actions/setup-python/releases), while zigpy uses `2.1.4`.

This is up to @puddly i was just adding what I recalled

blazewicz:
I see that 3.7 was re-added to `setup.py` quite recently in https://github.com/zigpy/zigpy/pull/1031, so I'll keep it in this PR and let @puddly decide.

puddly:
Thanks for the PR!

The changes in #1110 will get zigpy compatible with 3.11 so you can remove that exclusion once it's merged. All the other `zigpy-` packages will unfortunately need to be updated to support this change (to a CPython-internal API) so while the zigpy package will install, you won't be able just yet to use it with a radio.

Regarding 3.7: I wasn't aware of pip's `--ignore-requires-python` flag, which would alleviate this problem for anyone needing to install it older Python environments while allowing us to not advertise 3.7 support on PyPI (@pipiche38). I think 3.7 can be removed from all packaging files now.

pipiche38:
we still have a lot of user on python3.7 , but I do understand your issue.

blazewicz:
That's great news @puddly! I've updated the PR.

PR #1114, opened 2022-11-29
Deprecate the IKEA OTA test feed

puddly:
Addresses https://github.com/zigpy/zigpy/pull/1104#issuecomment-1329495639

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -1,5 +1,7 @@
 """Test configuration."""
 
+import warnings
+
 import pytest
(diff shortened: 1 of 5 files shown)


puddly:
The default Codecov branch somehow switched from `dev` to `master`. Coverage should work now.

MattWestb:
Thanks **P** !!
I think its better not confusing our brave user with inaccurate / old information.

Im looking how its looking in Zigpy Wiki and is need updated but i think i have writing its outdated.

PR #1124, opened 2022-12-06
Add big endian int types

Shulyaka:
There are several places where we can replace workarounds and reimplementations with a central implementation of big endian types.
[Example1](https://github.com/zigpy/zha-device-handlers/blob/31245e0e3643c5011be849a4ebbda39a5eb379e6/zhaquirks/tuya/__init__.py#L129)
[Example2](https://github.com/zigpy/zha-device-handlers/blob/790c90b2bf490345d959776abaa03abb515da785/zhaquirks/xiaomi/aqara/feeder_acn001.py#L148)
[Example3](https://github.com/zigpy/zigpy-xbee/blob/dev/zigpy_xbee/types.py#L33)

### DIFF
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -12,6 +12,8 @@ def test_abstract_ints():
     assert not issubclass(t.uint8_t, t.int_t)
     assert t.int_t._signed is True
     assert t.uint_t._signed is False
+    assert t.int_t._bigendian is False
+    assert t.be_int_t._bigendian is True
 
     with pytest.raises(TypeError):
         t.int_t(0)
@@ -139,6 +141,18 @@ def test_ints_signed():
     n = t.int8s(-126)
     bits = [1, 0] + t.Bits.deserialize(n.serialize())[0]
     assert t.int8s.from_bits(bits) == (n, [1, 0])
+
+
+def test_bigendian_ints():
+    assert t.be_uint32_t(0x12345678).serialize() == b"\x12\x34\x56\x78"
+    assert t.be_uint32_t.deserialize(b"\x12\x34\x56\x78") == (0x12345678, b"")
+    assert t.be_int32s(0x12345678).serialize() == b"\x12\x34\x56\x78"
+    assert t.be_int32s(-1).serialize() == b"\xff\xff\xff\xff"
+    assert t.be_int32s.deserialize(b"\xfe\xdc\xba\x98") == (-0x01234568, b"")
+    assert (
+        t.be_uint32_t(0x12345678).serialize()[::-1]
+        == t.uint32_t(0x12345678).serialize()
+    )
 
 
 def test_bits():
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -12,8 +12,8 @@ def test_abstract_ints():
     assert not issubclass(t.uint8_t, t.int_t)
     assert t.int_t._signed is True
     assert t.uint_t._signed is False
-    assert t.int_t._bigendian is False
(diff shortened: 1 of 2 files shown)


zigpy/types/basic.py
```diff
@@ -81,6 +81,7 @@ class FixedIntType(int):
     _signed = None
     _bits = None
     _size = None  # Only for backwards compatibility, not set for smaller ints
+    _bigendian = None
```
puddly, line 84: I think `_byteorder` and turning it into a string would be a better approach, it matches the `int.from_bytes` API and I think is a clear name.

zigpy/types/basic.py
```diff
@@ -171,7 +177,9 @@ def serialize(self) -> bytes:
         if self._bits % 8 != 0:
             raise TypeError(f"Integer type with {self._bits} bits is not byte aligned")
 
-        return self.to_bytes(self._bits // 8, "little", signed=self._signed)
+        return self.to_bytes(
+            self._bits // 8, "big" if self._bigendian else "little", signed=self._signed
```
puddly, line 181: And then `"big" if self._bigendian else "little"` would just become `byteorder=self._byteorder`.

zigpy/types/basic.py
```diff
@@ -183,16 +191,18 @@ def deserialize(cls, data: bytes) -> tuple[FixedIntType, bytes]:
         if len(data) < byte_size:
             raise ValueError(f"Data is too short to contain {byte_size} bytes")
 
-        r = cls.from_bytes(data[:byte_size], "little", signed=cls._signed)
+        r = cls.from_bytes(
+            data[:byte_size], "big" if cls._bigendian else "little", signed=cls._signed
```
puddly, line 195: Same here: `byteorder=self._byteorder`

tests/test_types.py
```diff
@@ -12,6 +12,8 @@ def test_abstract_ints():
     assert not issubclass(t.uint8_t, t.int_t)
     assert t.int_t._signed is True
     assert t.uint_t._signed is False
+    assert t.int_t._bigendian is False
+    assert t.be_int_t._bigendian is True
```
puddly, line 16: What do you think about `t.int_t_be`? Having a `be_` prefix makes it sound a little strange when read aloud.

puddly (commented):
Thanks! I think consolidating types will be a useful change.

tests/test_types.py
Shulyaka: Ha-ha, never thought of it!
I have no particular preference, and I couldn't find any existing convention. Will change.

Shulyaka:
Rebased to the latest commit

zigpy/types/basic.py
```diff
@@ -171,7 +177,7 @@ def serialize(self) -> bytes:
         if self._bits % 8 != 0:
             raise TypeError(f"Integer type with {self._bits} bits is not byte aligned")
 
-        return self.to_bytes(self._bits // 8, "little", signed=self._signed)
+        return self.to_bytes(self._bits // 8, self._byteorder, signed=self._signed)
```
javicalle, line 180: I haven't find any reference here or in `zha-device-handlers` but maybe a breaking change if somewhere extending `FixedIntType` and not initializing class with `byteorder="little"`.

Shulyaka: I think they are not supposed to extend `FixedIntType`

Shulyaka: Anyway, I've made little-endian the default.

PR #1128, opened 2022-12-13
Update python version in CONTRIBUTING.md

TacoV:
The Python version 3.7 in the example results in an error.

````
$ pip install -e '.[testing]'
...
ERROR: Package 'zigpy' requires a different Python: 3.7.16 not in '>=3.8'
````

Updating to 3.8 resolves this.


TacoV:
Which is in line with the requirements in setup.py

https://github.com/zigpy/zigpy/blob/767342cc27cd28e7d87bb989823c862f9f4ffaf9/setup.py#L34

PR #1165, opened 2023-02-23
Generic OTA providers

puddly:
Implements the format specified in https://github.com/zigpy/zigpy/wiki/OTA-Information-for-Manufacturers:

```yaml
zha:
  zigpy_config:
    ota:
      remote_providers:
        - url: "https://fw.zigbee.example.org/ota.json"
          manufacturer_ids: [0x1234, 0x5678]
        - url: "https://fw.zigbee.example.org/ota-beta.json"
```

This allows for new providers to be added to zigpy without extra code.  This also allows for manufacturers to distribute test feed URLs to customers without deploying them globally.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -104,7 +104,7 @@ def recursive_dict_merge(
     result = copy.deepcopy(obj)
 
     for key, update in updates.items():
-        if isinstance(update, dict):
+        if isinstance(update, dict) and key in result:
             result[key] = recursive_dict_merge(result[key], update)
         else:
             result[key] = update
--- a/tests/ota_providers/test_ota_provider_remote.py
+++ b/tests/ota_providers/test_ota_provider_remote.py
@@ -1,0 +1,151 @@
+import hashlib
+import logging
+from unittest import mock
+
+import pytest
+
+import zigpy.ota
+import zigpy.ota.image
+from zigpy.ota.provider import RemoteImage, RemoteProvider
+
+from tests.async_mock import AsyncMock
+from tests.conftest import make_app
+
+MANUFACTURER_ID_1 = 0x1234
+MANUFACTURER_ID_2 = 0x5678
+IMAGE_TYPE = 0xABCD
+
+
+@pytest.fixture
+def provider():
+    p = RemoteProvider(
+        url="https://example.org/ota/",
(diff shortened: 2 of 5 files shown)


Hedda:
> Implements the format specified in https://github.com/zigpy/zigpy/wiki/OTA-Information-for-Manufacturers

@puddly believe a schema/format like this could have the potential of becoming more than a zigpy-only standard, therefore, I would like to suggest that you create a separate dedicated repository for that JSON schema and move the documentation for it from the zigpy wiki to files in that repo instead, (similarly to how you did with the "[Open ZigBee Coordinator Backup Format](https://github.com/zigpy/open-coordinator-backup)"), as then it might be both simpler to keep different versions of the schemas in a Git repository instead of a wiki as well as be easier for developers from other projects to contribute to it a common standard as it could have the potential to become a universal [open standard](https://en.wikipedia.org/wiki/Open_standard) and common [open file format](https://en.wikipedia.org/wiki/Open_file_format) that can be shared between different projectas and companies, at least if it someday also become commonly reused by different project within the Zigbee open-source scene in the future, such as  adopted by others open source Zigbee framework/gateway projects, and used by more than a few commercial hardware device manufacturer companies as a shared open standard.

MattWestb:
Great work done i hope IKEA is doing one updated OTA for there Dirigera with this method then is very likely in the future is getting more device then the old then all devices is not working with the old system like the upcoming windows and door sensor.

puddly:
> i hope IKEA is doing one updated OTA for there Dirigera with this method

This method relies on the zigpy-specific [JSON index format](https://github.com/zigpy/zigpy/wiki/OTA-Information-for-Manufacturers) so I doubt IKEA will be implementing it any time soon 😄.

What I hope to do is create a `zigpy/ota` repo to host all OTA images (including IKEA) and to use that instead of the manufacturer-specific HTTP feeds we currently have. This will let us vet OTA images before releasing them instead of allowing manufacturers to push arbitrary, potentially-breaking updates (e.g. IKEA changing their remotes' behavior).

MattWestb:
You / we also need documenting if devices is blocked for firmware downgrade like all IKEA controllers is (last checked / tested all first gen i having for one year ago its also in the Zigbee cert report) so users is knowing if its possible doing it or not but it can being made in the wiki so users can adding known devices then have testing and it was not working.

#### Great work done as normally **P** !!

PR #1178, opened 2023-03-13
Do not treat IEEE address responses as rejoins

puddly:
They will only be treated as such when then NWK address changes. Otherwise, nothing should happen (because no device info changed). Mainly affects bellows, as it passes through stack-generated ZDO responses back to the application.


dmulcahey (approved):
Thanks for taking care of this!!

PR #1183, opened 2023-03-20
Implement energy scanning during startup and on network formation

puddly:
`ControllerApplication.energy_scan` will allow us to perform a startup energy scan on the configured channel to warn about interference, in addition to performing an energy scan during network formation to pick a better Zigbee channel than just a hard-coded default of 15.

TODO:

- [x] Log a warning on startup if the channel utilization is high (what is a good threshold?)
- [x] Drop 15 as a hard-coded default and scan channels during network formation


Hedda:
Suggest that Home Assistant's ZHA integration still primarily recommends and pick Zigbee Channels 15, 20, 25 when forming the network, (and as standard by default not recommend or pick any other Ziggee channels), even if other Zigbee Channels appear to be less utilized during the initial energy scan, as odds are there is a greater chance that the frequency range used by the other Zigbee Channels will be utilized later by Wi-Fi if the user set their WiFi router to automatic or using more WiFi Access Points in their house or migrate to a WiFi Mesh system.

https://www.home-assistant.io/integrations/zha#defining-zigbee-channel-to-use

_The general recommendation is to only use channels 15, 20, or 25 in order to avoid interoperability problems with Zigbee devices that are limited to only being compatible with the ZLL (Zigbee Light Link) channels as well as lessen the chance of Wi-Fi networks interfering too much with the Zigbee network. Note that especially using Zigbee channels 11, 24, 25, or 26 on your Zigbee Coordinator could mean it will probably not be accessible to older devices as those Zigbee channels are commonly only supported by relatively modern Zigbee hardware devices with newer Zigbee firmware._ 

_Regardless, note that the best practice recommendation is, however, not to change the Zigbee channel from default as not all Zigbee devices support all channels. If you have issues with overlapping frequencies, then it will generally be a better idea to just change Wi-Fi channels on your Wi-Fi Router or all your Wi-Fi Access Points instead._

https://www.metageek.com/training/resources/zigbee-wifi-coexistence/

https://support.metageek.com/hc/en-us/article_attachments/115017048148/ZigBee_Channels.pdf

MattWestb:
I suggesting adding ch 11 as it one primary channel in ZLL and Zigbee 3 but some device can joining on it but on the other hand some other devices is having problem with other channels.

I think is wrong way blocking using primary channels better informing the user looking if having devices that require special treatment for working OK in the system.

PR #1205, opened 2023-04-30
Keep track of the last updated time of attributes

puddly:
This will allow ZHA to more intelligently decide which devices need to be polled on startup. I would like to make the default behavior not overload the network with requests, with an optional setting to re-enable the old behavior for people who turn off bulbs.

Exposing this information will require some sort of new attribute API, since accessing private dictionaries is not a good one.

---

In the future, I think it would be useful to have some sort of unified proxy for attribute polling/subscriptions:

```Python
await self.subscribe_to_attribute("attr_name", min=5, max=30, delta=1, callback=self.attr_updated)
```

This can then automatically setup attribute reporting (if necessary) and resort to polling when a device does not report in the expected time period.

### DIFF
--- a/tests/test_appdb.py
+++ b/tests/test_appdb.py
@@ -1,5 +1,5 @@
 import asyncio
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 import pathlib
 import sqlite3
 import sys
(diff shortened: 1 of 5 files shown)


MattWestb:
Puddly have you looking of implanting query for commands to sleepers and using pull control functionality then the device is doing check in to the coordinator and then sending queered commands to the device and then all is sent sending long pull to the device ((Z2M have implanting it with some logic) it shall working good with ZB3 devices that have implanting it OK like IKEA controllers.

MattWestb:
Ops was forgetting:
## GREAT WORK DONE !!

PR #1206, opened 2023-05-04
New cluster attribute and command format

dmulcahey:
This PR migrates cluster commands and attributes from dicts to concrete objects. This will aid in typing and it removes the need to pass magic #'s and raw strings around in the ZHA and associated lib codebases. 

# Notes

Documenting some of the VSCode find / replace regex that were used to do a chunk of the work:

## Attributes
FIND: `(.*): ZCLAttributeDef\("(.*)", type=(.*),`
REPLACE: `$2: Final = ZCLAttributeDef(id=$1, type=$3`

## Commands
FIND: `(.*): ZCLCommandDef\("(.*)", \{(.*)\},(.*),`
REPLACE: `$2: Final = ZCLCommandDef(id=$1, schema={$3}, direction=$4`

### DIFF
--- a/tests/test_quirks.py
+++ b/tests/test_quirks.py
@@ -330,23 +330,12 @@ def test_custom_cluster_idx():
                 id=0x01, schema={"param1": t.uint8_t}, direction=True
             )
 
-    def _test_cmd(cmd_set, cmd_set_idx):
-        assert hasattr(TestClusterIdx, cmd_set_idx)
-        idx_len = len(getattr(TestClusterIdx, cmd_set_idx))
-        cmd_set_len = len(getattr(TestClusterIdx, cmd_set))
-        assert idx_len == cmd_set_len
-        for cmd_name, cmd_id in getattr(TestClusterIdx, cmd_set_idx).items():
-            assert getattr(TestClusterIdx, cmd_set)[cmd_id].name == cmd_name
-
     assert hasattr(TestClusterIdx, "attributes_by_name")
     attr_idx_len = len(TestClusterIdx.attributes_by_name)
     attrs_len = len(TestClusterIdx.attributes)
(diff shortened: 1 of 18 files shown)


jeverley:
Hi @dmulcahey, @puddly  is there a reason the bits of _MeasurementType_ and _ACAlarmsMask_ are shifted left by 2 rather than 1?

I'm looking to use the MeasurementType values in my energy meter quirk rather than manually defining them, however the current class values are shifted 1 bit to the left of where I'd expect them to be i.e. it gives 0b00000010 for Active_measurement_AC rather than the expected 0b00000001.

It doesn't look like the classes are currently used outside of _homeautomation.py_ where you're using them for the attribute types. I'd be happy to raise a PR amending these to use the correct bitshift values.


```
class MeasurementType(t.bitmap32):
    Active_measurement_AC = 2 << 0
    Reactive_measurement_AC = 2 << 1
    Apparent_measurement_AC = 2 << 2
    Phase_A_measurement = 2 << 3
    Phase_B_measurement = 2 << 4
    Phase_C_measurement = 2 << 5
    DC_measurement = 2 << 6
    Harmonics_measurement = 2 << 7
    Power_quality_measurement = 2 << 8

class ACAlarmsMask(t.bitmap16):
    Voltage_Overload = 2 << 0
    Current_Overload = 2 << 1
    Active_Power_Overload = 2 << 2
    Reactive_Power_Overload = 2 << 3
    Average_RMS_Over_Voltage = 2 << 4
    Average_RMS_Under_Voltage = 2 << 5
    RMS_Extreme_Over_Voltage = 2 << 6
    RMS_Extreme_Under_Voltage = 2 << 7
    RMS_Voltage_Sag = 2 << 8
    RMS_Voltage_Swell = 2 << 9
```

PR #1230, opened 2023-07-06
Merge `master` into `dev`

puddly:
Consolidates the history of both branches so that `master` can be deprecated.


gspr:
Would it be possible to *actually* merge the master branch into dev as well, not just "consolidate the histories"? I'm asking because the Debian packaging for zigpy is based off of the old master branch. Switching to be based on dev is much cleaner if that old master is actually merged into dev, rather than dev just "consolidating histories". Much appreciated! (One might also argue that it makes the zigpy development history easier to follow, too).

puddly:
@gspr sure! Could you explain a little more what this would require?

gspr:
puddly ***@***.***> writes:

> @gspr sure! Could you explain a little more what this would require?

I assume that the idea is that "master" is to be retired, and "dev" is
to continue as the main development branch?

On "dev", you would then "git merge master" (or whatever the appropriate
"end commit" in master is).

puddly:
Ah, I assumed the GitHub PR would do the same. Should be done now.

gspr:
puddly ***@***.***> writes:

> Ah, I assumed the GitHub PR would do the same. Should be done now.

Perfect, now everything looks good! Thanks.

And just to confirm: the idea is for "dev" to be the main development
branch going forward, right?

PR #1256, opened 2023-09-08
Use first existing endpoint for basic cluster instead of hardcoded number

Shulyaka:
On #1238 we add Basic cluster to endpoint 1.
This PR makes sure that that endpoint actually exists.

Mostly applicable to xbee coordinators: https://github.com/zigpy/zigpy-xbee/blob/dev/zigpy_xbee/zigbee/application.py#L381

### DIFF
--- a/zigpy/application.py
+++ b/zigpy/application.py
@@ -1221,6 +1221,9 @@ class ControllerApplication(zigpy.util.ListenableMixin, abc.ABC):
         return self.get_device(ieee=self.state.node_info.ieee)
 
     def _persist_coordinator_model_strings_in_db(self) -> None:
+        if 1 not in self._device.endpoints:
+            self._device.add_endpoint(1)
+
         cluster = self._device.endpoints[1].add_input_cluster(
             zigpy.zcl.clusters.general.Basic.cluster_id
         )
--- a/zigpy/application.py
+++ b/zigpy/application.py
@@ -1221,10 +1221,7 @@ class ControllerApplication(zigpy.util.ListenableMixin, abc.ABC):
         return self.get_device(ieee=self.state.node_info.ieee)
 
     def _persist_coordinator_model_strings_in_db(self) -> None:
-        if 1 not in self._device.endpoints:
-            self._device.add_endpoint(1)
-
-        cluster = self._device.endpoints[1].add_input_cluster(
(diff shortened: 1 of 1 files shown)


puddly:
Thanks! Does the xbee have an API to register new endpoints? Zigpy radio libraries should implement the `register_endpoint` method, which would have set up endpoint 1 on startup.

zigpy/application.py
```diff
@@ -1221,6 +1221,9 @@ def _device(self) -> zigpy.device.Device:
         return self.get_device(ieee=self.state.node_info.ieee)
 
     def _persist_coordinator_model_strings_in_db(self) -> None:
+        if 1 not in self._device.endpoints:
+            self._device.add_endpoint(1)
+
         cluster = self._device.endpoints[1].add_input_cluster(
             zigpy.zcl.clusters.general.Basic.cluster_id
         )
```
Shulyaka: No, it does not, however I believe new endpoints can be implemented on the application level.

If you feel uncomfortable adding an endpoint, we could instead pick the first existing one like this:

```suggestion
        cluster = self._device.non_zdo_endpoints[0].add_input_cluster(
            zigpy.zcl.clusters.general.Basic.cluster_id
        )
```

Please advise.

puddly: Perhaps it would be best for the XBee implementation to call `await self.register_endpoints()` within `start_network` and then to implement `register_endpoint` to just call `add_endpoint`. This is more of a radio library bug more than a zigpy bug, in my opinion.

Shulyaka:
Updated. It looks like a more clean solution

zigpy/application.py
Shulyaka: Is endpoint 1 mandatory or is zigpy just making an assumption?

puddly: Yes. Zigpy registers endpoints explicitly on startup: https://github.com/zigpy/zigpy/blob/32a3f050327618a66f88a8b3d77306804826a9cd/zigpy/application.py#L701

Shulyaka: Thanks! I will check.

Shulyaka:
Closed in favor of https://github.com/zigpy/zigpy-xbee/pull/145 and https://github.com/zigpy/zigpy-zigate/pull/146.

**UPD**: Do we also need a similar fix for zigpy_deconz?

PR #1262, opened 2023-09-29
Don't shut down the database when `startup` has been manually called

puddly:
ZHA currently initializes the zigpy application before attempting to connect multiple times. The application database listener does not properly work after the second shutdown/reconnect attempt.

To work around this, `shutdown` must manually be called when disconnecting.

### DIFF
--- a/tests/test_application.py
+++ b/tests/test_application.py
@@ -437,24 +437,6 @@ async def test_remove_parent_devices(app, make_initialized_device):
         assert parent.zdo.request.await_count == 1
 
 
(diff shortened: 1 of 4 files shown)


MattWestb:
I was having one RCP system that was have the DB clocked and was not able to writing most things like last seen and so on for 3 days and was not able getting it away also after reboot of the OS. After rebooting one other kernel for testing with Debrian the error is gone and have restarting HA 5 times and its working OK. I hope this is fixing it if its happening agen.
PS: I was rebooting the system from ubuntu CLI and and not taking down HA before it so not 100% clean HA shutdown.

PR #1283, opened 2023-11-13
Create a `SendError` exception type

puddly:
This will allow send failures (CCA failure) to be differentiated from delivery failures (no ACK).

To ensure compatibility with existing downstream applications, it is a subclass of `DeliveryError` and not a completely new exception type.


pipiche38:
looks an interesting one.

PR #1287, opened 2023-11-15
Add `baudrate` and `flow_control` keys to zigpy config and centralize `probe` function

puddly:
Radio libraries should ensure `connect` is all that is required to probe and then implement a list of probe configs with `_probe_configs`, leaving probing to zigpy.

The new config keys are already present in all radio libraries. This will allow their configs to be simplified. In the future, radio libraries will expose only a single library-specific key and not modify the zigpy config directly.


Hedda:
@DamKast have you implementated this in your zigpy-zboss radio library?

https://github.com/kardia-as/zigpy-zboss/

PR #1295, opened 2023-12-10
Only feed the watchdog when we have time

puddly:
It appears that ZHA startup is so noisy that the per-radio watchdog serial commands get jammed up, causing continuous disconnects.

To temporarily mitigate this issue, do not feed the watchdog if the concurrent request semaphore is full.

### DIFF
--- a/tests/test_application.py
+++ b/tests/test_application.py
@@ -1,5 +1,4 @@
 import asyncio
-from datetime import datetime, timezone
 import errno
 import logging
 from unittest import mock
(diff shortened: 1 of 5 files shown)


puddly:
I've reverted pausing the watchdog on device activity as bellows relies on it for counter polling.

dmulcahey:
Looks like cancel all was reverted too?

puddly:
I reverted it because it changed functionality on disconnect and introduced a new exception type. Without it, this PR is very small.

I think it would be best to figure out how to centralize this within radio libraries so that all outstanding requests are cancelled automatically on disconnect.

dmulcahey:
Think it is worth striking from the PR description? JIC we go hunting through history at some point

PR #1297, opened 2023-12-15
Implement priority within semaphore and create a priority lock

puddly:
This allows radio libraries to replace their use of `asyncio.Lock` with `zigpy.datastructures.PriorityLock` and prioritize commands:

```python
if request in ("send_aps_data",):
    priority = -1
elif request in ("write_watchdog_ttl",):
    priority = 1000
else:
    priority = 0

async with self._lock(priority=priority):
    ...
```

This class can be extended in the future to account for time spent in a lock as well, to prevent low-priority commands from holding the lock indefinitely if they are run in a tight loop. This isn't done in any radio so for now, this should allow APS requests to be deprioritized within radio libraries.

In the future, we can also pull the semaphore use out of individual radio libraries and allow zigpy to prioritize requests.


puddly:
There are a few linked PRs now for a few radio libraries. The most important are deCONZ and bellows, as they are currently affected by watchdog issues the worst.

To test, I've increased the watchdog period in bellows to 5s. ZHA is able to start up with nearly 100 enqueued requests without the watchdog ever complaining a single time.

Hedda:
@DamKast does this also affect the zigpy-zboss radio library?

https://github.com/kardia-as/zigpy-zboss

dumpfheimer:
Would it easily be possible to let higher level code assign a priority to single packets?
It could be pretty awesome to prioritize within commands for example (deprioritize async attribute updates vs light toggle commands)

Also, is it somehow possible to filter a semaphore? I do not really know what it technically is, but it would also be pretty nice to cancel all existing requests that target the same address+endpoint+cluster+attribute/command which could reduce network load and usability issues (for example if the lights react slowly, and a bunch of lights are turned on and off in succession, it would be great to have only the last command be executed. for example a series of on/off/on/off could be cut down to only the last off command.

puddly:
> Would it easily be possible to let higher level code assign a priority to single packets?

In the future, yes. I want to move this into zigpy directly and move the concurrency limiting out of radio libraries.

> I do not really know what it technically is

A semaphore for us is like a generalized lock: a lock lets only one thing access a resource, a semaphore lets no more than *n* things access a resource. If anything else tries and there are no more slots, they wait until a slot opens up. So a lock is just a semaphore with *n = 1*

> Also, is it somehow possible to filter a semaphore?

I think that's outside the scope of a semaphore directly but we can look into it globally. It would be interesting to group commands based on intent (`on`/`off`/`toggle` all controlling the state so they should not be queued up) and cancel existing commands in a retry state if another is queued up.

PR #1307, opened 2023-12-30
Only create pending requests if we are expecting a reply

puddly:
@dmulcahey noticed a race condition with attribute reports: if a device expects a default reply to its attribute report *and* our reply takes a while to send, the device may send a duplicate attribute report while retrying. This duplicate will require us to reply with the same TSN, which is currently locked by the current request, causing it to never send.

We should not create a `Request` context if we do not expect a reply.


MattWestb:
Its have being around for some releases on EZSP NCP and RCP and being great fixed !!!!

PR #1317, opened 2024-01-05
Create CODE_OF_CONDUCT.md

MattWestb:
Zigpy is missing Code of Conduct and i think its perhaps is not needed but it good to have then somthing bad is happening so i have copy HAs and if some one like doing changes is more then welcome from my point of view.


puddly:
I think this would be fine to have. Thank you!

MattWestb:
Thanks **P** !!

PR #1319, opened 2024-01-07
Fix dangling coroutine 

puddly:
If the request context manager throws an exception, we currently leave a dangling coroutine. This PR avoids creating the coroutine until we are ready to `await`.


dmulcahey (approved):
🎉

PR #1321, opened 2024-01-08
Update OTA handling

dmulcahey:
Original Ota manager implementation from the description here: https://github.com/zigpy/zigpy/pull/1102. (Thanks @puddly)

All Ota handling is removed from the cluster definition in Zigpy except for image querying. Thinking the usage will be something like this: 

HA update entity executes some flavor of this code at some pre defined interval to determine if there is an update available:

```python
        await self._ota_cluster_handler.image_notify(
            payload_type=(
                self._ota_cluster_handler.cluster.ImageNotifyCommand.PayloadType.QueryJitter
            ),
            query_jitter=100,
        )
```

This will cause Zigpy to check for a new FW and if one is found an event with the firmware will be pushed to HA. 

when the user clicks the update button in HA `async def update_firmware` is executed passing in the progress callback from HA.

Note: I tried to keep as much of the logging similar to what was being sent before.


Hedda:
That sounds great! For reference please also check out the directly related OTA update feature discussions for ZHA here -> https://github.com/zigpy/zigpy/discussions/951

PR #1334, opened 2024-02-04
Rename input and output clusters to server and client

puddly:
This better matches the names given to the clusters in the Zigbee specifications.


puddly:
It looks like some more research needs to be done for this PR.

The ZCL and ZDO specifications talk about both input and output clusters, in addition to server and client clusters 🤔

PR #1340, opened 2024-02-12
OTA v2: provider and metadata rewrite

puddly:
Corresponding HA Core branch: https://github.com/puddly/core/tree/puddly/zigpy-new-ota

---

This is a rewrite of all the current current OTA providers and OTA image metadata logic. Now that ZHA handles initiating updates, no provider will ever be contacted by default on startup, even if you enable them: the index will be downloaded only if a device compatible with the provider requests an image.

The main reason for this rewrite is image matching and metadata. OTA images alone do not contain enough data to exclude bad devices (i.e. Tuya). Secondary metadata needs to be bundled with the OTA image. For this reason, all classes dealing with OTA will now work with `OtaImageWithMetadata` objects instead of raw OTA images.

New OTA metadata:

 - `min_current_file_version` and `max_current_file_version`: allows for chained updates (i.e. A -> B -> C)
 - `manufacturer_names`: whitelist only specific manufacturers
 - `model_names`: whitelist only specific models
 - `specificity`: boost the numerical specificity of a specific OTA image.

All of this metadata will be used to exclude bad devices (i.e. Tuya).

From an end-user perspective, `otau_dir` is gone and has been replaced with `z2m_local_index`/`z2m_remote_index`. If you want to risk updating with OTA files without metadata, `advanced_ota_dir` exists.

Example:

```yaml
zigpy:
  ota:
    # It's enabled by default
    enabled: true

    # Periodically send out a network-wide (including end devices!) image notification
    broadcast_enabled: true
    broadcast_initial_delay: 14400
    broadcast_interval: 14400

    # You probably do not need this, it's for development
    advanced_ota_dir: /path/to/raw/ota/files
    allow_advanced_ota_dir: ...

    # Points to Z2M OTA `index.json` if stored locally
    z2m_local_index: /path/to/z2m-ota/index.json

    # Points to Z2M OTA `index.json` URL
    z2m_remote_index: "https://raw.githubusercontent.com/Koenkk/zigbee-OTA/master/index.json"
```


puddly:
I've added one final feature that I think is useful for OTA: `app.ota.broadcast_notify()`. This command will send a network-wide broadcast to *all* devices (including end devices!) and ask that they check in. It is enabled by default.

For me, it pulls in OTA updates for most battery-powered sensors. It'll be re-sent once every 3.9 hours.

puddly:
I've tuned the `query_jitter` down from 100 (every device receiving the broadcast will check in) to a value computed based on the number of devices on your network. The goal is to have 15 random devices check at once per 4 hour background broadcast.

- For a 60 device network, that's an 82% chance every device will respond within 24 hours. 97% in two days, etc.
- For a 200 device network, hitting 97% will take a week.

If you want to check for updates more quickly, you can do so with the "check for updates" button in ZHA. This is just for automatic notifications.

dmulcahey:
image

PR #1341, opened 2024-02-13
Allow clearing attributes from the cache

puddly:
This PR introduces an `attribute_cleared` event (perhaps there's a more suitable name?) which indicates that an attribute has been cleared from cache. You trigger it setting the attribute value to `None`:

```python
ep.basic.update_attribute(Basic.AttributeDefs.zcl_version.id, None)
```

The purpose of this PR is to clear the OTA cluster's `current_file_version` attribute upon completing an OTA update. I think it would also be useful to fully re-initialize a device after OTA by purging its attributes from cache and reading them anew. (This would also be good default behavior for the "Reconfigure" button in ZHA).

### DIFF
--- a/tests/test_appdb.py
+++ b/tests/test_appdb.py
@@ -1147,7 +1147,12 @@ async def test_appdb_attribute_clear(tmp_path):
         dev2.endpoints[1].basic._attr_cache[Basic.AttributeDefs.zcl_version.id] == 0x12
     )
 
+    # This attribute exists
     dev2.endpoints[1].basic.update_attribute(Basic.AttributeDefs.zcl_version.id, None)
+
+    # This attribute won't exist
+    dev2.endpoints[1].basic.update_attribute(Basic.AttributeDefs.manufacturer.id, None)
+
     assert Basic.AttributeDefs.zcl_version.id not in dev2.endpoints[1].basic._attr_cache
     await app2.shutdown()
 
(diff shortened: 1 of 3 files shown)


MattWestb:
Out of this PR about function / changing device EP / cluster OTA updates:
Some update is adding / deleting / moving endpoints and cluster on them = function changing OTA updates i calling them.
The prblem is that quirk is matched ageist the saved information zigbee.db that  is not being updated.
If doing it nice wee need updating all so need discovery EP and correcting if being changes and then all cluster on them and also the attributes saved from them = complete init of the device.

The question is shall we doing this work or letting the user deleting the device and adding it new ?
Current case is IKEA styrbar that was having wrong  metadata in TF feed but looks being OK with Dirigera so users have getting updated device but the battery is using the old matched quirk, Deleting  and adding new we is losing bindings and mst redo them and can being that automation is also being broken and must being fixed of the user (not sure of the last).

If looking how IKEA is doing in Diriger then joining one device looks very interesting they is sending fast pull to the device (if end device) after getting the information of device type and then requesting endpoint list and then requesting all cluster information from all endpoints as they is only having real Zigbee 3 devices its looks working well (have not looking how its working with one device on old ZLL firmware).
The same shall being possible also with one full initiated device after OTA is being OK and the version is ticked (its needed then the device can throwing the OTA then its rebooting and not being updated after sending OTA OK).

If you like having one sniff from Dirigera pairing i can do one for you with one router and one end device for see how they is doing it.

PR #1353, opened 2024-03-03
Attributes can also be called by names.

gmsoft-tuxicoman:
While working on adding tests for https://github.com/zigpy/zha-device-handlers/pull/1961, ZCL was unable to log the correct error when writing the cluster attribute "schedule".

This fixes the problem by handling both int and other types of attribute names.


dmulcahey:
thx

PR #1371, opened 2024-03-21
fix: convert CONF_OTA_ADVANCED_DIR to a Path

nepeat:
This fixes the following when doing OTA updates with OTA files.

```
2024-03-21 00:34:26.503 DEBUG (MainThread) [zigpy.ota] Failed to load provider <zigpy.ota.providers.AdvancedFileProvider object at 0x7fb84ff994c0>
  File "/usr/local/lib/python3.12/site-packages/zigpy/ota/__init__.py", line 333, in get_ota_image
    index = await self._load_provider_index(provider)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/zigpy/util.py", line 491, in replacement
    return await tasks[key]
           ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/zigpy/ota/__init__.py", line 309, in _load_provider_index
    return await provider.load_index()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/zigpy/ota/providers.py", line 160, in load_index
    return [meta async for meta in self._load_index(session)]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/zigpy/ota/providers.py", line 432, in _load_index
    for path in self.image_dir.rglob("*"):
                ^^^^^^^^^^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'rglob'
```

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -1,4 +1,5 @@
 """Test configuration."""
+
 import pathlib
 import warnings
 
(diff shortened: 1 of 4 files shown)


puddly:
Ah, good catch! I think it would be best to convert it during validation instead of passing a string to begin with: https://github.com/zigpy/zigpy/blob/3d1b60c9181fb40638eaf3e70c5815967b07300c/zigpy/config/__init__.py#L188

There is a similar code path in the config for accepting a JSON file.

nepeat:
> Ah, good catch! I think it would be best to convert it during validation instead of passing a string to begin with:
> 
> https://github.com/zigpy/zigpy/blob/3d1b60c9181fb40638eaf3e70c5815967b07300c/zigpy/config/__init__.py#L188
> 
> There is a similar code path in the config for accepting a JSON file.

Much better thought than my quick and dirty fix. Today I'm one of the lucky 10k to learn about voluptuous.

I've done exactly that and added a test with the validator. I think it should be good now!

PR #1373, opened 2024-03-25
API to access protected attributes

pipiche38:
Purpose is to provide way to know if a topology scan is in progress and what is the % of completion


pipiche38:
I understand that there is no test coverage for the added function, but I have no clue how and what to do.

pipiche38:
@puddly we would like to integrate and use the neighbors and routing reports in the zigbeefordomoticz plugin. This PR is quiet essential as it allows us to know if there is a scan in progress and how much %progress the scan is.

thanks for facilitating the merge and get it in the next zigpy package release

zigpy/topology.py
```diff
@@ -166,6 +174,7 @@ async def _scan(
             LOGGER.debug(
                 "Scanning topology (%d/%d) of %s", index + 1, len(devices), device
             )
+            self.scan_progress = index / len(devices)
```
puddly, line 177: If you are subscribing to scan progress, it would be better to emit an event (`self.listener_event("scan_progress", index, len(devices))`) instead of computing the percentage and storing it as an attribute. This should also make it easier to test.

PR #1381, opened 2024-04-12
Merge `in_clusters` and `out_clusters` in the database

puddly:
While working on OTA startup caching issues, I noticed that our database format only really "operates" on `in_clusters` and ignores `out_clusters` for the attribute cache and the unsupported attribute table. This feels like a bug, as this isn't mirrored in `zcl/__init__.py`. To solve this, I am merging the two cluster tables into one and allowing output/client cluster attributes to be cached and marked unsupported.

This requires a database migration so I've also renamed `attrid` to `attr_id` in the columns to clean things up.

### DIFF
--- a/zigpy/endpoint.py
+++ b/zigpy/endpoint.py
@@ -135,6 +135,13 @@ class Endpoint(zigpy.util.LocalLogMixin, zigpy.util.ListenableMixin):
             cluster = zigpy.zcl.Cluster.from_id(self, cluster_id, is_server=False)
 
         self.out_clusters[cluster_id] = cluster
+
+        if self._device.application._dblistener is not None:
+            listener = zigpy.zcl.ClusterPersistingListener(
+                self._device.application._dblistener, cluster
+            )
+            cluster.add_listener(listener)
+
         return cluster
 
     async def add_to_group(self, grp_id: int, name: str | None = None) -> ZCLStatus:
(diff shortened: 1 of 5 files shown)


TheJulianJES:
Oh nice! This should also allow us to remove this workaround in HA:
- https://github.com/home-assistant/core/pull/90749
- https://github.com/home-assistant/core/blob/e3680044fedc253ec00649103fad5fe34ae4f72a/homeassistant/components/zha/binary_sensor.py#L158-L167

zigpy/endpoint.py
```diff
@@ -90,6 +90,11 @@ async def initialize(self) -> None:
 
         self.status = Status.ZDO_INIT
 
+    @property
+    def clusters(self) -> list[zigpy.zcl.Cluster]:
+        """Return all clusters on this endpoint."""
+        return list(self.in_clusters.values()) + list(self.out_clusters.values())
```
TheJulianJES: Might not matter a lot, but since this is called every time the attribute cache is saved, this might be a bit faster(?)
```suggestion
        return [*self.in_clusters.values(), *self.out_clusters.values()]
```

puddly: `_save_attribute_cache` is called when a device object is initially committed into the database, after initialization. Attribute updates go through `_save_attribute`, which shouldn't do anything too expensive. But I do like this syntax, I didn't know you could use it for lists!

TheJulianJES: > `_save_attribute_cache` is called when a device object is initially committed into the database, after initialization. Attribute updates go through `_save_attribute`

Ah, of course! 😄

PR #1386, opened 2024-04-17
Inovelli - Update Firmware URL For ZHA New Update Method

InovelliUSA:


puddly:
Thanks! Update this line as well, as we run a mock server during unit testing:

https://github.com/zigpy/zigpy/blob/18a7add71a546ba009e811c153a61c60d7f0987a/tests/ota/test_ota_providers.py#L368

What is the difference between the images in the old URL and the new one?

InovelliUSA:
We are going to leave the old url in place for users that don't update to new versions of HA/ZHA. The new URL will be for users that keep HA/ZHA up to date. They will get exposed to new firmware updates (because we like the new way that OTA is being handled). :)

PR #1388, opened 2024-04-24
Get list of Third Reality manufacturer IDs via url link

hwzolin:
Change-Id: I2ff2af50c65600d84b18d377f16666b23b60222c

Hello, considering that our company has many chip manufacturers, the manufacturer ID may change later. Since it is inconvenient to resubmit the code and wait for review every time the manufacturer ID is changed, we decided not to hard-code the list of manufacturer IDs, but to obtain it through a URL link.

### DIFF
--- a/zigpy/ota/providers.py
+++ b/zigpy/ota/providers.py
@@ -540,7 +540,7 @@ class Inovelli(BaseOtaProvider):
 
 
 class ThirdReality(BaseOtaProvider):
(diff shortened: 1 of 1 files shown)


puddly:
How many IDs do you expect to have for it to be dynamic?

The purpose of the manufacturer ID filter is to avoid sending requests unless the devices exist. Downloading the list of IDs on startup like this isn't appropriate, since it's sending a request.

PR #1389, opened 2024-04-24
Modify manufacturer IDs of Third Reality devices

hwzolin:
Hello, according to your request, I have confirmed the new manufacturer IDs. Please review them. Thank you.

### DIFF
--- a/zigpy/ota/providers.py
+++ b/zigpy/ota/providers.py
@@ -540,7 +540,7 @@ class Inovelli(BaseOtaProvider):
 
 
 class ThirdReality(BaseOtaProvider):
-    MANUFACTURER_IDS = []
+    MANUFACTURER_IDS = [4659,5127]
 
     JSON_SCHEMA = {
         "type": "object",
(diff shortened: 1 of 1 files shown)


zigpy/ota/providers.py
```diff
@@ -540,7 +540,7 @@ async def _load_index(
 
 
 class ThirdReality(BaseOtaProvider):
-    MANUFACTURER_IDS = [4659, 4877]
+    MANUFACTURER_IDS = [4659,5127]
```
TheJulianJES: Was the manufacturer id `4877` not used before?
Also, add the space back.

zigpy/ota/providers.py
```diff
@@ -593,6 +593,7 @@ async def _load_index(
             )
 
 
+
```
TheJulianJES: Can you remove this added space again?

zigpy/ota/providers.py
hwzolin: At the request of company leaders, future manufacturer IDs will be 4659 and 5127.

TheJulianJES: Ok, I see that. Just wanted to confirm that no (OTA capable) devices are currently using `4877`?

zigpy/ota/providers.py
hwzolin: OK

TheJulianJES: Ok, now you're removing an empty line instead of adding an additional one. Please make sure the diff is the same for these empty lines.

zigpy/ota/providers.py
hwzolin: OK, thanks for the reminder, after communicating again, we decided to keep the current manufacturer IDs and add a new manufacturer ID just in case. I've made the changes in the latest commit.

PR #1397, opened 2024-05-06
Pass all compatible firmware metadata when receiving an OTA upgrade request

### LINKED ISSUE #1382: OTA does not distinguish between up-to-date devices and devices without updates
Our OTA handling doesn't distinguish between an update not being available due to the device being already up-to-date and an update being unavailable because no image is compatible. This is because we filter out "incompatible" OTA updates during the selection process.

One possible solution is to return the latest firmware image as opposed to `None`. This has one major downside, as we are as lazy as possible when sending HTTP requests but also as safe as possible when validating metadata. We can't be sure that an OTA image is compatible with a device until we download the metadata index *and* the entire OTA file. This is wasteful: we would download a full OTA image for every compatible device just to indicate that it is up-to-date.

Another possibility is to mark OTA images with metadata as "complete" or "incomplete", allowing us to skip downloading the image unless necessary. We can also cache firmware images to disk (keyed by their metadata), allowing us to implement the above version scheme without extraneous HTTP traffic.

puddly:
Fixes #1382.

This PR replaces `get_ota_image` with `get_ota_images`, which returns all compatible OTA image metadata. For upgrades, this metadata will also contain the full firmware image. For downgrades, it will contain just metadata.

This allows applications to infer the state of the device with much more granularity:

1. If there are upgrades, the first one is what should be upgraded to.
2. If there are no upgrades but there are downgrades, there are two possibilities:
   1. The most recent downgrade's current firmware version is identical to the version of the device: this means the device is up-to-date.
   2. If the most recent downgrade's firmware version is *lower*, this means that the current device firmware is higher than what is in any publicly available image repository.
3. If there are no upgrades and no downgrades, this device has no compatible OTA images.

### DIFF
--- a/tests/ota/test_ota_matching.py
+++ b/tests/ota/test_ota_matching.py
@@ -147,6 +147,9 @@ async def test_ota_matching_priority(tmp_path: pathlib.Path) -> None:
     ota.register_provider(BrokenProvider(index))
 
     images1 = await ota.get_ota_images(device, query_cmd)
+
+    # The image that will be chosen is the correct one, others with less specificity
+    # will still be present but they will be deprioritized
     assert images1.upgrades[0] == zigpy.ota.OtaImageWithMetadata(
         metadata=index[3],
         firmware=zigpy.ota.image.OTAImage.deserialize(index[3].test_data)[0],
--- a/tests/ota/test_ota_matching.py
+++ b/tests/ota/test_ota_matching.py
@@ -316,11 +316,11 @@ async def test_ota_concurrent_fetching() -> None:
     with patch.object(
         provider, "_load_index", wraps=provider._load_index
     ) as load_index:
-        image1, image2 = await asyncio.gather(
-            ota.get_ota_image(device, query_cmd),
-            ota.get_ota_image(device, query_cmd),
+        images1, images2 = await asyncio.gather(
+            ota.get_ota_images(device, query_cmd),
+            ota.get_ota_images(device, query_cmd),
         )
 
     # Concurrent requests were combined
     assert len(load_index.mock_calls) == 1
-    assert image1 is image2
+    assert images1 == images2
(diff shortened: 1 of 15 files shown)


tests/ota/test_ota_matching.py
```diff
@@ -146,16 +146,14 @@ async def test_ota_matching_priority(tmp_path: pathlib.Path) -> None:
     ota.register_provider(SelfContainedProvider(index))
     ota.register_provider(BrokenProvider(index))
 
-    image1 = await ota.get_ota_image(device, query_cmd)
-    assert image1 is not None
-    assert image1 == zigpy.ota.OtaImageWithMetadata(
+    images1 = await ota.get_ota_images(device, query_cmd)
```
TheJulianJES, line 149: Would it make sense to also assert the length of `images1.upgrades` == 1 here to make sure we don't get any other unwanted images later in the list?

E.g. (second line):
```suggestion
    images1 = await ota.get_ota_images(device, query_cmd)
    assert len(images1.upgrades) == 1
```

zigpy/ota/providers.py
```diff
@@ -59,6 +59,7 @@ class BaseOtaImageMetadata(t.BaseDataclassMixin):
     max_current_file_version: int | None = None
     specificity: int | None = None
 
+    complete: bool = False
```
TheJulianJES: This is not used yet, right?

TheJulianJES (approved):
Looks good!

zigpy/ota/providers.py
puddly: Oops. This was supposed to denote if the metadata for a provider is 100% complete, meaning that the image file doesn't need to be downloaded. It's not used.

tests/ota/test_ota_matching.py
puddly: With the way OTA is structured, four images in this test will match, just one of them has a higher specificity given that it has more metadata to match against.

Issue #1410, opened 2024-06-09
MultipleQuirksMatchException when reloading ZHA integration with custom quirks v2


vinzent:
I'm on HomeAssistant Green with OS 12.3, HA 2024.6.1 and SkyConnect.

I've created 2 quirks v2:

``` 
find /config/custom_zha_quirks/
/config/custom_zha_quirks/
/config/custom_zha_quirks/__pycache__
/config/custom_zha_quirks/__pycache__/zg-204zs.cpython-312.pyc
/config/custom_zha_quirks/__pycache__/zg-204zm.cpython-312.pyc
/config/custom_zha_quirks/zg-204zs.py
/config/custom_zha_quirks/zg-204zm.py
```

Source: [zg-204zs.py](https://gist.github.com/vinzent/2194a8ef531d8e8ec3bc17ea17eab0e9), [zg-204zm](https://gist.github.com/vinzent/2cd645b848fd3b6a0c3e5762956ec89f).

After adding one of the custom quirks, I can't reload the ZHA integration anymore. More or less instantly after reloading (ZHA integration -> 3dots -> Reload) it fails:

![grafik](https://github.com/zigpy/zigpy/assets/296297/a75c709c-76fe-418d-851e-7138c91902fd)

From `/homeassistant/home-assistant.log`:

```
2024-06-09 18:33:47.327 WARNING (ImportExecutor_0) [zhaquirks] Loaded custom quirks. Please contribute them to https://github.com/zigpy/zha-device-handlers
2024-06-09 18:33:47.504 ERROR (MainThread) [homeassistant.config_entries] Error setting up entry Home Assistant SkyConnect for zha
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/config_entries.py", line 594, in async_setup
    result = await component.async_setup_entry(hass, self)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/components/zha/__init__.py", line 134, in async_setup_entry
    async with radio_mgr.connect_zigpy_app() as app:
  File "/usr/local/lib/python3.12/contextlib.py", line 210, in __aenter__
    return await anext(self.gen)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/components/zha/radio_manager.py", line 182, in connect_zigpy_app
    app = await self.radio_type.controller.new(
[...]

PR #1415, opened 2024-06-27
QuirksRegistryV2: BugFix MultipleQuirksMatchException when reloading ZHA integration

cemizm:
solves MultipleQuirksMatchException when reloading zha integration. for further details see issue #1410

### DIFF
--- a/tests/test_quirks_v2.py
+++ b/tests/test_quirks_v2.py
@@ -14,7 +14,6 @@ from zigpy.const import (
     SIG_MODELS_INFO,
 )
 from zigpy.device import Device
-from zigpy.exceptions import MultipleQuirksMatchException
 from zigpy.profiles import zha
 from zigpy.quirks import CustomCluster, CustomDevice, signature_matches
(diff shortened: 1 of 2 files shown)


cemizm:
like i mentioned in the linked issue, this is just a quick fix in my opinion. it should be considered to use a flat key, value pair collection instead of allowing the registration of multiple entries for the same manufacturer, model:

https://github.com/zigpy/zigpy/blob/dev/zigpy/quirks/registry.py#L36

maybe there is a good reason for the collection, but that would make the handling of the quirks much easier and render the multiple quirk exception obsolete.

dmulcahey:
I think we need to prevent the double import of the code itself. We are aware of this and I am actively looking at a permanent fix. Just have a couple more pressing items atm.

cemizm:
> I think we need to prevent the double import of the code itself. We are aware of this and I am actively looking at a permanent fix. Just have a couple more pressing items atm.

that sounds great. can we at least use this as a workaround until the permanent fix is available?

PR #1422, opened 2024-07-13
Do not open serial ports exclusively by default

puddly:
Partially reverts https://github.com/zigpy/zigpy/pull/1418.


puddly:
Looks like this is not enough! Needs more investigation 🤔

puddly:
`exclusive` is blocked by https://github.com/home-assistant-libs/pyserial-asyncio-fast/pull/15.

PR #1425, opened 2024-07-14
Clean serial port shutdown

puddly:
There is currently a connection lifecycle bug[^1] with pyserial-asyncio(-fast) that causes a race condition on shutdown. Various radio libraries and other implementations have workarounds for this issue (usually `await asyncio.sleep(0)` or similar) but the crux of the problem is that we do not *wait* for the serial connection to close. There are two events:

1. `connection_made`
2. `connection_lost`

We currently wait for `connection_made` but react synchronously when shutting down a transport, assuming `transport.close()` shuts down the serial connection in the same event loop iteration. This is not true.

To help solve this issue, I would like to incorporate a serial protocol base class into zigpy so that radio libraries can have a more coordinated implementation and not have to re-implement basic functionality repeatedly.

This will unfortunately have to be an API change: `disconnect()` in the various radio libraries does not actually wait for the port to disconnect, nor do failures during `connect()`, and so on.

Radio library progress:
- https://github.com/zigpy/zigpy-deconz/pull/259
- https://github.com/NabuCasa/universal-silabs-flasher/pull/75
- https://github.com/zigpy/zigpy-zigate/pull/160
- https://github.com/zigpy/zigpy-znp/pull/256
- https://github.com/zigpy/zigpy-xbee/pull/177

[^1]: https://github.com/home-assistant-libs/pyserial-asyncio-fast/pull/16

### DIFF
--- a/tests/test_serial.py
+++ b/tests/test_serial.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import pathlib
-import unittest.mock
+from unittest.mock import AsyncMock, Mock, call, patch
 
 import pytest
 
@@ -39,7 +39,7 @@ async def test_serial_normal(
     expected_kwargs: dict[str, bool],
 ) -> None:
     loop = asyncio.get_running_loop()
-    protocol_factory = unittest.mock.Mock()
+    protocol_factory = Mock()
 
     kwargs = {"url": url}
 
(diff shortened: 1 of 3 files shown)


puddly:
I've tested this change somewhat extensively with the flasher and with the radio libraries as I convert them.

From the ZHA side nothing needs to change: this is rolled into `await app.disconnect()`.

PR #1430, opened 2024-07-26
Block updates for IKEA image types 8704 and 8710

### LINKED ISSUE #1428: Ikea Tradifi Update Brick the light
Zha update brick two new bulb i buy recently due update

The bulbs (shipped firmware)

TRADFRI bulb E27 WS globe 1055lm
Firmware: 0x03000010

Update to: Firmware: 0x03000022

I have old ones (that update seems not appears)

TRADFRIbulbE27WSglobeopal1055lm
Firmware: 0x00010012

The result is the same that someone posted in this video

https://www.reddit.com/r/tradfri/comments/1e0ohmc/why_are_my_lamps_in_disco_mode/

Seems a known manufacture error:

https://ww8.ikea.com/ikeahomesmart/releasenotes/releasenotes.html

as i understand they fixed in 22 no? but no, 22 break that new bulbs

I do not known how ZHA deal with updates but probably will be good if block  updates to this firmware.

puddly:
Fixes #1428.

We'll re-enable them whenever IKEA fixes this problem. I'd rather not take the risk of figuring out which exact devices are affected.

### DIFF
--- a/tests/ota/test_ota_providers.py
+++ b/tests/ota/test_ota_providers.py
@@ -279,7 +279,7 @@ async def test_tradfri_provider_old(index_url: str, index_file: str) -> None:
         if obj["fw_type"] == 2 and obj["fw_image_type"] not in (8710, 8704)
     ]
     assert index
-    assert len(index) == len(index_obj) - 2 == len(filtered_version_info_obj)
+    assert len(index) == len(filtered_version_info_obj)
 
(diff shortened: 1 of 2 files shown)


puddly:
From what I've seen, this will only affect people who have manually overridden the IKEA feed URL in ZHA's config. Merely enabling OTA for IKEA will use the DIRIGERA feed, which doesn't seem to upgrade to an affected version.

PR #1431, opened 2024-08-02
Handle OTA image page requests

puddly:
First time I've seen this, it's with a ThirdReality night light. Instead of sending an image block request, the device sends image *page* requests:

image

Since image page requests explicitly *disable* ACKs, have no default response, and are optional, the fallback behavior is pretty terrible. The device waits about 10s after sending an image page request and then sends a normal image block request.

This PR implements support for this command and speeds up OTA by a significant amount for devices that require it.

The odd command sending structure is because we do not expose a simple way to send specific commands with specific header flags. This needs to be addressed in a followup PR.

### DIFF
--- a/tests/ota/test_ota_manager.py
+++ b/tests/ota/test_ota_manager.py
@@ -1,13 +1,22 @@
-from unittest.mock import AsyncMock, patch
+from unittest.mock import AsyncMock, MagicMock, call, patch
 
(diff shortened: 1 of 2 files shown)


RCapolungo:
Are there any updates for the merge of this PR?

puddly:
Tested successfully with the same Third Reality night light.

PR #1441, opened 2024-08-11
Add new quirks v2 metadata type that allows replacing all instances of a cluster on a device

dmulcahey:
This PR adds a new metadata type and a new builder method to allow replacing all instances of a cluster without having to add replaces for each occurrence. There are times where depending on the firmware version of device variant where a cluster may need to be replaced across all endpoints but the device signature isn't consistent. The existing replaces ensures the specified endpoint exists and is endpoint specific.

### DIFF
--- a/zigpy/quirks/v2/__init__.py
+++ b/zigpy/quirks/v2/__init__.py
@@ -83,6 +83,11 @@ class CustomDeviceV2(CustomDevice):
         for replace_meta in quirk_metadata.replaces_metadata:
             replace_meta(self)
 
+        for (
+            replace_occurrences_meta
+        ) in quirk_metadata.replaces_cluster_occurrences_metadata:
+            replace_occurrences_meta(self)
+
         for entity_meta in quirk_metadata.entity_metadata:
             entity_meta(self)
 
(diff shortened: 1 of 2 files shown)


zigpy/quirks/v2/__init__.py
```diff
@@ -539,6 +580,37 @@ def replaces(
...
+
+        This method allows replacing a cluster on a device across all endpoints
+        for the specified cluster types when the quirk is applied.
+
+        replacement_cluster_class should be a subclass of Cluster or CustomCluster and
+        will be used to create a new cluster instance to replace the existing cluster.
+
+        If cluster_id is provided, it will be used as the cluster_id for the cluster to
+        be removed. If cluster_id is not provided, the cluster_id of the replacement
+        cluster will be used.
```
TheJulianJES: `cluster_id` isn't a parameter for this. The text is copied from above, I think we can just remove this part? Don't see a use-case for this here at least.

```suggestion
```

dmulcahey: copy pasta error

TheJulianJES (approved):
Nice, looks good!

PR #1449, opened 2024-08-14
[RFC] metadata with class references cause issues when ZHA is reloaded.

dmulcahey:
This is a continuation of #1410 I think the issue is that any metadata classes that contain references to locally defined classes in custom quirks will not match after reloading quirks because the class definitions have different hashes. 

Additional feedback was given here: https://github.com/zigpy/zigpy/issues/1410#issuecomment-2287996706

This PR contains a test that reproduces the issue and a potential solution.


dmulcahey:
closing for a different solution

PR #1453, opened 2024-08-15
Expose options bitmasks for Color and LevelControl cluster

puddly:
Allows for smooth color transitions for Hue lights from an "off" state with the following ZHA patch:

```diff
diff --git a/zha/application/platforms/light/__init__.py b/zha/application/platforms/light/__init__.py
index 87c6034..1a29ea4 100644
--- a/zha/application/platforms/light/__init__.py
+++ b/zha/application/platforms/light/__init__.py
@@ -580,6 +580,8 @@ class BaseLight(BaseEntity, ABC):
                     enhanced_hue=int(hs_color[0] * 65535 / 360),
                     saturation=int(hs_color[1] * 2.54),
                     transition_time=int(10 * transition_time),
+                    options_mask=Color.OptionsMask.Execute_if_off_present,
+                    options_override=Color.Options.Execute_if_off,
                 )
                 t_log["enhanced_move_to_hue_and_saturation"] = result
             else:
@@ -602,6 +604,8 @@ class BaseLight(BaseEntity, ABC):
                 color_x=int(xy_color[0] * 65535),
                 color_y=int(xy_color[1] * 65535),
                 transition_time=int(10 * transition_time),
+                options_mask=Color.OptionsMask.Execute_if_off_present,
+                options_override=Color.Options.Execute_if_off,
             )
             t_log["move_to_color"] = result
             if result[1] is not Status.SUCCESS:
diff --git a/zha/zigbee/cluster_handlers/lighting.py b/zha/zigbee/cluster_handlers/lighting.py
index e9f034f..d404f6a 100644
--- a/zha/zigbee/cluster_handlers/lighting.py
+++ b/zha/zigbee/cluster_handlers/lighting.py
@@ -197,4 +197,4 @@ class ColorClusterHandler(ClusterHandler):
     @property
     def execute_if_off_supported(self) -> bool:
         """Return True if the cluster handler can execute commands when off."""
-        return Color.Options.Execute_if_off in self.options
+        return self.cluster.get(Color.AttributeDefs.options.name) is not None
```

The `execute_if_off_supported` patch is something that needs to be explored in ZHA. I think what we have now is not accurate: the ZCL expects the `options` attribute to be *writeable* in addition to readable. It'll be set to `0` on a newly joined light.

### DIFF
--- a/zigpy/zcl/clusters/general.py
+++ b/zigpy/zcl/clusters/general.py
@@ -978,6 +978,11 @@ class StepMode(t.enum8):
     Down = 0x01
 
 
+class OptionsMask(t.bitmap8):
+    Execute_if_off_present = 0b00000001
+    Couple_color_temp_to_level_present = 0b00000010
+
+
 class Options(t.bitmap8):
     Execute_if_off = 0b00000001
     Couple_color_temp_to_level = 0b00000010
@@ -991,6 +996,7 @@ class LevelControl(Cluster):
     MoveMode: Final = MoveMode
     StepMode: Final = StepMode
     Options: Final = Options
+    OptionsMask: Final = OptionsMask
 
     cluster_id: Final[t.uint16_t] = 0x0008
     name: Final = "Level control"
@@ -1034,8 +1040,8 @@ class LevelControl(Cluster):
             schema={
                 "level": t.uint8_t,
                 "transition_time": t.uint16_t,
-                "options_mask?": t.bitmap8,
-                "options_override?": t.bitmap8,
+                "options_mask?": OptionsMask,
+                "options_override?": Options,
             },
             direction=Direction.Client_to_Server,
         )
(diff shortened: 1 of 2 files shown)


TheJulianJES:
> I think what we have now is not accurate: the ZCL expects the options attribute to be writeable in addition to readable. It'll be set to 0 on a newly joined light.

What's inaccurate about what we have at the moment? To have the smooth transition from an off-state currently, the `options` attribute needs to be changed using the clusters UI (or a service call). Then, the light will always accept color commands when it's off (and ZHA will change to make use of that behavior and send the color commands first).

There was a PR adding a config entity for this (to every supporting light), but it's not been merged (correctly IMO, as having a config entity for this setting and for every light is not ideal):
- https://github.com/home-assistant/core/pull/86197

The issue with setting that attribute on join/configuration would be that some lights seem to break when `ExecuteIfOff` is enabled. I think I have a Tuya LED controller that behaved really weirdly with that option on, worse than with it being off.
I can check that later though.

I do think the approach that could be built with this PR is way better (instead of writing the attribute), but it's possible that we cannot enable this for all lights in ZHA (without breaking some).

TheJulianJES (approved):
These changes look good to me!

puddly:
> To have the smooth transition from an off-state currently, the options attribute needs to be changed using the clusters UI (or a service call). 

Ah, I was missing this part. I didn't see any code to write it during joining so I assumed that it was assuming the `options` attribute would be set by default.

PR #1456, opened 2024-08-16
Use zigpy serialization for attributes reads and writes in the database

puddly:
(WIP)

This PR splits up database loading into two halves: a mostly stateless phase, where devices are loaded and populated as much as possible within `appdb` alone, and then a part to mutate `ControllerApplication`. This is necessary because we need to load quirks completely when figuring out which attributes exist and which don't.

Bad attributes are logged and dropped during migration:

```
2024-08-16 11:51:40.386 WARNING (MainThread) [zigpy.appdb] [0x57fb:1:0xff01] Attribute 0x0001 with serialized value 1 does not exist, ignoring
2024-08-16 11:51:40.376 WARNING (MainThread) [zigpy.appdb] [0xe9e7:1:0xff01] Attribute 0x0080 with serialized value 16908288 does not exist, ignoring
...
```

This one is a bug in zigpy, we're missing the `reporting_status` attribute on a bunch of clusters:

```
2024-08-16 11:51:40.371 WARNING (MainThread) [zigpy.appdb] [0x0217:1:0x0702] Attribute 0xFFFD with serialized value 1 does not exist, ignoring
```

This one is a bug in quirks:

```
2024-08-16 11:51:40.378 WARNING (MainThread) [zigpy.appdb] [0xc7c8:1:0xfc31] Failed to deserialize attribute periodic_power_and_energy_reports (<class 'zigpy.types.basic.uint8_t'>) = 3600: ValueError('3600 is not an unsigned 8 bit integer')
```

Warnings will be logged when an unknown attribute is cached.

### DIFF
--- a/zigpy/appdb.py
+++ b/zigpy/appdb.py
@@ -29,7 +29,7 @@ from zigpy.zdo import types as zdo_t
 
 LOGGER = logging.getLogger(__name__)
 
-DB_VERSION = 13
+DB_VERSION = 14
 DB_V = f"_v{DB_VERSION}"
 MIN_SQLITE_VERSION = (3, 24, 0)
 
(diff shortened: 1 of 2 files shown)


puddly:
Let's keep this open.

PR #1463, opened 2024-08-21
Allow specifying attribute ZCL types independently from the Python type

### LINKED ISSUE #948: How to wrap a `t.Bool` attribute?
During a device quirk (https://github.com/zigpy/zha-device-handlers/issues/1442) we thought it would be a good idea to wrap one of the attributes of type `t.Bool` with a class that gives information about the True/False values meaning in a similar way to attributes that accept values. `t.enum`.
This way, the value shown in HA is more informative to the user as it has a functional meaning of the value.
Some examples would be:
https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/tuya/__init__.py#L831-L846

But for values of type `t.Bool` this can not be done. This:
```python
class ChildLock(t.Bool):
    """Tuya child lock states."""

    LOCK = True  # maybe 1 or 0x01
    UNLOCK = False  # maybe 0 or 0x00
```

will not work because of:
```
TypeError: ChildLock: cannot extend enumeration 'Bool'
```

A `t.enum8` can be used like this:
```python
class ChildLock(t.enum8):
    """Tuya child lock states."""

    LOCK = t.Bool.true
    UNLOCK = t.Bool.false
```

but then values can not be setted via HA UI.

Other combinations with 0/1 or True/False values will raise errors `Status.INVALID_DATA_TYPE`:
```
2022-03-24 11:38:32 DEBUG (MainThread) [homeassistant.components.zha.api] Read attribute for: cluster_id: [6] cluster_type: [in] endpoint_id: [1] attribute: [32768] manufacturer: [None] response: [0] failure: [{}],
2022-03-24 11:38:38 DEBUG (MainThread) [homeassistant.core] Bus:Handling <Event call_service[L]: domain=zha, service=set_zigbee_cluster_attribute, service_data=ieee=a4:c1:38:e0:b5:a5:bb:c6, endpoint_id=1, cluster_id=6, cluster_type=in, attribute=32768, value=True>
2022-03-24 11:38:38 DEBUG (MainThread) [homeassistant.components.zha.core.device] [0x574F](TS011F): set: True for attr: 32768 to cluster: 6 for ept: 1 - res: [[WriteAttributesStatusRecord(status=<Status.INVALID_DATA_TYPE: 141>, attrid=32768)]]
2022-03-24 11:38:38 DEBUG (MainThread) [homeassistant.components.zha.api] Set attribute for: cluster_id: [6] cluster_type: [in] endpoint_id: [1] attribute: [32768] value: [True] manufacturer: [None] response: [[[WriteAttributesStatusRecord(status=<Status.INVALID_DATA_TYPE: 141>, attrid=32768)]]]
```

So the question would be: is there a way to wrap a `t.Bool` attribute in a helper class?

I think it's a pretty common situation, but I haven't found any implementation or example of how to achieve it.

The goal would be to give more meaningful values to true/false attributes. 
Some other examples:
https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/tuya/__init__.py#L1044-L1050

puddly:
I've removed `DATA_TYPES` and replaced it with a structured `DataType` class. It's used in exactly one place outside of zigpy so this isn't much of a breaking change: https://github.com/zigpy/zha-device-handlers/blob/156aaa6bb704a13fe9500b833c293405fc5ffd79/tests/test_xiaomi.py#L630-L632

The main motivation here is to decouple the ZCL data type (`foundation.DataTypeId`) from the Python type of an attribute:

```python
class DeviceMode(t.enum8):
    Off = 0x00
    On = 0x01
    Something_Else = 0x02


class AttributeDefs(BaseAttributeDefs):
    device_mode: Final = ZCLAttributeDef(
        id=0x1234,
        zcl_type=foundation.DataTypeId.uint8,  # <-- you can now specify this
        type=DeviceMode,
        access="rw",
        manufacturer_specific=True,
    )
```

By default, the ZCL type is inferred from the Python type as before. This just allows you to override it. This is a somewhat "unsafe" operation because it will allow you to change the type of the written attribute, zigpy assumes the Python type serializes identically to the original type.

Fixes https://github.com/zigpy/zigpy/issues/948


requirements_test.txt
```diff
@@ -10,5 +10,6 @@ pytest-timeout
 freezegun
 tomli
 aioresponses
+backports.zoneinfo; python_version<"3.9"
```
TheJulianJES: This is only needed (even after https://github.com/zigpy/zigpy/pull/1464), because we're still running GH CI for 3.8, right?

puddly: Ooo, good catch. This is due to a bad merge.

tests/test_zcl_clusters.py
```diff
@@ -1,11 +1,21 @@
...
+from datetime import datetime, timezone
 import re
+from typing import Any
 
 import pytest
 
+try:
+    from zoneinfo import ZoneInfo
+except ImportError:
+    from backports.zoneinfo import ZoneInfo  # type: ignore[no-redef]
```
TheJulianJES, line 13: I guess we can also remove this then?
Although this might break 3.8 GH CI(?)

puddly: Ah, more from a bad merge. Good point. We can drop 3.8 from the main CI.

x86fr:
Unwanted, but it seems this change broke ZHA Toolkit : https://github.com/mdeweerd/zha-toolkit/issues/261

puddly:
@x86fr There's nothing that can be done about it, unfortunately. ZHA Toolkit doesn't specify that it depends on any specific version of zigpy.

x86fr:
Yes. The fix appears to be straightforward (only 3 lines to change), but I fear it may affect many users. Will test a patch and try to push a PR at ZHA Toolkit.

dmulcahey:
> Yes. The fix appears to be straightforward (only 3 lines to change), but I fear it may affect many users. Will test a patch and try to push a PR at ZHA Toolkit.

Sucks that that’s the case. We don’t use closed PR’s for discussion nor do we check the internals of custom components when we make changes. This is an inherent risk that folks take when using custom components in general and more so when using custom components that import the internals of other libs directly 🤷🏻‍♂️

PR #1464, opened 2024-08-26
Drop Python 3.8 support

puddly:
Python 3.8 is receiving only security updates. That will end next month.

@pipiche38 Domoticz seems to be the only user of zigpy on older Pythons. What version are you actually using right now? There are many language features that we miss with every older release we support so if you don't use 3.9, 3.10, or 3.11, we can drop them too.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.5.2
+    rev: v0.6.2
     hooks:
       - id: ruff
         args: ["--fix", "--exit-non-zero-on-fix", "--config", "pyproject.toml"]
@@ -15,6 +15,6 @@ repos:
         args: ["--toml", "pyproject.toml"]
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.10.1
+    rev: v1.11.2
     hooks:
       - id: mypy
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ authors = [
 ]
 readme = "README.md"
 license = {text = "GPL-3.0"}
-requires-python = ">=3.8"
+requires-python = ">=3.9"
 dependencies = [
     "attrs",
     "aiohttp",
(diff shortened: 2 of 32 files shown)


pipiche38:
> Python 3.8 is receiving only security updates. That will end next month.
> 
> @pipiche38 Domoticz seems to be the only user of zigpy on older Pythons. What version are you actually using right now? There are many language features that we miss with every older release we support so if you don't use 3.9, 3.10, or 3.11, we can drop them too.

I'll have a look to the user community.
Myself, I'm using 3.11, 3.10 and 3.9

pipiche38:
@puddly I do confirm, most of the users are on 3.9 or above. No issue on our side for 3.8 to be removed, but if you remove 3.9, 3.10 then we will have an issue

Hedda:
> There are many language features that we miss with every older release we support so if you don't use 3.9, 3.10, or 3.11, we can drop them too.

Related questions about two other dependencies in https://github.com/zigpy/zigpy/blob/dev/pyproject.toml

- Does 'importlib_resources' still require ' python_version<"3.9"' (as in Python 3.8 or lower/older)
   - https://pypi.org/project/importlib-resources/
     - https://github.com/python/importlib_resources 
- Does 'async-timeout' still require python_version<"3.11"'' (as in Python 3.11 or lower/older)
   - https://pypi.org/project/async-timeout/
     - https://github.com/aio-libs/async-timeout

So, can and should one/both of those two also use a newer python version too? Latest async-timeout look only to have support for Python 3.11 and importlib-resources look to at least support Python 3.11 as well (but not perfectly clear)

* https://github.com/python/importlib_resources/blob/main/pyproject.toml
  * https://github.com/python/importlib_resources/blob/main/.github/workflows/main.yml

* https://github.com/aio-libs/async-timeout/blob/master/pyproject.toml

PS: By the way, the async-timeout project seems like it might be abandoned?

puddly:
@pipiche38 It would be best to push your users to the highest version of Python possible. I'll set a tentative date of six months for 3.9.

3.9 is EoL and will stop receiving *all* updates (including security) next year so you ideally would want to be on 3.10 or 3.11. Domoticz appears to support down to Python 3.4 (!) which has been abandoned for over five years now!

pipiche38:
@puddly I understand your Point of View, but I'm not the Domoticz developer. I think what needs to be taken in consideration is most of the Domoticz users are running on Raspberry Pi with raspbian and consequently are not at the edge in terms of software version.

PR #1467, opened 2024-09-09
Remove quirks v2 device class validation

GitforZhangXL:
I'm not sure if the modification will affect the other, but it works for me
Used with https://github.com/zigpy/zha/pull/199

### DIFF
--- a/zigpy/quirks/v2/__init__.py
+++ b/zigpy/quirks/v2/__init__.py
@@ -256,18 +256,18 @@ class EntityMetadata:
 
     def _validate(self) -> None:
         """Validate the entity metadata."""
-        has_unit: bool = hasattr(self, "unit") and getattr(self, "unit") is not None
-        has_device_class: bool = hasattr(self, "device_class") and (
-            getattr(self, "device_class") is not None
-        )
-        if has_device_class and has_unit:
-            raise ValueError(
-                f"EntityMetadata cannot have both unit and device_class: {self}"
-            )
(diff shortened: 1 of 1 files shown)


TheJulianJES (changes requested):
Whilst I agree that both of these validations are incorrect, we should clean this up properly if we don't want to programmatically validate this—or do it differently.
Just commenting out the existing code isn't enough. We need to think about how we want to move forward with this (and check how e.g. ESPHome deals with this).

Also, I'm not sure if this is still the case, but there was a time where this validation was duplicated across ZHA and zigpy. We should make sure that's also addressed.

TheJulianJES:
The good thing is that we already have "proper" validation for quirks v2 provided entities in ZHA:
- https://github.com/zigpy/zha/blob/c99237fd8f6145560e9a697288c57a7b28a681af/zha/application/platforms/helpers.py#L81-L101
- https://github.com/zigpy/zha/blob/c99237fd8f6145560e9a697288c57a7b28a681af/zha/units.py#L165-L167

GitforZhangXL:
> * https://github.com/zigpy/zha/blob/c99237fd8f6145560e9a697288c57a7b28a681af/zha/application/platforms/helpers.py#L81-L101
> * https://github.com/zigpy/zha/blob/c99237fd8f6145560e9a697288c57a7b28a681af/zha/units.py#L165-L167

It is these checks that cause the generated sensor entities to lose units

PR #1474, opened 2024-09-19
Expose `fallback_name` in v2 quirks

puddly:
I think we should deprecate creating quirks v2 entities without a translation key or a fallback name. For now, this is logged as a warning but it should throw an error in the future.


dmulcahey:
In a lot of cases the attribute name is the translation key… shouldn’t we incorporate that as well?

puddly:
You mean in the warning?

There aren't many v2 entities right now, I wonder if making it explicit would be simpler?

PR #1479, opened 2024-10-04
Allow renaming devices with quirks

puddly:
This will help us implement `model_id` in Core and deal with white label devices (e.g. eWeLink/Sonoff).

### DIFF
--- a/zigpy/quirks/v2/__init__.py
+++ b/zigpy/quirks/v2/__init__.py
@@ -885,7 +885,7 @@ class QuirkBuilder:
         self.device_automation_triggers_metadata.update(device_automation_triggers)
         return self
 
-    def friendly_name(self, model: str, manufacturer: str) -> QuirkBuilder:
+    def friendly_name(self, *, model: str, manufacturer: str) -> QuirkBuilder:
         """Renames the device."""
         self.friendly_name_metadata = FriendlyNameMetadata(
             model=model, manufacturer=manufacturer
--- a/zigpy/quirks/v2/__init__.py
+++ b/zigpy/quirks/v2/__init__.py
@@ -414,6 +414,7 @@ class QuirkBuilder:
 
         self.registry: DeviceRegistry = registry
         self.manufacturer_model_metadata: list[ManufacturerModelMetadata] = []
+        self.friendly_name_metadata: FriendlyNameMetadata | None = None
         self.filters: list[FilterType] = []
         self.custom_device_class: type[CustomDeviceV2] | None = None
         self.device_node_descriptor: NodeDescriptor | None = None
@@ -886,7 +887,7 @@ class QuirkBuilder:
 
     def friendly_name(self, model: str, manufacturer: str) -> QuirkBuilder:
         """Renames the device."""
-        self.friendly_name = FriendlyNameMetadata(
+        self.friendly_name_metadata = FriendlyNameMetadata(
             model=model, manufacturer=manufacturer
         )
         return self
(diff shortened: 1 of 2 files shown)


zigpy/quirks/v2/__init__.py
```diff
@@ -875,6 +885,13 @@ def device_automation_triggers(
         self.device_automation_triggers_metadata.update(device_automation_triggers)
         return self
 
+    def friendly_name(self, model: str, manufacturer: str) -> QuirkBuilder:
```
TheJulianJES: Is the `model, manufacturer` order intentional?
For quirks, we have `manufacturer, model` as the order for matching them. Do we want to follow this here too?
Or should `model` be in front here because that's how it'll display in the HA frontend: "model by manufacturer"?

TheJulianJES (approved):

zigpy/quirks/v2/__init__.py
puddly: I always thought of it as `model`, `manufacturer` so I think that was unintentional 😄. I feel it would actually be even clearer to *mandate* kwargs:

```suggestion
    def friendly_name(self, *, model: str, manufacturer: str) -> QuirkBuilder:
```

I've seen a few instances of v2 entities being defined and the arguments being listed without kwargs looking a little opaque to me, it would be good to possibly propagate it within quirks too?

TheJulianJES: > I've seen a few instances of v2 entities being defined and the arguments being listed without kwargs looking a little opaque to me

Agreed. I think the initial v2 quirk examples provide `attribute_name` and `cluster_id` as positional arguments.
Everything else should definitely provided as a kwarg, but we could also enforce kwargs for all arguments. That might require some further changes in zigpy/ZHA/zha-quirks, but I can deal with those.

For `friendly_name` here, I'm not sure we should mandate it. `QurkBuilder` and `applies_to` are basically only provided with positional args at the moment. It might get a bit long/repetitive to enforce kwargs there too?
```python
.applies_to(manufacturer="IKEA of sweden", model="TRADFRI remote control")
.applies_to(manufacturer="IKEA of sweden", model="TRADFRI other variant")
.applies_to(manufacturer="IKEA of sweden", model="TRADFRI next variant")
```

I don't have a strong opinion regarding enforcing it for manufacturer/model though. We can also just do it for `friendly_name` to make it clear what's what.

puddly: > It might get a bit long/repetitive to enforce kwargs there too?

I personally think for quirks we should make it so that first-time contributors don't have to guess or be familiar with the code, even if it means a little bit of repetition within the repo as a whole. For `applies_to` it's usually pretty clear once they see one example but for others it may take a bit to figure out what the unnamed args mean.

For `friendly_name` I'll enforce it. We can do others later on in a separate PR if it's useful.

dmulcahey: > > It might get a bit long/repetitive to enforce kwargs there too?
> 
> I personally think for quirks we should make it so that first-time contributors don't have to guess or be familiar with the code, even if it means a little bit of repetition within the repo as a whole. For `applies_to` it's usually pretty clear once they see one example but for others it may take a bit to figure out what the unnamed args mean.
> 
> For `friendly_name` I'll enforce it. We can do others later on in a separate PR if it's useful.

I was on the fence with all kwargs too for a bit but I wholly agree with this now… the repetition is worth the added clarity here IMO.

PR #1481, opened 2024-10-07
Add a `ask_for_ack` kwarg to request methods

puddly:
Unblocks https://github.com/zigpy/zigpy/pull/1431

A side effect of this PR is that all of the request methods internally now support the same args and kwargs. This should help in a future PR to remove unnecessary methods (`reply`, for example) and consolidate everything to use `send_packet` indirectly.


puddly:
Quirks relied on the `call_args` pattern in a few spots so it'll need a few test adjustments: https://github.com/zigpy/zha-device-handlers/pull/3410

The only dubious portion of this PR is <https://github.com/zigpy/zigpy/pull/1481/commits/139fa3bd5c364af948087182a2201f20cee5c38f>. I believe that the only reason we have this is for backwards compatibility with existing unit tests: no real code actually uses these methods like this.

TheJulianJES (approved):

PR #1488, opened 2024-10-21
Require `fallback_name` and add `translation_key` validation

TheJulianJES:
# Proposed change

This PR makes a couple of changes to v2 quirks `EntityMetadata`:
- require `fallback_name` (and validate it)
- require `translation_key` or `device_class` (and validate it)
- adjust tests

The changes are also clearly separated in the commits of the branch/PR for easy review.

# Why?

## `fallback_name`

When we add v2 quirk entities, we need to add translation keys to `strings.json` in Home Assistant.
In order for quirk entities to include a proper (English) name that we can use to add to `strings.json`, we now require `fallback_name`. This will be the default entity name added to `strings.json`.

I already have some rudimentary scripts to automatically add these names when bumping ZHA (with it: zha-quirks) in HA and the test validating translations that I plan to re-add to HA soon (tracking issue: https://github.com/zigpy/zha/issues/72).

## `translation_key`

When we want Home Assistant to use the name of a device class, we want to explicitly set `translation_key=None`.
At the moment, ZHA does not allow for this and also defaults to `attribute_name` (or `command_name`) if the `translation_key` isn't provided. `None` is not accepted at all. ZHA code: [zha/application/platforms/__init__.py#L350-L355](https://github.com/zigpy/zha/blob/2cfd334cf9e3c1fb6337ddfb8062827692e24944/zha/application/platforms/__init__.py#L350-L355)*

IMO, it's best if we just always require `translation_key` and do not default to `attribute_name`.

### Example: Tamper binary sensor

See this example of a tamper binary sensor: [zhaquirks/philips/soc001.py#L48-L56](https://github.com/zigpy/zha-device-handlers/blob/c7887583a83e05ffaed149fd09e3c8a0b1079b0d/zhaquirks/philips/soc001.py#L48-L56)
Here, the `device_class` is set to `TAMPER`. That alone would automatically set a proper name in Home Assistant. However, we also set a `translation_key` that overrides that.

If we were to keep defaulting `translation_key` to `attribute_name` in ZHA, we would still need to explicitly add `translation_key=None` if we want to keep the name of the device class (and make some changes to zigpy/ZHA to allow differentiating between setting it to `None` versus not setting it at all).

It seems much clearer to just separate the `attribute_name` and `translation_key` completely, so to also require `translation_key` when we actually want to set it (instead of setting it to `None` when we do not want to set it).

---

*: A follow-up PR to ZHA will remove the section that defaults the `_attr_translation_key` to `attribute_name` or `command_name`.
`fallback_name` and `translation_key` will also be added to a few quirks v2 entities in ZHA tests.

### DIFF
--- a/zigpy/quirks/v2/__init__.py
+++ b/zigpy/quirks/v2/__init__.py
@@ -246,11 +246,24 @@ class EntityMetadata:
     translation_key: str | None = attrs.field(default=None)
     fallback_name: str = attrs.field(validator=attrs.validators.instance_of(str))
 
+    def __attrs_post_init__(self) -> None:
+        """Validate the entity metadata."""
+        self._validate()
+
     def __call__(self, device: CustomDeviceV2) -> None:
         """Add the entity metadata to the quirks v2 device."""
+        self._validate()
         device.exposes_metadata[
             (self.endpoint_id, self.cluster_id, self.cluster_type)
         ].append(self)
+
+    def _validate(self) -> None:
+        """Validate the entity metadata."""
+        has_device_class: bool = getattr(self, "device_class", None) is not None
+        if self.translation_key is None and not has_device_class:
+            raise ValueError(
+                f"EntityMetadata must have a translation_key or device_class: {self}"
+            )
 
 
 @attrs.define(frozen=True, kw_only=True, repr=True)
--- a/zigpy/quirks/v2/__init__.py
+++ b/zigpy/quirks/v2/__init__.py
@@ -243,7 +243,7 @@ class EntityMetadata:
     cluster_type: ClusterType = attrs.field(default=ClusterType.Server)
     initially_disabled: bool = attrs.field(default=False)
     attribute_initialized_from_cache: bool = attrs.field(default=True)
-    translation_key: str = attrs.field()
+    translation_key: str | None = attrs.field(default=None)
     fallback_name: str = attrs.field(validator=attrs.validators.instance_of(str))
 
     def __call__(self, device: CustomDeviceV2) -> None:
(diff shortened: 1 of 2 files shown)


TheJulianJES:
I'm still open to thoughts on this, of course. These were just the conclusions I came to whilst working/reviewing some v2 quirks.
And yes, these changes would be breaking to v2 quirks. It's easy to solve them though.
Since we don't have any proper v2 quirks docs yet, there also aren't a lot of v2 quirks yet, so we probably can make these changes still.

zigpy/quirks/v2/__init__.py
```diff
@@ -243,15 +243,28 @@ class EntityMetadata:
     cluster_type: ClusterType = attrs.field(default=ClusterType.Server)
     initially_disabled: bool = attrs.field(default=False)
     attribute_initialized_from_cache: bool = attrs.field(default=True)
-    translation_key: str = attrs.field()
-    fallback_name: str = attrs.field()
+    translation_key: str | None = attrs.field(default=None)
+    fallback_name: str = attrs.field(validator=attrs.validators.instance_of(str))
```
TheJulianJES, line 247: I've used the in-built validation of `attrs` here. This is the message printed when `fallback_name` isn't provided:
```python
TypeError: ("'fallback_name' must be <class 'str'> (got None that is a <class 'NoneType'>)."
```

If wanted, I can also add it to our own validation below and add a clearer error message. I think the in-built one should be fine though.

zigpy/quirks/v2/__init__.py
```diff
@@ -243,15 +243,28 @@ class EntityMetadata:
     cluster_type: ClusterType = attrs.field(default=ClusterType.Server)
     initially_disabled: bool = attrs.field(default=False)
     attribute_initialized_from_cache: bool = attrs.field(default=True)
-    translation_key: str = attrs.field()
```
TheJulianJES, line 246: It looks like `translation_key` was already "required" somewhat since https://github.com/zigpy/zigpy/pull/1474, but there was never any validation to it. I'm not sure that change was intended to be kept in that PR though.

We now change the type to allow `None` again, in cases for when `device_class` is set.
Do note that  both `translation_key` and `device_class` can still be set simultaneously.

PR #1502, opened 2024-11-12
Re-add the `CONF_STARTUP_ENERGY_SCAN` constant

puddly:
It seems that we use it in a bunch of libraries and unit tests. This will allow it to still be imported.


TheJulianJES (approved):

TheJulianJES:
We should try and clean up the libraries/tests in the future then 😄

PR #1522, opened 2024-12-22
Remove Salus OTA provider

TheJulianJES:
## Proposed change

This completely removes the Salus OTA provider from zigpy, as it no longer works due to Salus replacing their app and taking the links offline.
Only the config options are kept, so existing configuration doesn't error out for now.

## Additional information

In the future, we might be able to partially revert this if we have the new OTA link (assuming the structure is mostly the same).


zigpy/ota/providers.py
```diff
@@ -386,38 +353,18 @@ async def _load_index(
             )
 
 
+# stub provider to keep existing configurations working
 @register_provider
 class Salus(BaseOtaProvider):
     NAME = "salus"
-    MANUFACTURER_IDS = (4216, 43981)
```
TheJulianJES, line 392: Due to removing the manufacturer ids, `compatible_with_device` will always return `True` now.

We could either keep the Salus IDs (no images will be loaded), or just use `-1` as an ID, or override `compatible_with_device` with `return False`.

puddly: I think you can leave them in, since no images will ever be loaded.

PR #1524, opened 2025-01-04
Add quirk v2 builder methods for modifying endpoints

TheJulianJES:
This adds quirk v2 builder methods for modifying endpoints.

The methods closely align to the cluster `adds`, `removes`, and `replaces` methods and act the same.
I.e. similar to `adds`, `adds_endpoint` does not modify the `profile_id` and `profile_type` of an existing endpoint.

This is needed to improve https://github.com/zigpy/zha-device-handlers/pull/3564 and to address https://github.com/home-assistant/core/issues/134686.

Other considerations:
- Do we want a default for `profile_id` and/or `profile_type`?

### DIFF
--- a/zigpy/quirks/v2/__init__.py
+++ b/zigpy/quirks/v2/__init__.py
@@ -263,9 +263,9 @@ class AddsEndpointMetadata:
     def __call__(self, device: CustomDeviceV2) -> None:
         """Process the add."""
         if self.endpoint_id not in device.endpoints:
-            device.add_endpoint(self.endpoint_id)
-            setattr(device.endpoints[self.endpoint_id], SIG_EP_PROFILE, self.profile_id)
-            setattr(device.endpoints[self.endpoint_id], SIG_EP_TYPE, self.device_type)
+            ep = device.add_endpoint(self.endpoint_id)
+            ep.profile_id = self.profile_id
+            ep.device_type = self.device_type
 
 
 @attrs.define(frozen=True, kw_only=True, repr=True)
--- a/zigpy/quirks/v2/__init__.py
+++ b/zigpy/quirks/v2/__init__.py
@@ -88,6 +88,16 @@ class CustomDeviceV2(BaseCustomDevice):
             list[EntityMetadata],
         ] = collections.defaultdict(list)
 
+        # endpoints need to be modified before clusters
+        for add_endpoint_meta in quirk_metadata.adds_endpoint_metadata:
+            add_endpoint_meta(self)
+
+        for remove_endpoint_meta in quirk_metadata.removes_endpoint_metadata:
+            remove_endpoint_meta(self)
+
+        for replace_endpoint_meta in quirk_metadata.replaces_endpoint_metadata:
+            replace_endpoint_meta(self)
+
         for add_meta in quirk_metadata.adds_metadata:
             add_meta(self)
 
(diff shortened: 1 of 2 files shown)


zigpy/quirks/v2/__init__.py
```diff
@@ -243,6 +252,46 @@ def __call__(self, device: CustomDeviceV2) -> None:
...
+
+    endpoint_id: int = attrs.field()
+    profile_id: int = attrs.field()
+    device_type: int = attrs.field()
+
+    def __call__(self, device: CustomDeviceV2) -> None:
+        """Process the add."""
+        if self.endpoint_id not in device.endpoints:
+            device.add_endpoint(self.endpoint_id)
+            setattr(device.endpoints[self.endpoint_id], SIG_EP_PROFILE, self.profile_id)
```
puddly: Since this is working with the `Endpoint` object directly, I think `ep = device.add_endpoint(...)` and `ep.profile_id = ...` would be clearer.

puddly:
I think `profile_id=zigpy.profiles.zha.PROFILE_ID` would be a good default. The device type I'm not sure about though, do we rely on it at all when creating entities? If not, anything would work because there's no way other devices on the network can even query it.

zigpy/quirks/v2/__init__.py
TheJulianJES: Ah yeah, that's what I initially had, but mypy complained, as I accessed it using `device.endpoints` and that returns type `ZDO | Endpoint` and `ZDO` doesn't have `profile_id` and `device_type`.
Your solution should work, missed that at first.

puddly: I think `ep = device.endpoints[self.endpoint_id]` would do it but `ep = device.add_endpoint(...)` should always be a `Endpoint` type, not `Endpoint | ZDO`.

TheJulianJES: Yeah, I missed that part of your message. I initially had `ep = device.endpoints[self.endpoint_id]` which (obviously) didn't work. Fixed with https://github.com/zigpy/zigpy/pull/1524/commits/db5759b53eb454787c0105dc02bd9f1158a23f7e now.

TheJulianJES:
> The device type I'm not sure about though, do we rely on it at all when creating entities

Hmm, we do for differentiating switches vs lights and for trying to ignore `OnOff` entity creation for remotes in [discovery](https://github.com/zigpy/zha/blob/91cc37b567e703ff6082629ff00d13fc923bc0d6/zha/application/discovery.py#L476-L477), with the types defined [here](https://github.com/zigpy/zha/blob/91cc37b567e703ff6082629ff00d13fc923bc0d6/zha/application/registries.py#L27).

TheJulianJES:
We could use something undefined like `0xFF` for the device type. It's what's used in tests generally (`255`) and only the TI router device uses it, so I'm guessing it's some sort of default (in Z-Stack at least).
If we have that "undefined" device type, it won't conflict with any remote/switch lists we have in ZHA. I think that default or no defaults at all are both fine, as these methods will (hopefully) be rarely used.

I'd argue for `profile_id: int = zigpy.profiles.zha.PROFILE_ID` and `device_type: int = 0x00FF` as defaults, but we can also just set a default for `profile_id` otherwise (or none at all).

TheJulianJES:
Hmm, for the cluster methods, both `adds()` and `AddsMetadata` set defaults. That seems a bit redundant.
Here, I've set the defaults in `adds_endpoint()` and `replaces_endpoint()`, but not in the relevant metadata classes. I believe having it in the builder methods will make it easier for others to see the defaults, but can change if wanted of course.

Let me know what you think of these defaults now. Patch release seems to be targeted for tomorrow, so if we can get these changes (and the small adjustments in quirks using `adds_endpoint()`) in, that would be nice.

puddly (approved):
Looks good to me!

TheJulianJES:
👍  I'll test this with the quirks changes before merging.

TheJulianJES:
Ok, one more small (but crucial) change. The endpoint metadata needs to be processed before the cluster metadata is processed, as we can't actually fix the H1 wireless remote issue otherwise. -> https://github.com/zigpy/zigpy/pull/1524/commits/eb8c82869e8cf0643f573de150ab9e67ff380e52
I've expanded the test case to also cover this now.

PR #1530, opened 2025-01-17
Add missing CR2450, CR2032, CR1632 to `BatterySize` enum

tr4nt0r:


TheJulianJES (commented):
I think we made these values up for quirks/ZHA. Official ZCL spec (R8) only goes to `0x08`/`CR123A`, so I'm not sure we should add the others to zigpy.

tr4nt0r:
I thought I'll maybe add these because HA dispalys the battery type correctly
image

but reading from the cluster shows me this:

image
 
Also these id's are already used in various quirks

puddly:
Zigpy focuses on just implementing the Zigbee specification. We really don't want to merge anything from quirks back into zigpy like this.

tr4nt0r:
Should I then just put the number in the quirk instead of using the enum?

MattWestb:
**If** implanting it then use somthing like `0xEX` range then `0xFF` is unknown and likely its can coming more from the bottom of the list in the future.
PS: have not found how Matter 1.4 is implanting it but the shall have it 2.

TheJulianJES:
> Should I then just put the number in the quirk instead of using the enum?

Yeah, you can do that. I think we could also add an enum directly to zhaquirks (with the values that aren't part of ZCL spec), since we use them in multiple places/quirks.

PR #1533, opened 2025-01-19
Mark bind, unbind, request, reply as async

RubenVerborgh:
The ZDO and Cluster methods `bind`, `unbind`, `request`, `reply` are treated in most places (including tests) as async, but are not defined as such. Dependencies such as zha-device-handlers seem to use a mixture of both, whereas [ZHA awaits `bind`](https://github.com/zigpy/zha/blob/fd6cf2f570f1778a9c0933a56ebf476caafe859d/zha/zigbee/cluster_handlers/__init__.py#L280).

Based on the discussion in https://github.com/zigpy/zha-device-handlers/pull/3704#issuecomment-2585520722, this PR fixes the definitions.

### DIFF
--- a/zigpy/zcl/__init__.py
+++ b/zigpy/zcl/__init__.py
@@ -682,11 +682,11 @@ class Cluster(util.ListenableMixin, util.CatchingTaskMixin):
         args = self._write_attr_records(attributes)
         return self._write_attributes_undivided(args, manufacturer=manufacturer)
 
-    def bind(self):
-        return self._endpoint.device.zdo.bind(cluster=self)
+    async def bind(self):
+        return await self._endpoint.device.zdo.bind(cluster=self)
 
-    def unbind(self):
-        return self._endpoint.device.zdo.unbind(cluster=self)
+    async def unbind(self):
+        return await self._endpoint.device.zdo.unbind(cluster=self)
 
     def _attr_reporting_rec(
         self,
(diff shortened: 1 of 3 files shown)


puddly (approved):

zigpy/zcl/__init__.py
```diff
@@ -682,10 +682,10 @@ def write_attributes_undivided(
         args = self._write_attr_records(attributes)
         return self._write_attributes_undivided(args, manufacturer=manufacturer)
 
-    def bind(self):
+    async def bind(self):
         return self._endpoint.device.zdo.bind(cluster=self)
```
puddly: You will need to `await` the underlying function call as well:

```suggestion
        return await self._endpoint.device.zdo.bind(cluster=self)
```

zigpy/zcl/__init__.py
```diff
@@ -682,10 +682,10 @@ def write_attributes_undivided(
...
 
-    def unbind(self):
+    async def unbind(self):
         return self._endpoint.device.zdo.unbind(cluster=self)
```
puddly: Here too:

```suggestion
        return await self._endpoint.device.zdo.unbind(cluster=self)
```

puddly (changes requested):

zigpy/zcl/__init__.py
RubenVerborgh: Thanks, done in 8f64a19eaf061d0b02604735ecba4e29c6365769.

zigpy/zcl/__init__.py
RubenVerborgh: Done in 8f64a19eaf061d0b02604735ecba4e29c6365769.

PR #1541, opened 2025-01-26
Add `unique_id_suffix` to quirks v2

TheJulianJES:
## Proposed change

To allow multiple sensors for the same ZCL attribute (with different parsing each), we need to have separate unique IDs for each.
To allow this, we add a `unique_id_suffix` to the `EntityMetadata`. When implemented in ZHA, this will replace the default "unique id suffix" (which is the attribute name for v2 quirks)

This does NOT need to be set by default, it is only needed when there are two entities for the same attribute (on the same cluster and on the same endpoint). Then, it should ideally be set on both.
We should likely add tests to the quirks repo to see that these are correctly set. Generally, it should be pretty simple though.

## Additional information

We already have a couple of PRs blocked that use the quirks v2 button for the same attribute, but with a different value to be written. This is another important use case for `unique_id_suffix`.

Corresponding ZHA PR:
- https://github.com/zigpy/zha/pull/362


puddly:
You mention a use case of replacing the default `unique_id` naming scheme: could we replace this PR it, where an entity would explicitly define `unique_id="something_added_to_the_device_unique_id"`? This would let v2 quirks take over processing for existing entities or v1 quirks.

With v2 quirks requiring explicitly exposing every entity, I don't think there's a possibility of two exposed entities having the same unique ID (and we would have checks for this anyways within ZHA).

TheJulianJES:
> replacing the default `unique_id` naming scheme: could we replace this PR it, where an entity would explicitly define `unique_id="something_added_to_the_device_unique_id"`? This would let v2 quirks take over processing for existing entities or v1 quirks.

See this comment in the ZHA PR: https://github.com/zigpy/zha/pull/362/files#r1930901083

TheJulianJES:
Assuming we want to keep the `unique_id_suffix` name here, this PR should also be good. The name means the same as for "v1"/ZHA entities now, so I think it's fine. The implementation on the [ZHA side](https://github.com/zigpy/zha/pull/362) is also adjusted now.

PR #1546, opened 2025-01-29
Backport and use `StrEnum` for quirks v2 units

TheJulianJES:
## Proposed change

We need the quirks v2 unit enum members to be strings, similar to the unit implementation in ZHA and HA Core, so this PR backports `StrEnum` and adds test coverage for the backport.
We can't use the native `StrEnum` implementation due to needing to support older Python versions with zigpy.

The fix is confirmed working in ZHA.

## Additional information

This issue wasn't caught in ZHA tests, because they mistakenly imported ZHA units (which are `StrEnum`) and not zigpy's quirks v2 unit enums. The ZHA test fix is linked below.

With https://github.com/zigpy/zigpy/pull/1547, we'll also have a regression test for this in zigpy and with https://github.com/zigpy/zha/pull/371, we'll have one in ZHA too.

### DIFF
--- a/zigpy/backports/enum.py
+++ b/zigpy/backports/enum.py
@@ -3,15 +3,17 @@
 from __future__ import annotations
 
 from enum import Enum
-from typing import Any, Self
+from typing import Any, TypeVar
+
+_StrEnumSelfT = TypeVar("_StrEnumSelfT", bound="StrEnum")
 
 
 class StrEnum(str, Enum):
     """Partial backport of Python 3.11's StrEnum for our basic use cases."""
 
-    value: str
-
-    def __new__(cls, value: str, *args: Any, **kwargs: Any) -> Self:
+    def __new__(
+        cls: type[_StrEnumSelfT], value: str, *args: Any, **kwargs: Any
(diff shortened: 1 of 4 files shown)


puddly (approved):

zigpy/backports/enum.py
```diff
@@ -0,0 +1,35 @@
...
+
+_StrEnumSelfT = TypeVar("_StrEnumSelfT", bound="StrEnum")
+
+
+class StrEnum(str, Enum):
+    """Partial backport of Python 3.11's StrEnum for our basic use cases."""
+
+    def __new__(
+        cls: type[_StrEnumSelfT], value: str, *args: Any, **kwargs: Any
+    ) -> _StrEnumSelfT:  # noqa: PYI019
```
TheJulianJES, line 16: `# noqa: PYI019` is needed for ruff here, as it'll otherwise complain about this:
``PYI019 Methods like `__new__` should return `typing.Self` instead of a custom `TypeVar``
... which we can't use because `typing.Self` was only added in Python 3.11 and we still support 3.9, 3.10, and 3.10 (and don't use `typing-extensions`).

PR #1558, opened 2025-02-26
Expose an API to write lock the database

puddly:
For Home Assistant backups to be safe, we should write lock the database during `pre_backup` and unlock it during `post_backup`. This will effectively queue all zigpy database writes for many minutes but we don't rely on the database to hold data at runtime, just to act as a store.

https://github.com/home-assistant/core/issues/138544

### DIFF
--- a/zigpy/appdb.py
+++ b/zigpy/appdb.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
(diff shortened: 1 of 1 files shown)


pipiche38:
is there a risk that on bellows implementation, we come to a Timeout because the backup will last more than 5 seconds ?

PR #1564, opened 2025-03-07
Packet reordering

puddly:
Reported in https://github.com/zigpy/zha/issues/388.

This PR implements packet reordering for zigpy:
- if a packet with a non-consecutive TSN is received, we will hold out-of-order packets for a period of time (probably 3s for mains devices and 10s for battery?). 
- If any packets arrive in-order, we emit those immediately.
- Those that are received out of order will be emitted after the set period of time if they remain out of order.
- This replaces packet deduplication: duplicate packets will be filtered unless their contents actually differ, in which case the duplicate packet will be emitted.

For the vast majority of devices, packet order doesn't matter since a single packet usually corresponds to a single update. But for those that send batches of attribute reports in separate packets, we should try to keep the order consistent.

We can use missed packets as a reliability metric as well, giving us more information than LQI/RSSI, especially for end devices.

---

Here's a crude tool to analyze HA log files and find any out-of-order or skipped packets:

<details>
  <summary>Expand <code>out_of_order.py</code></summary>

```python
import re
import sys
import datetime

def parse_log_line(line):
    """
    Parses a single log line and extracts:
      - the leading log timestamp as a datetime object,
      - the source NWK address,
      - the TSN (Transaction Sequence Number), and
      - the full stripped log line.
    
    Only lines with NWK addressing (addr_mode=<AddrMode.NWK: 2>) are processed.
    
    Returns:
        tuple: (nwk_address (str), tsn (int), timestamp (datetime), packet (str))
               If a component is not found, None is returned in its place.
    """
    # Extract the timestamp at the beginning of the line (format: "YYYY-MM-DD HH:MM:SS.sss")
    timestamp_match = re.match(r'^(\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}\.\d+)', line)
    if timestamp_match:
        timestamp_str = timestamp_match.group(1)
        try:
            timestamp = datetime.datetime.strptime(timestamp_str, "%Y-%m-%d %H:%M:%S.%f")
        except Exception:
            timestamp = None
    else:
        timestamp = None

    # Only consider lines that mention NWK address mode.
    if "addr_mode=<AddrMode.NWK: 2>" not in line:
        return None, None, timestamp, line.strip()

    # Extract the source NWK address, e.g., address=0xD282
    src_match = re.search(r'src=AddrModeAddress\(addr_mode=<AddrMode\.NWK: 2>, address=(0x[0-9A-Fa-f]+)\)', line)
    if not src_match:
        return None, None, timestamp, line.strip()
    nwk_addr = src_match.group(1)

    # Extract the TSN (Transaction Sequence Number)
    tsn_match = re.search(r'tsn=(\d+)', line)
    if not tsn_match:
        return nwk_addr, None, timestamp, line.strip()
    tsn = int(tsn_match.group(1))
    
    return nwk_addr, tsn, timestamp, line.strip()

def process_log_file(file_path):
    """
    Processes the log file line by line, checking for:
      - out-of-order TSN values, and
      - printing the time delta between logged packets for each device.
    
    It maintains two dictionaries:
      - last_tsn: maps each NWK address to the last seen TSN.
      - last_timestamp: maps each NWK address to the last seen datetime.
    
    If a new packet for an NWK address has a TSN less than or not equal to the expected next TSN,
    it prints a warning. Additionally, if a previous timestamp exists for the device,
    the script prints the time delta (in seconds) since the last packet.
    """
    last_tsn = {}        # Maps NWK address to the last TSN seen.
    last_timestamp = {}  # Maps NWK address to the last timestamp seen.
    count = 0

    with open(file_path, 'r') as f:
        for line in f:
            nwk_addr, tsn, timestamp, packet = parse_log_line(line)
            if nwk_addr is None or tsn is None or timestamp is None:
                continue

            # Skip packets from 0x0000.
            if nwk_addr == "0x0000":
                continue

            count += 1

            # Compute and print time delta if a previous timestamp exists for this device.
            if nwk_addr in last_timestamp:
                delta = (timestamp - last_timestamp[nwk_addr]).total_seconds()

            # Update the last seen timestamp for this NWK address.
            last_timestamp[nwk_addr] = timestamp

            # If we have seen this NWK address before, check the TSN order.
            if nwk_addr in last_tsn:
                expected_next = (last_tsn[nwk_addr] + 1) % 256  # Wrap-around TSN at 0xFF + 1.
                if tsn < expected_next:
                    print(f"[{count:10}]  TSN rollback      {nwk_addr}: {tsn:3} ( < {last_tsn[nwk_addr]:3})  {delta:8.2f}s delta")
                elif tsn > expected_next:
                    print(f"[{count:10}]  Out-of-order TSN  {nwk_addr}: {tsn:3} (!= {expected_next:3})  {delta:8.2f}s delta")

            # Update the last seen TSN for this NWK address.
            last_tsn[nwk_addr] = tsn

if __name__ == '__main__':
    if len(sys.argv) != 2:
        print("Usage: python out_of_order.py <logfile>")
        sys.exit(1)

    log_file = sys.argv[1]
    process_log_file(log_file)

```
</details>


puddly:
CC @sigurasg

sigurasg:
I'm no expert, but from looking through the logs I have from my network, I see the following:
1. One device seems to issue attribute updates using every other TSN.
    Presumably this device's updates would always be delayed by the timeout?
3. Some devices roll on 2^7, some on 2^8.
    From reading random ZigBee SDK docs and QAs, it seems there are some stacks that reserve TSN 0x80-0xFF for the stack, whatever that means. I wonder how a device that alternates between the two ranges will look like under this implementation - especially when the two TSN counters are near each other.
4. There's at least one device on my network that has at least once skipped a bunch of TSNs.
5. I find only one TSN inversion, though there could be more.

Like you say, for most devices this is a non-issue. It's only where multiple attributes combine to a property where this is a problem, so perhaps it makes more sense to tackle this on a per-device basis, and then even on a per property basis?
IDK much about ZigBee in general, but is this perhaps unique to devices that implement the ElectricalMeasurement cluster?

Your script doesn't work for me, as all incoming frames seem to have `tsn=None`. Perhaps the ConBee implementation doesn't parse it? Not a big deal, I just need 10 minutes to "adopt" it to my logs :).

puddly:
> Your script doesn't work for me, as all incoming frames seem to have tsn=None.

Ah... Well, that is a show-stopper 🤦. My implementation relies on the APS sequence counter to reorder frames. We could reorder based on the NWK frame counter, which is another 8-bit counter (or rather, transaction identifier) that will be present in all of the relevant frames:

```python
2025-03-07 13:00:33.405 DEBUG (MainThread) [zigpy.zcl] [0xD282:1:0x0b04] Decoded ZCL frame header: ZCLHeader(frame_control=FrameControl<0x18>(frame_type=<FrameType.GLOBAL_COMMAND: 0>, is_manufacturer_specific=0, direction=<Direction.Server_to_Client: 1>, disable_default_response=1, reserved=0, *is_cluster=False, *is_general=True), tsn=87, command_id=10, *direction=<Direction.Server_to_Client: 1>)
```

Perhaps we'll restrict this feature to Zigbee stacks that expose this information. EmberZNet and Z-Stack do, I think only the legacy stacks do not.

puddly:
I've tested this out on my home network and I think APS-level reordering isn't going to work very well. Many of my IKEA end devices (namely remotes) routinely skip TSNs when transmitting, resulting in frequent 5s delays with this patch set.

I'll try it out with ZCL traffic only and see if that works better.

sigurasg:
> I've tested this out on my home network and I think APS-level reordering isn't going to work very well. Many of my IKEA end devices (namely remotes) routinely skip TSNs when transmitting, resulting in frequent 5s delays with this patch set.
> 
> I'll try it out with ZCL traffic only and see if that works better.

I think wrtg. the INSPELNING problem, the only case I've found to be a problem is when both attributes are updated in a single command, in which case this isn't a frame reordering problem. So I think it'd be good enough for that case to use a timer of 0s (or somehow to post to the top of the event loop) to issue the property update.

sigurasg:
> I think wrtg. the INSPELNING problem, the only case I've found to be a problem is when both attributes are updated in a single command, in which case this isn't a frame reordering problem.

... this is not to say that frame reordering isn't a problem, it's just much, much rarer than the multi property update case.

PR #1565, opened 2025-03-08
Make OTA more robust V2

FlorianBruckner:
This PR only changes the image transfer to not stop the upgrade process when a page or block transfer fails with an error on the stack.

The devices I have tested with re-request blocks if they are not received on the device after some time.

see #1552 for a more detailed discussion, this PR contains just a subset of the changes suggested there.


puddly:
I've tested this change and intentionally halted the OTA progress with a jammer for 20s, everything continues just fine even after zigpy fails to send an image block and the device internally retries.

PR #1568, opened 2025-03-10
Support new Z2M OTA index metadata

puddly:
Populates our `min_hardware_version`, `max_hardware_version`, and `changelog` fields.

### DIFF
--- a/tests/ota/files/z2m_index.json
+++ b/tests/ota/files/z2m_index.json
@@ -1,3286 +1,7492 @@
 [
-    {
-        "fileVersion": 36832016,
-        "fileSize": 134487,
-        "manufacturerCode": 4474,
-        "imageType": 53250,
-        "sha512": "53c6cadc6a83d39f2cfa2e29d3d7f15e48ead17f5e4d46ae81466c263696cbe9ec7656818a900f6d963d8bc081e0c0671662a12504e53ec73751e151c84c410a",
-        "url": "https://github.com/Koenkk/zigbee-OTA/raw/master/images/Insta/ZLL_WS_4f_J_Release_10.03.32.02.zigbee",
-        "path": "images/Insta/ZLL_WS_4f_J_Release_10.03.32.02.zigbee"
-    },
-    {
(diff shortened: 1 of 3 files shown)


puddly:
@FlorianBruckner just to clarify, this PR isn't necessary for ZHA to utilize the hardware versions. Using the Z2M metadata would let us filter the inapplicable image out *before* downloading them but we always use the metadata in the actual OTA file in the end.

FlorianBruckner:
@puddly this PR is vital for allowing Homeassistant to offer OTA for affected devices.

Before https://github.com/Koenkk/zigbee-OTA/pull/717, the catalog did only contain a single image for ubisys S1 with a hardware constraint that didn't match my device. Homeassistant offered the OTA, but bailed out after starting the upgrade because only then the hardware constraints were checked.

After [the PR](https://github.com/Koenkk/zigbee-OTA/pull/717), the catalog now contains two images for the same version but with different hardware constraints. This leads to `Multiple unique OTA images for version 0192020D with specificity 202 exist` _when scanning for available upgrades_. So, Homeassistant will never offer the upgrade in this case, because the two conflicing images are discarded from the catalog.

Only when including the properties from the catalog as contained in this PR it will offer the upgrade - I did try this in my installation with a patch similar to this PR, so I know that it would not offer the upgrade without it but does when hardware constraints are loaded from the catalog.

Only ubisys S1 devices have this kind of setup in Koenkk's catalog. So it probably is an edge case and I am already far too deep in that rabbit hole for the single ubisys S1 I own (and now have successfully upgraded).

puddly:
I forgot to mention that there was another bugfix PR (https://github.com/zigpy/zigpy/pull/1567) that fixed a bug where we did not filter out incompatible images after downloading them. I added a unit test for the exact situation you mention above and I'm pretty confident the correct behavior will happen now in Home Assistant Core 2025.3.2 (released a few hours ago).

This PR can't really me merged during the middle of a release cycle because it adds a few new features (namely release notes) but it'll be merged in a few weeks.

Issue #1566, opened 2025-03-09
OTA fails for ubisys devices


FlorianBruckner:
Koenkk/zigbee-OTA#717 fixed an issue in the firmware catalog for ubisys S1 devices. Ubisys provides two different images with the same version, but mutually exclusive hardware constraints.

Previously, OTA for the device would have been offered in HA, but could not be completed because only the wrong image was in the catalog,. It failed with NO_IMAGE_AVAILABLE due to checking the hardware constraint.

Now, both images are in the catalog:

```{
    "fileName": "10F2-7B02-0000-0001-0192020D-spo-fms.ota1.zigbee",
    "fileVersion": 26345997,
    "fileSize": 256474,
    "url": "https://raw.githubusercontent.com/Koenkk/zigbee-OTA/master/images/Ubisys/10F2-7B02-0000-0001-0192020D-spo-fms.ota1.zigbee",
    "imageType": 31490,
    "manufacturerCode": 4338,
    "sha512": "20ff31830a17149b353db48afb9ccd356ac75dec04ea0241ef996918471f4b5bc786916caeae79ba75fffbebd6412f06361032a01fd675e85bb931a5fefd7bf3",
    "otaHeaderString": "ubisys S1\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000",
    "originalUrl": "http://fwu.ubisys.de/smarthome/OTA/release/10F2-7B02-0000-0001-0192020D-spo-fms.ota1.zigbee",
    "hardwareVersionMax": 1,
    "hardwareVersionMin": 0
  },
  {
    "fileName": "10F2-7B02-0002-0007-0192020D-spo-fms.ota1.zigbee",
    "fileVersion": 26345997,
    "fileSize": 256200,
    "originalUrl": "http://fwu.ubisys.de/smarthome/OTA/release/10F2-7B02-0002-0007-0192020D-spo-fms.ota1.zigbee",
    "url": "https://raw.githubusercontent.com/Koenkk/zigbee-OTA/master/images/Ubisys/10F2-7B02-0002-0007-0192020D-spo-fms.ota1.zigbee",
    "imageType": 31490,
    "manufacturerCode": 4338,
[...]

PR #1569, opened 2025-03-11
Read hardwareVersionMin from Z2M catalog

FlorianBruckner:
It appears the root cause for conflicting images from #1566 is that hardware constraints are not taken into account for the Z2M catalog.


FlorianBruckner:
#1568 has already done these modifications.

PR #1576, opened 2025-03-18
Primary entity registration

puddly:
Add a new `primary` kwarg to all v2 entities, allowing us to specify which one should be considered primary. Some validation is performed to ensure only a single one can exist.


TheJulianJES (commented):
If we're adding `primary` to all entity metadata types, can't we add it to `EntityMetadata` directly (where translation key and fallback name are present as well)?

puddly:
You're totally right! Fixed.

TheJulianJES (approved):

PR #1578, opened 2025-03-28
Fix OTA file checksum validation error message formatting. 

aquaticus:
When the OTA image has an incorrect checksum, an exception is raised: *Image checksum is invalid: expected AAAAAAA, got BBBBBBBBB*.  

The expected value comes from the JSON file, but it should be the computed value instead.  

This patch swaps the A and B values in the error message.

### DIFF
--- a/zigpy/ota/providers.py
+++ b/zigpy/ota/providers.py
@@ -76,8 +76,8 @@ class BaseOtaImageMetadata(t.BaseDataclassMixin):
 
             if hasher.hexdigest() != checksum:
                 raise ValueError(
-                    f"Image checksum is invalid: expected {checksum},"
-                    f" got {hasher.hexdigest()}"
+                    f"Image checksum is invalid: expected {hasher.hexdigest()},"
+                    f" got {checksum}"
                 )
 
     async def fetch(self) -> BaseOTAImage:


puddly:
I think the JSON has the correct value for the hash so it should the value to expect, no? And instead of getting the expected value, we get the result of the checksum computation.

aquaticus:
Say you got in index.json file
```json
        {
            "path": "/config/local_ota/example_ota.bin",
            "file_version": 1,
            "image_type": 1234,
            "manufacturer_id": 5678,
            "file_size": 1749710,
            "checksum": "sha3-256:57a007f35af2b0d6567931932333633ee7712436438fec6fdc5eece23a45319b"
        }
```
`57a007f35af2b0d6567931932333633ee7712436438fec6fdc5eece23a45319b` is invalid hash, and the valid is `8a536bff63093b4c846075cb4345d0167e13a3fb81eaf3507058357034817338`
In the log you'll see:
`Image checksum is invalid: expected 57a007f35af2b0d6567931932333633ee7712436438fec6fdc5eece23a45319b, got 8a536bff63093b4c846075cb4345d0167e13a3fb81eaf3507058357034817338`,

What you **got** (from file) is `57a007f35af2b0d6567931932333633ee7712436438fec6fdc5eece23a45319b` and you **expect** hash value would be  `8a536bff63093b4c846075cb4345d0167e13a3fb81eaf3507058357034817338`. So, the values in the message should be swapped. _Expected_ value is the one you expect to be valid.

puddly:
The `index.json` value is always right and the expected value. There should never be an instance where the file has the correct hash but the index does not. That's the purpose of the index file: to hold metadata that isn't contained in the OTA file itself.

aquaticus:
I'm describing a scenario where `index.json` is generated (e.g., for zha_remote), but the computed value in `index.json` is incorrect due to issues like a bug in the generator or file corruption.

When Zigpy downloads an OTA file, it calculates its hash and compares it to the value in `index.json`. If the computed values don't match, an error is logged. In that workflow, expected value is the right one, computed by zigpy.

puddly:
> In that workflow, expected value is the right one, computed by zigpy.

The checksum of the actual OTA file is irrelevant unless there's a corresponding `index.json` file to hold the "correct" value. Zigbee OTA files do not have any sort of integrity checking so with just the OTA file alone, you cannot tell what its actual checksum should be unless you download it from a known-valid source.

If here's a mismatch, there are two scenarios:

1. The downloaded OTA file has been corrupted, either in transit (e.g. no TLS) or at rest (SD card corruption). This can also happen if the underlying OTA file is swapped out or renamed by accident. The authoritative source for the checksum is going to be the index file, which is the behavior zigpy has now.

2. The index file itself is invalid. We can't rely on the OTA file to have a valid checksum so this is just a hard failure and should not happen. If it does, patching the index file's checksum to match that of the OTA file defeats the purpose of an out-of-band checksum field and would be dangerous.

I think our current behavior is correct for the 99.9% case of the downloaded OTA file's checksum differing from the index's. If the index doesn't match the file, we have no way of knowing which one is right. I think it's a safe assumption to say that it will be the index.

aquaticus:
I'm not questioning the behavior. This is a standard pattern and it's perfectly OK. I'm only questioning what is expected value in the message. It should be the value computed not the value that comes in index.json.

To be honest, we wrote to much about just a simple cosmetic change. :-)
I see no point to keep PR open.

PR #1582, opened 2025-04-09
Add `cluster_type` to `prevent_default_entity_creation`

puddly:
This allows client cluster entities to be excluded from ZHA. Requires a small ZHA patch:

```diff
diff --git a/zha/zigbee/device.py b/zha/zigbee/device.py
index 4e3da83..317ea46 100644
--- a/zha/zigbee/device.py
+++ b/zha/zigbee/device.py
@@ -793,7 +793,11 @@ class Device(LogMixin, EventBase):
 
             if meta.cluster_id is not None and not any(
                 cluster_handler.cluster.cluster_id == meta.cluster_id
-                for cluster_handler in entity.cluster_handlers.values()
+                for cluster_handler in (
+                    entity.cluster_handlers
+                    if meta.cluster_type == ClusterType.Server
+                    else entity.client_cluster_handlers
+                ).values()
             ):
                 continue
```

### DIFF
--- a/zigpy/quirks/v2/__init__.py
+++ b/zigpy/quirks/v2/__init__.py
@@ -445,6 +445,7 @@ class PreventDefaultEntityCreationMetadata:
 
     endpoint_id: int | None = attrs.field()
     cluster_id: int | None = attrs.field()
+    cluster_type: ClusterType | None = attrs.field()
     unique_id_suffix: str | None = attrs.field()
     function: Callable[Any, bool] | None = attrs.field()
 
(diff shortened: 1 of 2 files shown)


zigpy/quirks/v2/__init__.py
```diff
@@ -1102,14 +1103,19 @@ def prevent_default_entity_creation(
         *,
         endpoint_id: int | None = None,
         cluster_id: int | None = None,
+        cluster_type: ClusterType | None = None,
         unique_id_suffix: str | None = None,
         function: Callable[Any, bool] | None = None,
     ) -> QuirkBuilder:
         """Do not create default entities."""
+        if cluster_id is not None and cluster_type is None:
+            cluster_type = ClusterType.Server
```
TheJulianJES, line 1112: I'd have probably just put the Server cluster type as the default in the method definition, since ZHA won't care about the cluster type if there's no cluster id, but I'm also fine with this.

TheJulianJES (approved):

PR #1586, opened 2025-04-23
Implement suggested display precision for sensors

puddly:
Related to https://github.com/zigpy/zha/pull/408

### DIFF
--- a/zigpy/quirks/v2/__init__.py
+++ b/zigpy/quirks/v2/__init__.py
@@ -348,7 +348,7 @@ class ZCLSensorMetadata(EntityMetadata):
     reporting_config: ReportingConfig | None = attrs.field(default=None)
     divisor: int | None = attrs.field(default=None)
     multiplier: int | None = attrs.field(default=None)
-    suggested_display_precision: int = attrs.field(default=1)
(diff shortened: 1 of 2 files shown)


TheJulianJES (approved):

zigpy/quirks/v2/__init__.py
```diff
@@ -818,6 +819,7 @@ def sensor(
         endpoint_id: int = 1,
         divisor: int = 1,
         multiplier: int = 1,
+        suggested_display_precision: int = 1,
```
TheJulianJES, line 822: Do note that this will cause the suggested display precision to default to `1`.
This commit message might make it seem like we don't wanna do this: https://github.com/zigpy/zigpy/pull/1586/commits/9ca918b523d78f4a686db9d12017966eb1ad3ced
It was just changed in the metadata though, not the `sensor` method.

TheJulianJES: Ok, looking at the ZHA PR, it seems like `1` was the default before. So, this is fine (and setting the default here, instead of in the metadata follows how we do it for all parameters).

PR #1593, opened 2025-04-29
Reverse the field order of the `ApplicationType` int struct

puddly:
I believe our field order may be backwards. From the spec, the `index` should be the lowest-order 16 bits:

image


puddly:
CC @prairiesnpr, since I think you have a custom device implementing this feature.

PR #1597, opened 2025-05-07
Fix metering data types

puddly:
The `_formatting` attributes are strange in that they are `map8` data types but are meant to be interpreted as integers.

#1594 implemented the missing data types but failed to correctly set the ZCL data type. This PR fixes this


TheJulianJES (approved):
Seems good.

Since we use `zcl_type` for these now, I remembered that `zcl_type` might cause some issues when reporting is set up for these attributes, posted here: https://github.com/zigpy/zha-device-handlers/pull/2808/files#r1788545336

I don't think we set up reporting for any of these attributes though, but might be something to look at to make sure the correct datatype is used everywhere when `zcl_type` is also set.

PR #1600, opened 2025-05-24
Fix `TransmitOptions` bitmap member

puddly:
`FORCE_ROUTE_DISCOVERY` being set to `3` was not valid: it became an alias for `ACK | APS_Encryption`, since this is a bitmap, and broke retries for zigpy-znp.

Related: https://github.com/home-assistant/core/issues/145561#issuecomment-2907005753


puddly:
Looks like Copilot could have caught this 😄: https://github.com/zigpy/zigpy/pull/1592#discussion_r2105925893.

I've enabled Copilot reviews for all future zigpy PRs as a test, we can disable it if it's too noisy.

PR #1606, opened 2025-06-14
Quirk v2 firmware version filters

puddly:
Related: https://github.com/zigpy/zha-device-handlers/issues/3857

This would allow quirks to filter compatible firmware versions. As an example, here is the quirk for the above device rewritten to be a v2 quirk and using this proposed API:

```python
"""Innr SP 240 plug."""

from zigpy.quirks.v2 import QuirkBuilder
from zigpy.quirks import CustomCluster
from zhaquirks.innr import MeteringClusterInnr


class InnrCluster(CustomCluster):
    """Innr manufacturer specific cluster."""

    cluster_id = 0xE001

(
    QuirkBuilder(manufacturer="innr", model="SP 240")
    # Firmware version `421410437` fixed the divisor and multiplier bug
    .firmware_version_filter(max_version=0x191E3685)
    .replaces(MeteringClusterInnr, endpoint_id=1)
    .replaces(InnrCluster, endpoint_id=1)
    .add_to_registry()
)
```

### DIFF
--- a/tests/test_quirks_v2.py
+++ b/tests/test_quirks_v2.py
@@ -26,6 +26,7 @@ from zigpy.quirks.v2 import (
     EntityMetadata,
     EntityPlatform,
     EntityType,
+    FirmwareVersionFilterMetadata,
     NumberMetadata,
     PreventDefaultEntityCreationMetadata,
     QuirkBuilder,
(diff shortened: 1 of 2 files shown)


TheJulianJES:
Ideally, we probably also want zigpy/ZHA to re-match the quirks at runtime if the firmware version changes via an update (or if the attribute is just re-read when the update was done outside of ZHA and the broadcast image notify eventually refreshes the zigpy cache).

With the current zigpy implementation, I'd probably have kept a single quirk for all Innr plug versions and set the correct constant attributes depending on the OTA version attribute (or maybe `sw_build_id`).
Kind of similar to the IKEA remote quirk [doing this](https://github.com/zigpy/zha-device-handlers/blob/5f1f14b95d352ff1cf0cbd34ba86c496f259d2f5/zhaquirks/ikea/__init__.py#L276-L277), although we'd be able to avoid a lot of the "hacky stuff" done there to read an attribute on sleepy end devices when it reports an attribute, as well as the parsing of the `sw_build_id` value.

Either way, we need all possible firmware variants (of the different plug models) and what attributes they set. If a firmware update is done, old attributes only read during initial pairing are still cached, so we need to update/drop them, likely in the quirk.

I do agree that a quirk firmware matching implementation would be great to have and likely cleaner than having to do some of this in one quirk.

puddly:
Runtime swapping is definitely something to look at, I agree. I'm not sure how complicated that would be but it would be beneficial to at least have ZHA reconfiguration run through initialization a second time, re-reading all relevant attributes. We could then have OTA use similar functionality.

PR #1609, opened 2025-06-28
Pin pytest-asyncio to < 1.0.0

oscarh:
pytest-asyncio has incompatible changes. I guess the long-term solution would be to actually update the fixtures/tests to use pytest-asyncio 1.0.0, but this at least allows me to run the tests.


puddly:
Thanks!

PR #1612, opened 2025-07-03
Send the ZCL default response more often

puddly:
Supersedes https://github.com/zigpy/zigpy/pull/1599 and fixes https://github.com/zigpy/zha-device-handlers/issues/3579.

PR #1599, opened 2025-05-21
fix(zcl): send default response from handle_cluster_request

AnatolyRugalev:
This PR addresses an issue for Sonoff ZMBINIR2, discussed here: https://github.com/zigpy/zha-device-handlers/issues/3579

In short, this device expects a default response on some of its commands in order to function properly, and ZHA currently does not send default response packets in general cluster implementations.

I am default response to `OnOff` cluster, however, it may be sensible to have this in some other clusters as well

PR #1614, opened 2025-07-07
Remove the `ZCLCommand` `direction` attribute from command definitions

puddly:
I've been investigating #1612 / #1599 and have run across a few bugs. This is a fairly simple change but more than a few commands (including commonly-used ones!) have bad `direction` fields, including the `Time` cluster and a few `Ota` commands. We weren't really using the `direction` for much but it definitely impacts the generation of default replies.

This will likely have to be downgraded to a loud warning instead of a hard error due to custom quirks likely existing with this bug. Alternatively, we can break compatibility, since custom quirks are entirely the responsibility of the person using them.

### DIFF
--- a/zigpy/zcl/clusters/general.py
+++ b/zigpy/zcl/clusters/general.py
@@ -11,7 +11,6 @@ from zigpy.zcl import Cluster, foundation
 from zigpy.zcl.foundation import (
     BaseAttributeDefs,
     BaseCommandDefs,
-    Direction,
     ZCLAttributeDef,
     ZCLCommandDef,
 )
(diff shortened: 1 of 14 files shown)


zigpy/zcl/clusters/general.py
```diff
@@ -915,31 +875,21 @@ class AttributeDefs(BaseAttributeDefs):
...
-        off: Final = ZCLCommandDef(
-            id=0x00, schema={}, direction=Direction.Client_to_Server
-        )
-        on: Final = ZCLCommandDef(
-            id=0x01, schema={}, direction=Direction.Client_to_Server
-        )
-        toggle: Final = ZCLCommandDef(
-            id=0x02, schema={}, direction=Direction.Client_to_Server
-        )
+        off: Final = ZCLCommandDef(id=0x00, schema={})
```
uschindler, line 878: As far as I understand, this should fix the issue I have seen in #1612: without detached mode or on any other typical on/off relay switch, the change of the relay state wasn't reported correctly back to HASS?

Is my understanding correct?

PR #1618, opened 2025-07-16
[WIP] Rework initialization process

puddly:
This PR aggregates a bunch of features and fixes I'm making to clean up and speed up device joining. It will be split up.

- [x] Enable `CRITICAL` priority requests to bypass the request queue: https://github.com/zigpy/zigpy/pull/1619
- [x] Read the OTA `current_file_version` during initialization: https://github.com/zigpy/zigpy/pull/1631
- [x] Implement fast polling: we try to enable it for 30s upon joining if it is available: https://github.com/zigpy/zigpy/pull/1621
- [x] Disable ZHA handling of the poll control cluster and do it within zigpy: if we are trying to communicate with a device and the device checks in, we should ask it to start fast polling. Otherwise, we ask it to stop. https://github.com/zigpy/zha/pull/478
- [x] Fix ZCL attribute read response matching when reading from client clusters (e.g. OTA).

### DIFF
--- a/zigpy/application.py
+++ b/zigpy/application.py
@@ -72,7 +72,6 @@ class ControllerApplication(zigpy.util.ListenableMixin, abc.ABC):
         self._config = self.SCHEMA(config)
         self._dblistener = None
         self._groups = zigpy.group.Groups(self)
-        self._listeners = {}
         self._send_sequence = 0
         self._tasks: set[asyncio.Future[Any]] = set()
 
@@ -419,9 +418,15 @@ class ControllerApplication(zigpy.util.ListenableMixin, abc.ABC):
         if self._watchdog_task is not None:
             self._watchdog_task.cancel()
 
+        for task in self._tasks:
+            task.cancel()
+
         self.ota.stop_periodic_broadcasts()
         self.backups.stop_periodic_backups()
         self.topology.stop_periodic_scans()
+
+        for device in self.devices.values():
+            device.on_remove()
 
         try:
             await self.disconnect()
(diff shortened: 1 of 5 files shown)


MattWestb:
IKEA Dirigera is putting sleepers very fast in fast polling mode then joining them i think its after the joinee is have requested enddevice-timeout. = its have pull control functionality implemented and working.
Also replaying with fast poll with time out then devices is doing checkins is very good and if "missing" (network / routing / RF interference problems) sending stop its OK then its timing out.
I like implementations of real Zigbee  / PRO 2023 functionality then its possible then new devices can being more optimized in our network.

Great work **Puddly** !!!

MattWestb:
With ZHA 2025.8.0 i have no timeout (some time zigbee delivery error) then end devices is doing check in and ZHA is doing OTA version request both in production and the IKEA test system (was getting it daily in production system) so its 110% working !!
Updating to 2025.8.1 for getting IKEA Matter switch cluster working (i was missing it then not being used in production system).

One thing that can being good in the future is shall we implanting remotes buttons in  ZHA GUI like Matter have doing (one for every button = Symfonisk 2 is having 7 buttons that indicating being pressed and also the history log). I think it being one good thing for user understanding the system is working and making the system better.
Best was see how its working is adding controllers to IKEA Dirigera and then thru Matter bridge getting them in HA also some very bad tuya TH-sensor is working well that way !!

And ONE MORE TIME.
**Great work done !!!**

TheJulianJES:
This should be all done now. 🎉

PR #1621, opened 2025-07-23
Enable fast polling during joins and when requests are pending

puddly:
This should be merged alongside https://github.com/zigpy/zha/pull/478.


puddly:
This change helps a few problematic devices join for me so I think it's good to go, I just need to finish unit tests:

- IKEA PARASOLL door sensor
- Sonoff water leak sensor (SBZB-05P)

PR #1626, opened 2025-07-28
Allow changing entity metadata with quirks v2

puddly:
This will allow us to re-categorize existing entities, update translation strings, and so on:

```python
QuirkBuilder("manufacturer", "model")
.change_entity_metadata(
    endpoint_id=1,
    unique_id_suffix="something",
    new_primary=True,
)
.change_entity_metadata(
    endpoint_id=1,
    cluster_id=OnOff.cluster_id,
    new_translation_key="custom_key",
    new_fallback_name="Custom translation string",
)
```

### DIFF
--- a/tests/test_quirks_v2.py
+++ b/tests/test_quirks_v2.py
@@ -37,7 +37,8 @@ from zigpy.quirks.v2 import (
     ZCLSensorMetadata,
     add_to_registry_v2,
 )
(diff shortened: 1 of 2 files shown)


puddly:
Tested with a real device, works fine:

image
image

TheJulianJES (approved):

PR #1628, opened 2025-07-30
Implement `_find_zcl_cluster` and drop `deserialize` APIs

puddly:
For https://github.com/zigpy/zha-device-handlers/pull/4193, we need an API for devices to be able to "fix" broken ZCL cluster matching. The current API of cascading `deserialize` methods will be replaced with methods on `Device`.

### DIFF
--- a/tests/test_device.py
+++ b/tests/test_device.py
@@ -1349,7 +1349,7 @@ async def test_device_concurrency(dev: device.Device) -> None:
     ]
 
 
-async def test_duplicate_request_matching(dev: device.Device) -> None:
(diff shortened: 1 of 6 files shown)


puddly:
`deserialize` is used by only a single device: the XBee quirks. These will now use `custom_profile_packet_received`.

puddly:
Unit tests for https://github.com/zigpy/zha-device-handlers/pull/4193 pass with this PR so I think `_find_zcl_cluster` and `custom_profile_packet_received` cover all of the API edge cases for now.

The `packet_received` API isn't something I'm happy with, however. I'll break that up tomorrow.

PR #1631, opened 2025-07-30
Read OTA info for newly joining devices

puddly:
This will allow ZHA to immediately show an update entity for the device and allow us to use firmware version filtering for quirks v2.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,11 +3,12 @@
 from __future__ import annotations
 
 import asyncio
+from contextlib import contextmanager
 import copy
 import logging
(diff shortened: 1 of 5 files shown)


TheJulianJES (commented):
Nice! Thanks for finishing this.

Should there be a more explicit test checking if the `current_file_version` attribute was read during initialization? Maybe somewhat similar to `test_device::test_initialize`.

puddly:
Good point, I forgot that I never added one!

TheJulianJES (approved):
Looks good!

PR #1634, opened 2025-08-01
Active fast poll context manager

puddly:
For use with OTA, I've created an active fast poll context manager. This responds to all check-in requests for the duration of the context manager lifetime to maintain the device in fast poll mode. This theoretically will speed up OTA for some slow devices, eventually.

I'm not seeing a consistent way to *enter* fast poll mode for devices, however: on some, writing to the fast poll timeout attribute works. On others, it does not. Others do not respond to an unsolicited checkin response.

### DIFF
--- a/tests/test_device.py
+++ b/tests/test_device.py
@@ -1,6 +1,7 @@
 import asyncio
 from datetime import datetime, timezone
 import logging
+import math
 from unittest.mock import call
 
 import pytest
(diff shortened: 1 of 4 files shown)


MattWestb:
What i have seen is IKEA Dirigera sending fast pull then end devices have joining the network and getting the updated link key = the device is still in "init mode" so its not normal run mode.
You have some sniffs that i have sending you before if you like looking how they is doing but i think in the end the best is looking how end device is doing it in "real" then implementation can being very different.

By the way **GREAT WORK DONE !!!**

PR #1635, opened 2025-08-01
Reserve capacity for high priority requests

puddly:
This is a test PR that replaces our priority semaphore with a data structure that has cascading tiers. To demonstrate the problem we have right now, say we currently limit concurrency to 8 requests. This means that 8 low-priority requests can be spawned, using up all available capacity and leaving higher-priority ones starved for resources. This isn't ideal.

Instead of a global concurrency, this approach uses tiers where higher tiers have reserved capacity while also being able to use the resources of all lower tiers. So instead of an 8 request limit, we will reserved 2 slots for high priority requests, 4 for normal, and 2 for low. This limits "background" requests to 2 but also always leaves open 2 slots for high priority requests.

We would need to implement request priority within ZHA for this to actually work well. All user-initiated commands and configuration should be high priority. Polling, especially during startup, should be low. This will allow for us to use available resources to query device info without compromising interactive use.


puddly:
I'll run this for a few weeks at home to test but I don't expect any surprises.

For this to be fully useful, we need a clean way to indicate the priority of a request. Actions (e.g. `turn_on`) should take priority over `async_update`. We've already applied request priorities in most other places (OTA and device joining) so the only other consideration would be to increase the number of levels if we ever run into issues.

PR #1639, opened 2025-08-04
Auto correct cluster direction and log a warning

puddly:
Strict matching is unfortunately catching too many bad devices, some of which are caused by bad quirks. Instead of failing, we should log a warning but match regardless to figure out which quirks and devices are problematic.

### DIFF
--- a/zigpy/device.py
+++ b/zigpy/device.py
@@ -613,7 +613,10 @@ class Device(zigpy.util.LocalLogMixin, zigpy.util.ListenableMixin):
                 packet,
             )
             LOGGER.warning(
-                "Cluster 0x%04x on %r has incorrect direction (got %r for %r cluster)",
+                (
+                    "Cluster 0x%04x on %r has incorrect direction (got %r for %r cluster)."
+                    " Please report this here: https://github.com/zigpy/zigpy/issues/1640"
+                ),
                 packet.cluster_id,
                 self,
                 hdr.frame_control.direction,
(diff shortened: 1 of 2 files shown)


zigpy/device.py
```diff
@@ -595,6 +595,33 @@ def _find_zcl_cluster(
...
+            cluster = self._find_zcl_cluster_strict(
+                hdr.replace(
+                    frame_control=hdr.frame_control.replace(
+                        direction=hdr.frame_control.direction.flip()
+                    )
+                ),
+                packet,
+            )
+            LOGGER.warning(
+                "Cluster 0x%04x on %r has incorrect direction (got %r for %r cluster)",
```
TheJulianJES: Do we we wanna mention something like "Please report this in the zigpy [/quirks?] repo" (or even link to the issue tracker)?
I'm usually not a big fan of that but if we do end up removing this in a couple releases, we wanna correct all devices/quirks, and explicitly providing a call to action might increase the chances to catch everything. 😅

TheJulianJES (approved):
Looks good

zigpy/device.py
TheJulianJES: Though linking to the zigpy or quirks issue tracker might not be ideal for other users of the lib, like "Zigbee for Domoticz".

puddly: I've created a single issue with an easy-to-search name tracking this: https://github.com/zigpy/zigpy/issues/1640

PR #1646, opened 2025-08-11
Avoid `frozendict.deepfreeze`

puddly:
Seen here: https://github.com/zigpy/zha-device-handlers/pull/4251

This function makes a pretty bizarre assumption:

```python
In [1]: from frozendict import deepfreeze

In [2]: deepfreeze?
Signature: deepfreeze(o, custom_converters=None, custom_inverse_converters=None)
Docstring:
Converts the object and all the objects nested in it in its
immutable counterparts.

The conversion map is in getFreezeConversionMap().

You can register a new conversion using `register()` You can also
pass a map of custom converters with `custom_converters` and a map
of custom inverse converters with `custom_inverse_converters`,
without using `register()`.

By default, if the type is not registered and has a `__dict__`
attribute, it's converted to the `frozendict` of that `__dict__`.                <------------------

This function assumes that hashable == immutable (that is not
always true).

This function uses recursion, with all the limits of recursions in
Python.

Where is a good old tail call when you need it?
```

This fails for basically every primitive type subclass:

```python
In [3]: class MyInt(int):
   ...:     pass
   ...:

In [4]: deepfreeze(MyInt())
Out[4]: frozendict.frozendict({})
```

### DIFF
--- a/zigpy/quirks/v2/__init__.py
+++ b/zigpy/quirks/v2/__init__.py
@@ -167,7 +167,7 @@ class AddsMetadata:
     endpoint_id: int = attrs.field(default=1)
     cluster_type: ClusterType = attrs.field(default=ClusterType.Server)
     constant_attributes: frozendict[ZCLAttributeDef, Any] = attrs.field(
-        factory=frozendict
+        factory=frozendict, converter=frozendict
     )
 
     def __call__(self, device: CustomDeviceV2) -> None:
--- a/zigpy/quirks/v2/__init__.py
+++ b/zigpy/quirks/v2/__init__.py
@@ -13,7 +13,7 @@ from types import FrameType
 from typing import TYPE_CHECKING, Any, Callable
 
 import attrs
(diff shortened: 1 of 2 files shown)


zigpy/quirks/v2/__init__.py
```diff
@@ -530,7 +530,10 @@ class QuirksV2RegistryEntry:
     ] = attrs.field(factory=tuple)
     device_automation_triggers_metadata: frozendict[
         tuple[str, str], frozendict[str, str]
-    ] = attrs.field(factory=frozendict, converter=deepfreeze)
+    ] = attrs.field(
+        factory=frozendict,
+        converter=lambda d: frozendict({k: frozendict(v) for k, v in d.items()}),
```
TheJulianJES, line 535: This change already causes the `TypeError: unhashable type: 'dict'` in the one quirks test @puddly. I've only been looking at all the translation placeholder changes this whole time and was wondering why it didn't work 😅

PR #1653, opened 2025-08-14
Migrate counters to events

puddly:
This is a simple PR to migrate the ZHA event base into zigpy. We will slowly get rid of callback listeners and replace them with events.

My goal is to move availability and counters to events this release, to remove the endless logging of ZHA scanning device objects for availability state changes.


puddly:
Will need a corresponding PR to bellows, since it's the only library that uses counters.

PR #1657, opened 2025-08-19
Reduce log verbosity in zigpy.topology route/neighbor scan

sairon:
Successful topology scan can create a significant log noise printing all neighbors and routes with INFO verbosity. As it's not expected the details will be needed when casually checking the logs, lower these log messages to DEBUG verbosity.


puddly (approved):
Thanks!

PR #1659, opened 2025-08-28
Green Power Clusters and supporting Schemas

konistehrad:
Adds green power clusters to the ZCL, including a mess of supporting CommandSchemas and Schemas.

Organizational changes from original PR:

- Remove `GP` prefix, preferring instead using module-level scoping
- Split `Struct` types into `types.py`, leave `CommandSchema` types in `greenpower.py`

Questions:

- For unit tests, should I be parsing known-good ZGP frames for commands? Or since these are just structs basically, are we good for unit tests on this one?

Thanks everyone for your patience, and of course @puddly for your help!

### DIFF
--- a/zigpy/zcl/clusters/greenpower.py
+++ b/zigpy/zcl/clusters/greenpower.py
@@ -16,18 +16,19 @@ from zigpy.zcl.foundation import (
 import zigpy.zgp.types as zgptypes
 
 
+class CommissioningNotificationOptions(t.Struct):
+    application_id: zgptypes.ApplicationID
+    rx_after_tx: t.uint1_t
+    security_level: zgptypes.SecurityLevel
+    security_key_type: zgptypes.SecurityKeyType
+    security_failed: t.uint1_t
+    bidirectional_cap: t.uint1_t
+    proxy_info_present: t.uint1_t
+    _reserved: t.uint6_t
+
+
 # Figure 27
 class CommissioningNotificationSchema(foundation.CommandSchema):
-    class CommissioningNotificationOptions(t.Struct):
-        application_id: zgptypes.ApplicationID
-        rx_after_tx: t.uint1_t
-        security_level: zgptypes.SecurityLevel
-        security_key_type: zgptypes.SecurityKeyType
-        security_failed: t.uint1_t
-        bidirectional_cap: t.uint1_t
-        proxy_info_present: t.uint1_t
-        _reserved: t.uint6_t
-
     options: CommissioningNotificationOptions
     gpd_id: zgptypes.DeviceID
     frame_counter: t.uint32_t
@@ -42,12 +43,13 @@ class CommissioningNotificationSchema(foundation.CommandSchema):
     mic: t.uint32_t = StructField(requires=lambda s: s.security_failed, optional=True)
 
 
+class ResponseOptions(t.Struct):
+    application_id: zgptypes.ApplicationID
+    _reserved: t.uint5_t
+
+
 # Figure 45
 class ResponseSchema(foundation.CommandSchema):
-    class ResponseOptions(t.Struct):
-        application_id: zgptypes.ApplicationID
-        _reserved: t.uint5_t
-
     options: ResponseOptions
     temp_master_short_addr: t.uint16_t
     temp_master_tx_channel: t.uint8_t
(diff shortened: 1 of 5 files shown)


zigpy/zcl/clusters/greenpower.py
```diff
@@ -0,0 +1,407 @@
...
+    payload: t.LVBytes
+    gpp_short_addr: t.uint16_t = StructField(
+        requires=lambda s: s.proxy_info_present, optional=True
+    )
+    distance: t.uint8_t = StructField(
+        requires=lambda s: s.proxy_info_present, optional=True
+    )
+    mic: t.uint32_t = StructField(requires=lambda s: s.security_failed, optional=True)
+
+    @property
```
konistehrad, line 34: Oh right, I did have a specific question: is the use of properties like this to access the options bitfield appropriate, or should I be laying them out with `reserved` padding in the Struct itself?

zigpy/zcl/clusters/greenpower.py
```diff
@@ -18,7 +18,17 @@
 
 # Figure 27
 class CommissioningNotificationSchema(foundation.CommandSchema):
```
konistehrad, line 20: Went through the tests today and found this convenient IntStruct type, which I've migrated to instead of properties. Question though: is the `_reserved` padding field required, or can it be safely omitted?

zigpy/zcl/clusters/greenpower.py
```diff
@@ -0,0 +1,331 @@
...
+    )
+    distance: t.uint8_t = StructField(
+        requires=lambda s: s.proxy_info_present, optional=True
+    )
+    mic: t.uint32_t = StructField(requires=lambda s: s.security_failed, optional=True)
+
+
+class ResponseOptions(t.Struct):
+    application_id: zgptypes.ApplicationID
+    _reserved: t.uint5_t
```
konistehrad, line 48: Should these be `Struct` types, or `IntStruct` types? Also: if they're `IntStruct` types with fixed width, can the reserved fields be removed?

zigpy/zcl/clusters/greenpower.py
```diff
@@ -0,0 +1,331 @@
...
+        optional=True,
+    )
+    sink_nwk_addr: t.NWK = StructField(
+        requires=lambda s: not s.options.remove_gpd
+        and s.options.communication_mode
+        in (
+            zgptypes.CommunicationMode.Unicast,
+            zgptypes.CommunicationMode.UnicastLightweight,
+        ),
+        optional=True,
```
konistehrad: Is `optional` needed here if `requires` is provided?

konistehrad:
OK cool thanks for bearing with me as I get back into the project. I think this is ready to go barring the questions above.

zigpy/zcl/clusters/greenpower.py
puddly: If these fields exist entirely depending on `requires` and should never be missing, I would omit `optional`. It would otherwise allows deserialization to succeed if the serialized data is truncated.

zigpy/zcl/clusters/greenpower.py
puddly: I'd keep it a `Struct`, it's better to be explicit with what bits are used/unused. Integer structs are really only useful when the object itself is an integer and is expected to be used as one (bitwise operations, increments, and so on).

On example is the deCONZ serial protocol version. It's a 32 bit integer where specific octets have special meaning but it is also used in explicit comparisons to fixed friwmare version numbers.

zigpy/zcl/clusters/greenpower.py
```diff
@@ -0,0 +1,331 @@
...
+    ResponseSchema: Final = ResponseSchema
+    CommissioningNotificationSchema: Final = CommissioningNotificationSchema
+    NotificationResponseSchema: Final = NotificationResponseSchema
+    ProxyCommissioningModeSchema: Final = ProxyCommissioningModeSchema
+
+    class AttributeDefs(BaseAttributeDefs):
+        max_sink_table_entries: Final = ZCLAttributeDef(
+            id=0x0000,
+            type=t.uint8_t,
+            access=foundation.ZCLAttributeAccess.Read,
```
puddly: You can use `access="r"` and `access="rw"`, we use it elsewhere in the ZCL

konistehrad: Ah darn, I had actually refactored it to use the full attribute paths instead of "rw" looking at another cluster. I'll get this fixed up.

zigpy/zcl/clusters/greenpower.py
konistehrad: Checked the spec again and sure enough, removed all `optional` fields. Thanks!

konistehrad:
All comments resolved, thanks!

zigpy/zgp/types.py
```diff
@@ -0,0 +1,64 @@
+from __future__ import annotations
+
+import typing
+
+from zigpy.types import basic
+
+if typing.TYPE_CHECKING:
+    pass
```
puddly: This block can also be deleted (and the `import typing` above)

konistehrad: My bad, I had used another types file as a template for this one.

puddly: No worries at all. Thanks!

PR #1620, opened 2025-07-22
Drop Python 3.9 and 3.10 support

puddly:
3.9 will be completely end-of-life at the end of October and is currently requiring is to write backports to support. We don't need to follow this release schedule for zigpy.

PR #1660, opened 2025-09-01
Improve typing in v2 QuirkBuilder

epenet:
I noticed this whilst working on a v2 quirk for Tuya, where we loose the inheritance details in `TuyaQuirkBuilder`

See https://github.com/zigpy/zha-device-handlers/pull/4302
See https://peps.python.org/pep-0673/

As you can see in the picture below, just adding `applies_to` means we have already lost the `TuyaQuirkBuilder` reference and we end up back onto a base `QuirkBuilder`
image

### DIFF
--- a/zigpy/quirks/v2/__init__.py
+++ b/zigpy/quirks/v2/__init__.py
@@ -10,7 +10,7 @@ import inspect
 import logging
 import pathlib
 from types import FrameType
-from typing import TYPE_CHECKING, Any, Callable
+from typing import TYPE_CHECKING, Any, Callable, Self
 
 import attrs
 from frozendict import frozendict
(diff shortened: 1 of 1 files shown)


TheJulianJES (approved):
Nice, thanks! This was on my todo list as well, ever since we got the Tuya builder, but I just didn't get to it yet. Thanks for doing it!

zigpy/quirks/v2/__init__.py
```diff
@@ -14,6 +14,7 @@
 
 import attrs
 from frozendict import frozendict
+from typing_extensions import Self
```
TheJulianJES, line 17: Note for later: after https://github.com/zigpy/zigpy/pull/1620 is finally merged, we can replace this import with one from `typing` directly.

PR #1668, opened 2025-09-12
Add translation_placeholders to v2 quirks

epenet:
Added as kwarg only to avoid breaking changes (may conflict with #1652)

Needed for https://github.com/zigpy/zha/pull/533

PR #1682, opened 2025-10-10
Change `translation_placeholders` type to `frozendict`

TheJulianJES:
This fixes `translation_placeholders` not being hashable.

Follow-up to:
- https://github.com/zigpy/zigpy/pull/1668

### DIFF
--- a/zigpy/quirks/v2/__init__.py
+++ b/zigpy/quirks/v2/__init__.py
@@ -313,7 +313,9 @@ class EntityMetadata:
     attribute_initialized_from_cache: bool = attrs.field(default=True)
     unique_id_suffix: str | None = attrs.field(default=None)
     translation_key: str | None = attrs.field(default=None)
-    translation_placeholders: dict[str, str] | None = attrs.field(default=None)
+    translation_placeholders: frozendict[str, str] = attrs.field(
+        default=frozendict, converter=frozendict
+    )
     fallback_name: str = attrs.field(validator=attrs.validators.instance_of(str))
     primary: bool | None = attrs.field(default=None)
 
@@ -471,7 +473,9 @@ class ChangedEntityMetadata:
     new_primary: bool | None = attrs.field(default=None)
     new_unique_id: str | None = attrs.field(default=None)
     new_translation_key: str | None = attrs.field(default=None)
-    new_translation_placeholders: dict[str, str] | None = attrs.field(default=None)
+    new_translation_placeholders: frozendict[str, str] | None = attrs.field(
+        default=None, converter=lambda d: None if d is None else frozendict(d)
+    )
     new_device_class: (
         BinarySensorDeviceClass | NumberDeviceClass | SensorDeviceClass | None
     ) = attrs.field(default=None)
@@ -894,7 +898,7 @@ class QuirkBuilder:
                 reporting_config=reporting_config,
                 unique_id_suffix=unique_id_suffix,
                 translation_key=translation_key,
-                translation_placeholders=translation_placeholders,
+                translation_placeholders=translation_placeholders or {},
                 fallback_name=fallback_name,
                 enum=enum_class,
                 attribute_name=attribute_name,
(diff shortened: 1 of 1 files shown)


zigpy/quirks/v2/__init__.py
```diff
@@ -313,7 +313,9 @@ class EntityMetadata:
     attribute_initialized_from_cache: bool = attrs.field(default=True)
     unique_id_suffix: str | None = attrs.field(default=None)
     translation_key: str | None = attrs.field(default=None)
-    translation_placeholders: dict[str, str] | None = attrs.field(default=None)
+    translation_placeholders: frozendict[str, str] = attrs.field(
+        default=frozendict, converter=frozendict
+    )
```
TheJulianJES, line 318: This is the variant we used so far when we needed (frozen) dicts.

zigpy/quirks/v2/__init__.py
```diff
@@ -471,7 +473,9 @@ class ChangedEntityMetadata:
     new_primary: bool | None = attrs.field(default=None)
     new_unique_id: str | None = attrs.field(default=None)
     new_translation_key: str | None = attrs.field(default=None)
-    new_translation_placeholders: dict[str, str] | None = attrs.field(default=None)
+    new_translation_placeholders: frozendict[str, str] | None = attrs.field(
+        default=None, converter=lambda d: None if d is None else frozendict(d)
+    )
```
TheJulianJES, line 478: This is an alternative that would still allow the attribute being `None`. IMO, it makes sense for `new_translation_placeholders`.

We could use either variant for both options though. Like, do we want `translation_placeholders` to always be present but empty, or do we want it to be `None` if not provided.
Same for `new_translation_placeholders` here, though so far, everything has been `None` here if not provided.

Any opinions?

puddly: I think all the other fields genuinely have `None` as an option because the empty version of the datatype is invalid (e.g. `translation_key=""` or `primary=False`). We could just use a frozendict? It is a `dict` subclass and should work as a default.

puddly: Unless there would be a distinction between empty and missing for translation placeholders but I can't immediately see a difference, since we pass these to Core and can swap `{}` -> `None`.

TheJulianJES: Agreed.
I'm not sure when you'd ever want to use `change_entity_metadata` with `new_translation_placeholders`, but in that case, there would technically be a difference between `None` and `{}`. Like, making no changes (`None`) vs removing them (`{}`). Hmm.

puddly: Yeah, we should use an `UNDEFINED` singleton for all of the `change_entity_metadata` to accommodate this.

TheJulianJES: Are you fine with the PR as is for now? For a minor zigpy release?
We'd have an empty (frozen) dict for `translation_placeholders`, similar to the `constant_attributes` for the quirks v2 cluster replacement methods/metadata.

For the change entity metadata, we'd have `None` if the translation placeholders are unchanged, or whatever is provided for the dict.

puddly: Sorry, I just saw this comment!

We can address `None` / `UNDEFINED` in a future PR, for this one it's totally fine to release the translation placeholders as-is.

TheJulianJES:
Wanna check one more thing 😄

zigpy/quirks/v2/__init__.py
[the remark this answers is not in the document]
TheJulianJES: `default=frozendict` is actually not what we want, but already used for `ZCLCommandButtonMetadata . We rather want `factory=frozendict` instead, like used for `constant_attributes`.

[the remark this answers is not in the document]
TheJulianJES: Fixed with [`34f1c55` (#1682)](https://github.com/zigpy/zigpy/pull/1682/commits/34f1c551ece9829412ce5d702aba6caf5704b000) in this PR.
Fixed with https://github.com/zigpy/zigpy/pull/1685 for `constant_attributes.

PR #1686, opened 2025-10-14
Add route table to network backup

puddly:
In preparation for https://github.com/NabuCasa/silabs-firmware-builder/pull/118, I'm extending the network info JSON with a next-hop route table. While the EmberZNet structures reference route status, route age, cost, and so on, we do not actually need any of this information to restore the route table: we just need the destination node and the next-hop node.

### DIFF
--- a/zigpy/backups.py
+++ b/zigpy/backups.py
@@ -18,7 +18,7 @@ if TYPE_CHECKING:
     import zigpy.application
 
 LOGGER = logging.getLogger(__name__)
-BACKUP_FORMAT_VERSION = 1
+BACKUP_FORMAT_VERSION = 2
 
 
 @dataclasses.dataclass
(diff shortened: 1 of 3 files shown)


zigpy/backups.py
```diff
@@ -100,6 +100,13 @@ def from_dict(cls, obj: dict[str, Any]) -> NetworkBackup:
                 obj["node_info"]["version"] = None
                 version = 1
 
+            # Version 2 introduced the `route_table` and `tx_power` fields
```
TheJulianJES, line 103: The `tx_power` field doesn't seem to be included in this PR yet:
```suggestion
            # Version 2 introduced the `route_table` field
```
If you merge the other PR shortly after this, feel free to leave it in.

TheJulianJES (approved):
Looks good to me, one small note about `tx_power` still being in a comment:

PR #1687, opened 2025-10-14
Allow configuring TX power during network formation

puddly:
Depends on https://github.com/zigpy/zigpy/pull/1686.

This is a simple tweak (mainly for `bellows` that globally defaults to 8dBm). Unless an override is provided at formation time, we continue using 8dBm as a safe default. A loud warning is logged if you set it to anything beyond 10dBm to discourage copy/pasting +20dBm config from the internet.


zigpy/config/__init__.py
```diff
@@ -152,6 +155,38 @@
...
+        # Similarly, a higher TX power will cause the coordinator's messages to travel
+        # farther, which can cause it to be heard by more devices, but also "interrupt"
+        # other devices that are transmitting far away, resulting in more retries and
+        # thus less reliability.
+        #
+        # Finally, a Zigbee network requires reliable *bidirectional* communication for
+        # many operations: just because a device is able to hear the coordinator does
+        # not mean that the device can actually send a reply back.
+        #
+        # Tweak this setting at your own risk. Turning it up to 20 is
```
TheJulianJES: nit: What is "turning it up to 20"? 😄 
(-> some negative word(s) missing at the end of that line)

zigpy/config/__init__.py
```diff
@@ -152,6 +155,38 @@
...
+                limit=10,
+                message=(
+                    "Increasing the TX power will not increase the range of your"
+                    " network, devices still need to be able to respond to your"
+                    " coordinator and will do so at their default transmit power."
+                    " Changing the TX power may cause routing issues and result in end"
+                    " devices being unable to join reliably. Modify this setting at"
+                    " your own risk."
+                ),
+            ),
```
TheJulianJES, line 189: minor thing: Would it also make sense to throw in a warning about potential regulatory compliance issues in that message? But I'm not sure anyone will care about that if they're ignoring this message...

zigpy/config/__init__.py
```diff
@@ -65,6 +67,7 @@
 CONF_NWK_MAX_RETRIES = "max_retries"
 CONF_NWK_TC_ADDRESS = "tc_address"
 CONF_NWK_TC_LINK_KEY = "tc_link_key"
+CONF_NWK_TX_POWER = "tx_power"
```
TheJulianJES, line 70: note: I'm sure this is already on your radar, but zigpy-znp should probably be updated then to respect this zigpy option (if the znp specific `tx_power` option isn't set). Though the transmit power can be set at any time there IIRC, not just set during network formation (/restoring backup).

If switched over, I'd be in favor of removing the znp specific option, but I'm not sure how many people pasted something with `tx_power: 19` in their configuration and forgot about it 😅

tests/test_application.py
```diff
@@ -527,6 +527,34 @@ async def test_form_network(app):
...
+        if should_warn:
+            with pytest.warns(UserWarning, match="Increasing the TX power"):
+                if config_override is None:
+                    await app.form_network()
+                else:
+                    await app.form_network(config=config_override)
+        elif config_override is None:
+            await app.form_network()
+        else:
+            await app.form_network(config=config_override)
```
TheJulianJES: Core would like their conditional and nested tests back 😆 

(it's fine here)

TheJulianJES (approved):
also looks good in the current state, few small comments:

PR #1693, opened 2025-10-21
Allow loading backups with unsupported versions

puddly:
Downgrading ZHA after we do a major change to the backup schema causes zigpy to fail to load. This won't help us this release but will from this point forward.

Alternatively, we could keep the backup format version at 1 and perform the migration in-place, bumping it next release?

### DIFF
--- a/zigpy/backups.py
+++ b/zigpy/backups.py
@@ -91,6 +91,14 @@ class NetworkBackup(t.BaseDataclassMixin):
         elif "network_info" in obj:
             version = obj.get("version", 0)
 
+            if version > BACKUP_FORMAT_VERSION:
+                LOGGER.warning(
+                    "Network backup has version %d but current backup format"
+                    " version is %d. Downgrading is not recommended.",
+                    version,
+                    BACKUP_FORMAT_VERSION,
+                )
+
             # Version 1 introduced the `model`, `manufacturer`, and `version` fields
             if version == 0:
                 obj = copy.deepcopy(obj)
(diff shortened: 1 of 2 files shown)


zigpy/backups.py
```diff
@@ -108,8 +116,6 @@ def from_dict(cls, obj: dict[str, Any]) -> NetworkBackup:
                 obj["network_info"]["tx_power"] = None
                 version = 2
 
-            assert version == BACKUP_FORMAT_VERSION
```
TheJulianJES, line 111: At least on production HA instances, this assertion wouldn’t be triggered, right? So it’s only an issue during development really.

TheJulianJES (commented):
> Downgrading ZHA after we do a major change to the backup schema causes zigpy to fail to load.

I think most users usually downgrade via HAOS backups anyway, so it might not be a big issue.


> we could keep the backup format version at 1 and perform the migration in-place, bumping it next release?

So backups taken with the new zigpy version would still have version 1, but already contain routing and tx info? If so, I‘d probably go with what we have in this PR.

zigpy/backups.py
puddly: Does Core disable Python assertions?

TheJulianJES: I always assumed so 😅
Let me check in a few minutes

TheJulianJES: Ok hmm, Core does not disable assertions. Thought at least HAOS would do that, but I'm finding issues from HAOS users reporting `AssertionError`, so not sure where I got the idea..

puddly: I think it was discussed at some point to enable the highest level of Python runtime optimization (the one that strips asserts and docstrings) but I don't think that was ever done.

PR #1694, opened 2025-10-21
Downgrade network backup version from 2 to 1

puddly:
We haven't made a zigpy release yet that incorporates the new keys and major version change so this is just out of an abundance of caution (and to make my life easier, since I downgrade quite a lot 😄).

Next month we can revert this PR and properly bump the backup format version.

### DIFF
--- a/zigpy/backups.py
+++ b/zigpy/backups.py
@@ -110,8 +110,6 @@ class NetworkBackup(t.BaseDataclassMixin):
             # versions of zigpy. The backup format version will be incremented from 1
             # to 2 in 2025.12.0, to allow one release cycle to pass before making a
             # backwards-incompatible breaking change.
-            #
-            # Version 2 introduced the `route_table` and `tx_power` fields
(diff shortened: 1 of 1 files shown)


TheJulianJES (approved):

zigpy/backups.py
```diff
@@ -106,13 +106,19 @@ def from_dict(cls, obj: dict[str, Any]) -> NetworkBackup:
                 obj["node_info"]["version"] = None
                 version = 1
 
+            # Incrementing the version number raised an assertion error in older
+            # versions of zigpy. The backup format version will be incremented from 1
+            # to 2 in 2025.12.0, to allow one release cycle to pass before making a
+            # backwards-incompatible breaking change.
+            #
             # Version 2 introduced the `route_table` and `tx_power` fields
```
TheJulianJES: nit: do we wanna keep this comment, even if it's not fully true right now?

PR #1696, opened 2025-10-27
Replace pyserial-asyncio with pyserial-asyncio-fast

### LINKED ISSUE #1604: Blocking I/O - consider dropping `pyserial-asyncio` dependency
`pyserial-asyncio` does blocking I/O and should be replaced by `pyserial-asyncio-fast`
See https://github.com/home-assistant/core/pull/116635

epenet:
Fixes #1604

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,8 +21,7 @@ dependencies = [
     "cryptography",
     "voluptuous",
     "jsonschema",
-    'pyserial-asyncio; platform_system!="Windows"',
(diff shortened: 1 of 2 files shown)


TheJulianJES (commented):
Seems like "Zigbee for Domoticz" already uses `pyserial-asyncio-fast` for zigpy, so we shouldn't cause any issues there: https://github.com/search?q=owner%3Azigbeefordomoticz%20pyserial-asyncio&type=code

PR #1697, opened 2025-10-27
Remove `incorrect direction` warning

puddly:
We've collected enough reports in https://github.com/zigpy/zigpy/issues/1640 to fix all of the quirks and to get a good idea of how many real-world devices do this poorly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,8 +10,6 @@ jobs:
     with:
       CODE_FOLDER: zigpy
       CACHE_VERSION: 3
-      PRE_COMMIT_CACHE_PATH: ~/.cache/pre-commit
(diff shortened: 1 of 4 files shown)


TheJulianJES (commented):
Agreed.
Since we keep the message at debug level, doesn't it also make sense to keep the test? Or is this so trivial we just don't want the test for that?

puddly:
We don't test any other debug logging so I just thought it wasn't worth the extra code.

TheJulianJES:
Yeah, ok. We don't really have any explicit tests for the matching/flipping logic then, right? At least none were added with https://github.com/zigpy/zigpy/pull/1639/files, other than the one that mainly checks for the warning to be present.

puddly:
Ooo, good point. I'll see about adding one.

PR #1703, opened 2025-11-05
API for setting the TX power at runtime

puddly:
The country code will be pulled from the zigpy config dict at network startup, which ZHA will update with the country code supplied by Home Assistant. Radio libraries implementing this API need to supply `_set_tx_power`.

- bellows: https://github.com/zigpy/bellows/pull/698
- zigpy-znp: https://github.com/zigpy/zigpy-znp/pull/266


puddly:
This requires a small patch to Core and ZHA to test out. Works fine with https://github.com/zigpy/bellows/pull/698. Unless you've modified TX power config, things work automatically with supported adapters.

PR #1711, opened 2025-11-25
Add exposed features to v2 quirks

TheJulianJES:
After https://github.com/zigpy/zha/pull/575, there'll be another PR to also parse these quirks v2 exposed features in ZHA, so it might be nice to already get this into zigpy.

### DIFF
--- a/tests/test_quirks_v2.py
+++ b/tests/test_quirks_v2.py
@@ -28,6 +28,7 @@ from zigpy.quirks.v2 import (
     EntityMetadata,
     EntityPlatform,
     EntityType,
+    ExposesFeatureMetadata,
     FirmwareVersionFilterMetadata,
     NumberMetadata,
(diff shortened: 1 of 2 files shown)


puddly (approved):
I think this is a fine approach to use for the short term. It would be nice to be able to pass config to these features...at which point you're basically creating entity objects directly, just with a string API in the middle 😄.

TheJulianJES:
Ha, that's a nice idea, actually. Whatever we do for that in quirks v2 will be backwards compatible though.

For now, I want to use this for the Frient sirens to clean up non-working select entities and remove siren entity features. It's also useful for IAS tamper sensors, so we don't have to have to duplicate that in quirks v2 everywhere.
We'd finally have IAS tamper support directly in ZHA (requiring an opt-in by the quirk). 😅

PR #1712, opened 2025-11-25
Allow specifying `config` for exposed features

TheJulianJES:
This PR would also allow optionally specifying config options for exposed features on v2 quirks.

Follow-up to:
- https://github.com/zigpy/zigpy/pull/1711

I'm not sure if this is something we want/need to merge in the current state though. Otherwise, it's so simple, that we might as well(?) 😄


puddly (approved):

TheJulianJES:
Alright, let's go with this then for now. I can't see this becoming much different and it can be helpful to have it in already.

PR #1715, opened 2025-11-29
fix: catch exceptions for checkin_response

elupus:
Add callback wrapper to application controller
that catches exceptions heading for event loop
and logs these with device context

Remove the now unneeded handling of this
in CallbackListener.

### DIFF
--- a/tests/test_application.py
+++ b/tests/test_application.py
@@ -1797,3 +1797,58 @@ async def test_shutdown_device_remove_fails(app, ieee, caplog):
 
     assert "Failed to remove device" in caplog.text
     assert "Boom!" in caplog.text
+
+
+async def test_callback_wrapping(
+    app: zigpy.application.ControllerApplication, ieee, caplog
+) -> None:
+    # The default is True
+    dev = app.add_device(ieee, 0x1234)
+
+    def _callback():
+        raise ValueError("Boom!")
+
+    callback = app.wrap_callback(dev, _callback)
+    with caplog.at_level(logging.WARNING):
+        callback()
+
+    assert caplog.record_tuples == [
+        (
+            "zigpy.application",
+            logging.WARNING,
+            (
+                "Device <Device model=None manuf=None nwk=0x1234 "
+                "ieee=07:06:05:04:03:02:01:00 is_initialized=False> "
+                "callback failed - ValueError('Boom!')"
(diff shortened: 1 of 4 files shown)


MattWestb:
This fix is needed for problems with device checkins and also other cases like end device is offline or having problems that is not possible see what device is having the problems only "some" ZHA devices as its today.

tests/test_application.py
```diff
@@ -1797,3 +1797,58 @@ async def test_shutdown_device_remove_fails(app, ieee, caplog):
 
     assert "Failed to remove device" in caplog.text
     assert "Boom!" in caplog.text
+
+
+async def test_callback_wrapping(
+    app: zigpy.application.ControllerApplication, ieee, caplog
+) -> None:
+    # The default is True
```
TheJulianJES, line 1805: nit: stale comment
```suggestion
```

tests/test_application.py
```diff
@@ -1797,3 +1797,58 @@ async def test_shutdown_device_remove_fails(app, ieee, caplog):
...
+                "callback failed - ValueError('Boom!')"
+            ),
+        ),
+    ]
+
+
+async def test_callback_wrapping_async(
+    app: zigpy.application.ControllerApplication, ieee, caplog
+) -> None:
+    # The default is True
```
TheJulianJES, line 1831: nit: stale comment
```suggestion
```

zigpy/application.py
```diff
@@ -95,6 +97,35 @@ def __init__(self, config: dict) -> None:
             "request_priority", default=t.PacketPriority.NORMAL
         )
 
+    def wrap_callback(
+        self, src: zigpy.device.Device | zigpy.listeners.ANY_DEVICE, callback: _C
+    ) -> _C:
```
TheJulianJES, line 102: So this likely doesn't really matter with how we use it, but for a sync callback, the returned callable can now also return `None` if an exception occurs. So, the return type is not really `_C` anymore.

For async callbacks, the returned wrapper is a sync function that fires the task and returns `None`, so also not `_C` at all.

elupus: Hmm. I suppose it might be possible to unwrap the callable into its basic elements like the register function is defined and add the None return. We could probably also just have it be a none return.

PR #1718, opened 2025-12-04
Define WWAH cluster

puddly:
These commands and attributes are pulled from Simplicity SDK.

Interestingly, this seems to be an Amazon-led initiative: the manufacturer code is `0x1217` for all commands and attributes. Most of these are for hardening network operations and likely require the coordinator to advertise supporting this cluster.

This PR depends on the attribute read/write rewrite PR, since it moves `manufacturer_code_override: Final[t.uint16_t] = 0x1217` out of quirks and into the ZCL.

### DIFF
--- a/zigpy/zcl/clusters/wwah.py
+++ b/zigpy/zcl/clusters/wwah.py
@@ -156,7 +156,6 @@ class WorksWithAllHubs(CustomCluster):
             schema={},
         )
 
-        # Enable WWAH App Event retry algorithm.
         enable_wwah_app_event_retry_algorithm: Final = ZCLCommandDef(
             id=0x04,
             schema={
@@ -167,7 +166,6 @@ class WorksWithAllHubs(CustomCluster):
             },
         )
 
-        # Disable WWAH App Event retry algorithm.
         disable_wwah_app_event_retry_algorithm: Final = ZCLCommandDef(
             id=0x05,
             schema={},
@@ -179,7 +177,6 @@ class WorksWithAllHubs(CustomCluster):
             schema={},
         )
 
-        # Enable WWAH rejoin algorithm.
         enable_wwah_rejoin_algorithm: Final = ZCLCommandDef(
             id=0x07,
             schema={
@@ -191,7 +188,6 @@ class WorksWithAllHubs(CustomCluster):
             },
         )
 
-        # Disable WWAH rejoin algorithm.
         disable_wwah_rejoin_algorithm: Final = ZCLCommandDef(
             id=0x08,
             schema={},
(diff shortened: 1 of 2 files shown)


zigpy/zcl/clusters/wwah.py
```diff
@@ -0,0 +1,509 @@
...
+            manufacturer_code=0x1217,
+        )
+
+        survey_beacons_response: Final = ZCLCommandDef(
+            id=0x09,
+            schema={"beacon": t.LVList[WwahBeaconSurvey, t.uint8_t]},
+            manufacturer_code=0x1217,
+        )
+
+        # This extra command's integration into the spec is being discussed. Do not remove if updating XML
```
TheJulianJES: nit: This comment comes from the SDK, but it might be nice to update slightly, so the "Do not remove if updating XML" part is a bit clearer. Someone could assume we're using XML elsewhere 😄

zigpy/zcl/clusters/wwah.py
```diff
@@ -0,0 +1,509 @@
...
+            type=t.uint16_t,
+            access="r",
+            mandatory=False,
+            manufacturer_code=0x1217,
+        )
+        cluster_revision: Final = ZCL_CLUSTER_REVISION_ATTR
+        reporting_status: Final = ZCL_REPORTING_STATUS_ATTR
+
+    class ServerCommandDefs(BaseCommandDefs):
+        # Client-to-Server commands (sent by client, received by server)
```
TheJulianJES, line 199: nit: We generally also never had comments describing what `ServerCommandDefs` and `ClientCommandDefs` are, but if it helps, ok.

TheJulianJES (approved):
Looks fine

zigpy/zcl/clusters/wwah.py
puddly: Thanks! I've clarified the comment.

puddly:
We can include these definitions into the next zigpy release, I'd rather not change too much at once given how large our release will be.

zigpy/zcl/clusters/wwah.py
```diff
@@ -0,0 +1,509 @@
...
+        )
+        ota_max_offline_duration: Final = ZCLAttributeDef(
+            id=0x0013,
+            type=t.uint16_t,
+            access="r",
+            mandatory=False,
+            manufacturer_code=0x1217,
+        )
+        cluster_revision: Final = ZCL_CLUSTER_REVISION_ATTR
+        reporting_status: Final = ZCL_REPORTING_STATUS_ATTR
```
TheJulianJES, line 196: Note for later: As the WWAH cluster is inside the manufacturer-specific range and neither `is_manufacturer_specific` nor `manufacturer_code` are defined here, `_get_effective_manufacturer_code` would return the manufacturer code of the device. That's likely not expected.

PR #1719, opened 2025-12-04
Fix reading/writing attributes with mixed manufacturer codes

puddly:
This PR introduces a few small API tweaks.

- Attribute and command definitions now accept `manufacturer_code`, allowing you to define commands and attributes whose IDs "collide". Unless we heavily rely on base class implementations and inheritance to swap `manufacturer_code_override`, we could consider moving to explicitly specifying the manufacturer code for cluster commands and attributes?
- This allows attribute reads and writes (TODO: attribute report configuration) to be bundled by manufacturer ID.
- This unfortunately changes the attribute cache format: the attribute ID is no longer enough to uniquely look up an attribute. To fix this, we allow passing around attribute definition objects (or strings). Reading/writing attributes by ID will be soft deprecated and we should migrate ZHA and quirks away from using it. Ambiguous lookups will now fail.

---

Quirks poke around in zigpy internals and unfortunately make this migration more painful than it should be. I think this calls for a few new APIs to be added to zigpy to address limitations:
- Virtual ZCL attributes being added to `_attr_cache`.
- `def _update_attribute` overrides
- `def update_attribute` overrides
- Explicit calls to `self._update_attribute`

All of these can I think be grouped into a few broad categories:
- Virtual attributes to make ZCL entities out of non-ZCL attributes that can be slightly remapped to look ZCL-ish.
- Virtual clusters to map non-ZCL functionality into a "fake" ZCL cluster that is turned into a ZCL entity.
- Virtual attributes to pass to the ZHA library, which implements custom entities (`feeder_acn001.py`).

We can't solve all of these uses today but maybe we can provide a public API (for now, mapping back to the ZCL cache 😄) that makes these uses more explicit?


puddly:
I've done a pretty hefty refactor and found I think a pretty simple way to make things work. The attribute cache object has been changed from a dict and into a `AttributeCache` that exports _some_ interfaces to make it backwards compatible with existing code.

We have a ton of new events:
- `AttributeReadEvent`
- `AttributeReportedEvent`
- `AttributeWrittenEvent`
- `AttributeUpdatedEvent`
- `AttributeUnsupportedEvent`
- `AttributeReportingConfiguredEvent`
- `AttributeClearedEvent`

`AttributeUpdatedEvent` is the most interesting one. This is emitted when attributes are changed by something other than reads, writes, or reports: namely, quirks. This lets us explicitly deal with quirk-initiated attribute changes and store them in a simple key-value `AttributeCache._legacy_cache`.

One other major change is that appdb no longer loads just `model` and `manufacturer` before matching quirks: *all* attributes are loaded, then we match quirks, then we load them _again_. This should in theory give quirks more context and let them filter better at creation time. The purpose of this change is to allow us to run data migrations. We can match up as many "legacy" cache entries as possible and update the manufacturer code correctly for as much data as possible.

I have unit tests passing for zigpy + ZHA + quirks in unison but the changeset is quite expansive. I'd like to shrink it down first but I think this approach may work and may let us migrate away from these "manufacturer code" hacks once and for all.

tests/databases/zigbee_puddly2.db
TheJulianJES, line 1: Probably fine, but this file is 2.84 MB, so increases the repo size a bit.

zigpy/appdb.py
```diff
@@ -205,6 +227,17 @@ async def shutdown(self) -> None:
...
 
+    def register_cluster_events(self, cluster) -> None:
+        cluster.on_event(AttributeReadEvent.event_type, self.on_attribute_read)
+        cluster.on_event(AttributeReportedEvent.event_type, self.on_attribute_reported)
+        cluster.on_event(AttributeUpdatedEvent.event_type, self.on_attribute_updated)
+        cluster.on_event(AttributeWrittenEvent.event_type, self.on_attribute_written)
+        cluster.on_event(
+            AttributeUnsupportedEvent.event_type, self.on_attribute_unsupported
+        )
+        # cluster.on_event(AttributeReportingConfiguredEvent.event_type, self.on_attribute_reporting_configured)
```
TheJulianJES: Since the `AttributeReportingConfiguredEvent` seems to be finished and fired for initialization already, does it make sense to already save this to the DB with the v14 schema, or do we want to do that for another release? (and likely another schema?)

Hmm, but I guess we don't fire it for manual bindings and unbindings yet..
I guess it's fine to keep as a comment here for future reference when implementing.

zigpy/appdb.py
```diff
@@ -275,82 +308,31 @@ async def _save_device_relays(self, ieee: t.EUI64, relays: t.Relays | None) -> N
...
-        self.enqueue(
-            "_unsupported_attribute_added",
-            cluster.endpoint.device.ieee,
-            cluster.endpoint.endpoint_id,
-            cluster.cluster_type,
-            cluster.cluster_id,
-            attrid,
-        )
+    def on_attribute_unsupported(self, event: AttributeUnsupportedEvent) -> None:
+        self.enqueue("_unsupported_attribute_added", event)
```
TheJulianJES: Side note pt. 1: `on_attribute_read`, `on_attribute_reported`, `on_attribute_updated`, and `on_attribute_written` are all in the same place below.

 `on_attribute_unsupported` is a bit further above here. `on_attribute_cleared` is even further below the others.

Does it make sense to move these two methods to the other ones to have all the "event methods" in one place?

zigpy/appdb.py
```diff
@@ -555,59 +543,75 @@ async def _save_unsupported_attributes(self, ep: Endpoint) -> None:
...
-    async def _clear_attribute(
-        self,
-        ieee: t.EUI64,
-        endpoint_id: int,
-        cluster_type: ClusterType,
-        cluster_id: int,
-        attrid: int,
-    ) -> None:
+    def on_attribute_cleared(self, event: AttributeClearedEvent) -> None:
+        self.enqueue("_clear_attribute", event)
```
TheJulianJES, line 584: Side note pt. 2: `on_attribute_cleared` is here, a bit off from the other event methods.

zigpy/zcl/__init__.py
```diff
@@ -547,295 +746,545 @@ def handle_cluster_general_request(
...
+                for record in result[0]:
+                    attr_def = potential_attributes[record.attrid]
+
+                    if record.status == foundation.Status.SUCCESS:
+                        if record.value.value is None:
+                            # TODO: remove this workaround when `LocalDataCluster` and
+                            # `_VALID_ATTRIBUTES` are removed from quirks. There is no
+                            # way for `value` to actually be `None` when read from a
+                            # real device.
+                            value = record.value.value
```
TheJulianJES: Heh, yeah. Some thoughts on this:

### Simplification?

Side note: The check with `record.value.value is None` and then essentially setting `value = None` could be a bit simplified/clarified (with just setting `value = None` directly). Or do you think this is clearer?

(Could also use the all beloved walrus operator but that's kind of useless here when the record value is guaranteed to be `None` anyway..)

---

### Background for reference

The workaround just exists, so virtual Tuya attributes aren't marked as unsupported when an initial read is done on them. Before this existed, you had to restart HA after some data arrived. Now, we properly create the entities in HA and they will show as unknown or unavailable (I forgot what exactly), until the first data arrives.

Doing that is actually better than what we do for some Aqara devices, where we use some default data, which might not be entirely accurate.

### Removal

We'll only be able to get rid of this workaround when ZHA entities no longer directly access ZCL attribute cache, but some kind of other intermediate layer (i.e. "directly access Tuya + custom Aqara attributes" that aren't ZCL ones). Only then can we (finally) get rid of `LocalDataCluster` (hopefully).

For actual ZCL attributes, we could have some kind of "default converter" to this intermediate layer that's persisted in the DB and then accessed by ZHA (entities). But this can al be discussed at a later date.

zigpy/appdb.py
```diff
@@ -275,82 +308,31 @@ async def _save_device_relays(self, ieee: t.EUI64, relays: t.Relays | None) -> N
...
+        await self.execute(
+            q,
+            {
+                "ieee": event.device_ieee,
+                "endpoint_id": event.endpoint_id,
+                "cluster_type": event.cluster_type,
+                "cluster_id": event.cluster_id,
+                "attr_id": event.attribute_id,
+                "manufacturer_code": event.manufacturer_code,
+                "status": Status.UNSUPPORTED_ATTRIBUTE,
```
TheJulianJES: Interesting choice with reusing the ZCL status codes for our DB, but I think this is fine. 👍

zigpy/appdb.py
TheJulianJES: These log messages are for similar-ish things, so should they be streamlined a bit? (one includes nwk address, other doesn't, and so on) (likely also applies to the ones further up then)

TheJulianJES (commented):
I've not had a completely detailed look at everything yet, but have a very good overview now. This is awesome! It's much more elegant than I expected, given the requirement of keeping backwards compatibility.

Couple of small comments already:

zigpy/zcl/__init__.py
```diff
@@ -547,295 +746,545 @@ def handle_cluster_general_request(
...
                     value = attr.value.value
-                self._update_attribute(attr.attrid, value)
+                else:
+                    attr_name = attr_def.name
+                    value = attr_def.type(attr.value.value)
+
+                # Call _update_attribute for backwards compatibility with quirks,
+                # but suppress events since we emit AttributeReportedEvent below
+                with suppress_events():
+                    self._update_attribute(attr.attrid, value)
```
TheJulianJES: For the comment: It's also still needed to populate `self._attr_cache` properly for attribute reports and reading attributes, right? `self._attr_cache` is only updated directly when having successfully written attributes.

zigpy/appdb.py
puddly: Ah. I kept it while subscribing to the other events but I will remove it, since it's unnecessary. We can save persisting the binding and reporting tables for the v15 schema.

zigpy/zcl/__init__.py
puddly: `_update_attribute` itself updates `_attr_cache`. This setup is a tiny bit convoluted because we want to emit the correct events in all cases, even manual `_update_attribute` calls (which emit `AttributeUpdatedEvent`). Once `_update_attribute` calls are removed from quirks, we can get rid of it and move the attribute update event emission into the correct spots.

TheJulianJES: Yeah, I just meant that the comment implies the `_update_attribute` call is only needed for backwards compatibility. But even if we don't need that anymore, we still need to update `_attr_cache`, which is done in `_update_attribute` at the moment, so we can't just remove the entire block then.

puddly: Ahh, now I understand. I've updated the comment to explain the situation more clearly.

TheJulianJES: I think that's better. Thanks!

TheJulianJES (approved):
This looks awesome! (I've not yet tried this on my main setup though 😅)

PR #1721, opened 2025-12-11
Add more sensor entity metadata fields to `change_entity_metadata`

puddly:
`suggested_display_precision` is really the only one I'm interested in using but I've added the rest as well.


TheJulianJES (approved):

TheJulianJES:
I do kind of wish we'd require specifying the platform for `change_entity_metadata`. That would make it much easier to generate translation keys for new ones specified in `change_entity_metadata`, without the need of discovering all entities.
It should also be clearer in quirks to quickly see what kind of entity/platform is affected.
And lastly, it might also help avoid some accidental matching of multiple similar entities that are different per platform.

PR #1714, opened 2025-11-29
fix: avoid waiting on checkin response response

elupus:
Avoid waiting on the Default_Response for a checkin_response request. There is no use for the reply data.

Ensure we set disable_default_response on the frames whenever we are not expecting a reply. It does not matter if this is a client or server request, what matters is if we are going to be waiting for a reply.

This avoid these type of errors:

```
Loggare: homeassistant
Källa: /usr/src/homeassistant/homeassistant/runner.py:238
Inträffade först: 10 november 2025 kl. 07:47:15 (95 förekomster)
Senast loggade: 28 november 2025 kl. 21:11:41

Error doing job: Task exception was never retrieved (None)
Traceback (most recent call last):
  File "/usr/local/lib/python3.13/site-packages/zigpy/device.py", line 593, in request
    return await future
           ^^^^^^^^^^^^
asyncio.exceptions.CancelledError

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/lib/python3.13/site-packages/zigpy/device.py", line 359, in poll_control_checkin_callback
    await poll_control.checkin_response(
    ...<3 lines>...
    )
  File "/usr/local/lib/python3.13/site-packages/zigpy/zcl/__init__.py", line 405, in request
    return await self._endpoint.request(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<9 lines>...
    )
    ^
  File "/usr/local/lib/python3.13/site-packages/zigpy/endpoint.py", line 233, in request
    return await self.device.request(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<11 lines>...
    )
    ^
  File "/usr/local/lib/python3.13/site-packages/zigpy/device.py", line 592, in request
    async with asyncio_timeout(timeout):
               ~~~~~~~~~~~~~~~^^^^^^^^^
  File "/usr/local/lib/python3.13/asyncio/timeouts.py", line 116, in __aexit__
    raise TimeoutError from exc_val
TimeoutError
```

Note. We should likely also catch these errors and add device iee context so we can see what fails.

PR #1728, opened 2025-12-18
Allow for the default response bit to be explicitly controlled

puddly:
https://github.com/zigpy/zigpy/pull/1714 affected quirks unit tests. As an alternative implementation, I think it would be better to explicitly pass an override for the `disable_default_response` bit. This allows the exact behavior to be controlled by the caller as opposed to using indirect methods (e.g. `not self.is_client` or `expect_reply`).


TheJulianJES (approved):
🎉

PR #1730, opened 2025-12-18
Correct typing for callback wrapping

elupus:
Fixes for things spotted in post merge review: https://github.com/zigpy/zigpy/pull/1715#pullrequestreview-3594623577


puddly (approved):

TheJulianJES (approved):
Thanks!

(This does slightly change the behavior for sync callbacks by not returning on success, but as we never use that anywhere, it's fine IMO.)

PR #1734, opened 2025-12-29
Add Telink OTA parsing support and propagate telinkEncryption flag

Oniums:
### Goal
add first-class handling for Telink-encrypted Zigbee OTA images so they can be fetched, parsed, and validated without failing the generic Zigbee OTA parser.
### Key changes
- Added telink_encryption flag to OTA metadata and propagate it from the Z2M index (field telinkEncryption).
- In metadata fetch(), when telink_encryption is set, attempt Telink-specific parsing (TelinkOTAImage) before falling back to the generic parse_ota_image.
- Introduced Telink container parsing in image.py: Telink subelements carry tag_id, tag_length, tag_info, and data, differing from the standard subelement shape (which only has tag_id+ length-prefixed data). Telink images now deserialize to TelinkOTAImage.
### Behavior
- For Z2M-provided firmware entries that set telinkEncryption, the downloaded blob is parsed with the Telink layout; if successful, the Telink image is returned. If parsing fails, we fall back to the standard parser.
- Checksum/size validation still runs before parsing. This assumes the index’s checksum/size corresponds to the encrypted blob; if upstream moves to decrypted values, the validation order will need revisiting.
### Rationale
Telink OTA files use a distinct container format; the generic Zigbee OTA parser cannot interpret the extra tag_info/length fields, leading to failures. Dedicated parsing enables these images to be consumed like other vendors’ firmware.

### Additionally
We encountered the same issue in Z2M and applied a similar fix there. 
Z2M PR:  https://github.com/Koenkk/zigbee-herdsman-converters/pull/9984#issuecomment-3288777252

### DIFF
--- a/zigpy/ota/image.py
+++ b/zigpy/ota/image.py
@@ -4,9 +4,9 @@ from __future__ import annotations
 
 import hashlib
 import logging
-from typing import Self
 
 import attr
+from typing_extensions import Self
 
 import zigpy.types as t
 
@@ -250,6 +250,51 @@ class HueSBLOTAImage(BaseOTAImage):
 
         return cls(header=header, data=firmware), data[header.image_size :]
 
+@attr.s
+class TelinkSubElement:
+    tag_id = attr.ib()
+    tag_length = attr.ib()
+    tag_info = attr.ib()
+    data = attr.ib()
+
+    def serialize(self) -> bytes:
+        res = self.tag_id.serialize()
+        res += t.uint32_t(self.tag_length).serialize()
+        res += t.uint16_t(self.tag_info).serialize()
+        res += self.data
+        return res
+
+    @classmethod
+    def deserialize(cls, data: bytes) -> tuple["TelinkSubElement", bytes]:
+        tag_id, data = ElementTagId.deserialize(data)
+        tag_length, data = t.uint32_t.deserialize(data)
+        tag_info, data = t.uint16_t.deserialize(data)
+        if len(data) < tag_length:
+            raise ValueError("Not enough data for Telink element")
+        data_bytes, data = data[:tag_length], data[tag_length:]
+        return cls(tag_id=tag_id, tag_length=tag_length, tag_info=tag_info, data=data_bytes), data
+
+@attr.s
+class TelinkOTAImage(BaseOTAImage):
+    header = attr.ib()
+    subelements = attr.ib(factory=list)
+
+    def serialize(self) -> bytes:
+        res = self.header.serialize()
+        for sub in self.subelements:
+            res += sub.serialize()
+        return res
+
+    @classmethod
+    def deserialize(cls, data: bytes) -> tuple["TelinkOTAImage", bytes]:
+        header, data = OTAImageHeader.deserialize(data)
+        elements_len = header.image_size - header.header_length
+        subelements = []
+        element_data, data = data[:elements_len], data[elements_len:]
+        while element_data:
+            element, element_data = TelinkSubElement.deserialize(element_data)
+            subelements.append(element)
+        return cls(header=header, subelements=subelements), data
 
 def parse_ota_image(data: bytes) -> tuple[BaseOTAImage, bytes]:
     """Attempts to extract any known OTA image type from data. Does not validate firmware."""
--- a/zigpy/ota/providers.py
+++ b/zigpy/ota/providers.py
@@ -21,7 +21,7 @@ import voluptuous as vol
 
 import zigpy.config
 from zigpy.ota import json_schemas
-from zigpy.ota.image import BaseOTAImage, parse_ota_image
+from zigpy.ota.image import BaseOTAImage, TelinkOTAImage, parse_ota_image
 import zigpy.types as t
 import zigpy.util
 
@@ -90,6 +90,8 @@ class BaseOtaImageMetadata(t.BaseDataclassMixin):
     max_current_file_version: int | None = None
     specificity: int | None = None
 
+    telink_encryption: bool | None = None
+
     source: str = "Unknown"
 
     async def _fetch(self) -> bytes:
@@ -116,7 +118,12 @@ class BaseOtaImageMetadata(t.BaseDataclassMixin):
     async def fetch(self) -> BaseOTAImage:
         data = await self._fetch()
         await self._validate(data)
-
+        if self.telink_encryption:
+            try:
+                image, _ = TelinkOTAImage.deserialize(data)
+                return image
+            except Exception:
+                pass
         image, _ = parse_ota_image(data)
         return image
 
@@ -598,6 +605,7 @@ class BaseZ2MProvider(BaseOtaProvider):
                 "min_hardware_version": fw.get("hardwareVersionMin"),
                 "max_hardware_version": fw.get("hardwareVersionMax"),
                 "changelog": fw.get("releaseNotes"),  # Changelog is short
+                "telink_encryption": fw.get("telinkEncryption"),
                 "source": "",  # Set in a subclass
             }
 


puddly:
Thanks for the PR.

Could you explain a bit more about why this needs explicit handling by zigpy? Checksum validation is very welcome, we never want to send corrupt images, but why do we need the `telinkEncrypted` flag explicitly added to the OTA metadata format?

Oniums:
> Thanks for the PR.
> 
> Could you explain a bit more about why this needs explicit handling by zigpy? Checksum validation is very welcome, we never want to send corrupt images, but why do we need the flag explicitly added to the OTA metadata format?`telinkEncrypted`

Because image.py line 179 and basic.py line 775 perform length checks on the OTA image, a standard Zigbee OTA image parses correctly.

However, the Telink chip’s OTA image includes a special Tag Info field, so each element is laid out as Tag Id (2 bytes) + Tag Length (4 bytes) + Tag Info (2 bytes) + Data. During the length check, Tag Info gets counted as part of the Data, leaving an extra 2 bytes at the end and triggering an error.

Here is the error log from Home Assistant when I try to upgrade a device using a Telink OTA firmware image：
lQLPKIW19c8yDf3NA23NB26wjfCmYX-m_9gIm6YWr4nYAA_1902_877

There’s a similar issue in Z2M, with a more detailed explanation there.
https://github.com/Koenkk/zigbee-herdsman-converters/issues/9963#issue-3397137524

puddly:
Can you attach the OTA file you are testing with?

There are nearly 100 known OTA images for Telink devices: https://github.com/Koenkk/zigbee-OTA/blob/9009f7e68f43c4b5f4adf1d0cf610b7c611205b4/index-stackinfo.json. Gledopto, LEDVANCE, Innr, ThirdReality, and so on. None of them use a custom format. Is there a reason why these OTA files do not follow the specification? As far as I understand, the length of this custom "tag info" field just needs to be added to the total tag length, since the tag length is 2 bytes too short.

Oniums:
[snzb-01m_v1.0.5.ota.zip](https://github.com/user-attachments/files/24389427/snzb-01m_v1.0.5.ota.zip)

- The archive above contains our packaged Telink OTA firmware. For the OTA files that don’t use the custom format, I believe their Tag ID is not FF00.
- In Telink’s OTA packaging tool, enabling content encryption changes the Tag ID to FF00 and inserts TagInfo into the element field to improve firmware security.
- Adding the TagInfo length into tag length causes two issues: (1) the Zigbee Cluster Library explicitly states tag length is the length of the data segment; (2) Telink’s software SDK also does not count TagInfo in tag length when processing incoming OTA packets. So that approach won’t work.
image

puddly:
I think manually flagging `telink_encryption` is not a good approach here, it is fragile. We can autodetect the `0xF000` subelement and make a special case for Telink firmwares:

```diff
diff --git a/zigpy/ota/image.py b/zigpy/ota/image.py
index 84683ddd..4701f8a0 100644
--- a/zigpy/ota/image.py
+++ b/zigpy/ota/image.py
@@ -251,6 +251,100 @@ class HueSBLOTAImage(BaseOTAImage):
         return cls(header=header, data=firmware), data[header.image_size :]
 
 
+@attr.s
+class TelinkEncryptedSubElement:
+    tag_id: ElementTagId = attr.ib(default=None)  # Always 0xF000
+    tag_info: t.uint16_t = attr.ib(default=None)
+    data: bytes = attr.ib(default=None)
+
+    def __repr__(self) -> str:
+        if len(self.data) > 32:
+            data = self.data[:25].hex() + "..." + self.data[-7:].hex()
+        else:
+            data = self.data.hex()
+
+        return (
+            f"<{self.__class__.__name__}(tag_id={self.tag_id!r}"
+            f", tag_info={self.tag_info!r}, data=[{len(self.data)}:{data}])>"
+        )
+
+    @classmethod
+    def deserialize(cls, data: bytes) -> tuple[TelinkSubElement, bytes]:
+        if len(data) < 8:
+            raise ValueError("Data too short to contain encrypted Telink subelement")
+
+        tag_id, data = ElementTagId.deserialize(data)
+        tag_length, data = t.uint32_t.deserialize(data)
+
+        if len(data) < tag_length:
+            raise ValueError(
+                f"Data too short to contain Telink subelement data: expected"
+                f" {tag_length} bytes, got {len(data)}"
+            )
+
+        tag_info, data = t.uint16_t.deserialize(data)
+        tag_data, data = data[:tag_length], data[tag_length:]
+
+        return cls(tag_id=tag_id, tag_info=tag_info, data=tag_data), data
+
+    def serialize(self) -> bytes:
+        res = self.tag_id.serialize()
+        res += t.uint32_t(len(self.data)).serialize()
+        res += t.uint16_t(self.tag_info).serialize()
+        res += self.data
+
+        return res
+
+
+@attr.s
+class TelinkOTAImage(BaseOTAImage):
+    """Telink OTA image. Includes a proprietary "tag info" after the tag length."""
+
+    header: OTAImageHeader = attr.ib(default=None)
+    subelements: t.List[SubElement | TelinkEncryptedSubElement] = attr.ib(default=None)
+
+    @classmethod
+    def deserialize(cls, data: bytes) -> tuple[Self, bytes]:
+        hdr, data = OTAImageHeader.deserialize(data)
+        elements_len = hdr.image_size - hdr.header_length
+
+        if elements_len > len(data):
+            raise ValueError(
+                f"Data is too short for {cls}: expected at least {hdr.image_size} -"
+                f" {hdr.header_length} = {elements_len} bytes, got {len(data)}"
+            )
+
+        image = cls(header=hdr, subelements=[])
+        element_data, data = data[:elements_len], data[elements_len:]
+
+        while element_data:
+            tag_id, _ = ElementTagId.deserialize(element_data)
+
+            if tag_id == 0xF000:
+                element, element_data = TelinkEncryptedSubElement.deserialize(
+                    element_data
+                )
+            else:
+                element, element_data = SubElement.deserialize(element_data)
+
+            image.subelements.append(element)
+
+        return image, data
+
+    def serialize(self) -> bytes:
+        res = super().serialize()
+
+        if self.header.image_size != len(res):
+            raise ValueError(
+                f"Image size in header ({self.header.image_size} bytes)"
+                f" does not match actual image size ({len(res)} bytes)"
+            )
+
+        return res
+
+
 def parse_ota_image(data: bytes) -> tuple[BaseOTAImage, bytes]:
     """Attempts to extract any known OTA image type from data. Does not validate firmware."""
 
@@ -304,4 +398,13 @@ def parse_ota_image(data: bytes) -> tuple[BaseOTAImage, bytes]:
         # subelements after that. Try it first.
         return HueSBLOTAImage.deserialize(data)
     except ValueError:
+        pass
+
+    try:
+        # Otherwise, try parsing as a spec-compliant OTA image
         return OTAImage.deserialize(data)
+    except ValueError:
+        pass
+
+    # Finally, try the Telink proprietary format
+    return TelinkOTAImage.deserialize(data)
```

puddly:
That being said, it would be best if Telink fixed this (CC @binYang1 from https://github.com/telink-semi/telink_zigbee_sdk). The Zigbee specification explicitly states that OTA files are a simple header with appended TLVs. Custom data, cryptographic signatures, certificates, firmwares for different chips, anything you want can be encoded as long as it follows the basic TLV structure . There's no way to extend this to `tag_id, data_len, tag_info, data` without breaking compatibility with every other client, this `tag_info` field is Telink-internal and needs to be part of `data` to keep the TLV structure intact.

puddly:
We'd like to get this into the Home Assistant beta ASAP so I've implemented the above approach in https://github.com/zigpy/zigpy/pull/1736.

PR #1740, opened 2026-01-09
Better enum typing

puddly:
Since we support more recent Python versions now, we can ditch some of the typing hacks that make type checkers very unhappy. This PR fixes a bug affecting type checking in an unrelated repo.

Our internal `enum_factory` function can be removed. This is unfortunate used in a single place in quirks but this is a mistake, since the subclass should be `t.enum8`.

A side effect of typing this more correctly is that it is no longer `Any` so a few type issues with enums have shown up elsewhere, which have now been fixed.


puddly:
Missing coverage is for code copied from CPython so I think it's fine for now.

PR #1741, opened 2026-01-09
Improve type annotations in OTA modules

TheJulianJES:
## Proposed changes
- Fixes some typing related issues in various OTA modules.
- `jitter` is now also converted from a `float` to an `int`. I guess zigpy must have done an implicit conversion later?
- Adds a missing import needed for `zigpy.profiles.zha.PROFILE_ID` (only imported implicitly before by something else).

### DIFF
--- a/zigpy/ota/__init__.py
+++ b/zigpy/ota/__init__.py
@@ -372,10 +372,8 @@ class OTA:
             return await provider.load_index()
 
     @zigpy.util.combine_concurrent_calls
-    async def _fetch_image(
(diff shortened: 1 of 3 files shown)


TheJulianJES:
There's also a completely invalid import here:
https://github.com/zigpy/zigpy/blob/d44c73733744d95ae4d38d825667b8961cc295b9/zigpy/ota/manager.py#L18

Fixing it reveals a lot of other mypy errors, as `image.firmware` can be `None`, which the individual methods do not check for.

PR #1743, opened 2026-01-09
Fix mypy `operator` ignore

puddly:
This mostly affects types alone so it should be safe to merge.


zigpy/backports/enum.py
```diff
@@ -3,17 +3,13 @@
 from __future__ import annotations
 
 from enum import Enum
-from typing import Any, TypeVar
-
-_StrEnumSelfT = TypeVar("_StrEnumSelfT", bound="StrEnum")
+from typing import Any, Self
 
 
 class StrEnum(str, Enum):
```
TheJulianJES: We could remove this entirely now, right?

zigpy/types/basic.py
```diff
@@ -3,11 +3,9 @@
 import enum
 import inspect
 import struct
-import typing
-from typing import Self
+from typing import Literal, Self, TypeVar
 
-CALLABLE_T = typing.TypeVar("CALLABLE_T", bound=typing.Callable)
-T = typing.TypeVar("T")
+T = TypeVar("T")
```
TheJulianJES: I think this is no longer used now?

zigpy/types/basic.py
TheJulianJES: The `cls` is no longer typed as `T`, so should this comment be adjusted?

zigpy/types/basic.py
```diff
@@ -751,9 +749,9 @@ class bitmap64_be(
 
 
 class BaseFloat(float):
-    _exponent_bits = None
-    _fraction_bits = None
-    _size = None
+    _exponent_bits: int = None
+    _fraction_bits: int = None
+    _size: int = None
```
TheJulianJES, line 752: These can realistically not stay `None`, so no need to change the typing to include `None`?

TheJulianJES (commented):
Looks good

zigpy/types/basic.py
puddly: I have a follow-up PR that fixes a lot of typing errors related to these classes, it addresses this.

TheJulianJES (approved):

PR #1750, opened 2026-01-27
Populate `Device.original_signature` before loading quirks

puddly:
This won't help our existing diagnostics but it will help future diagnostics contain the correct information, even when v2 quirks are used.

ZHA will need to be modified to dump `.original_signature` and use it for loaded devices.

### DIFF
--- a/zigpy/device.py
+++ b/zigpy/device.py
@@ -135,6 +135,8 @@ class Device(zigpy.util.LocalLogMixin, zigpy.util.ListenableMixin):
             )
         )
 
+        self._signature: dict[str, Any] | None = None
+
     def create_task(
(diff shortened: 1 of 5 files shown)


TheJulianJES (approved):
I think this is fine 👍

zigpy/device.py
```diff
@@ -135,6 +135,10 @@ def __init__(self, application: ControllerApplication, ieee: t.EUI64, nwk: t.NWK
             )
         )
 
+        # Persist the original signature information for the device, before quirks are
+        # applied
+        self._original_signature: dict[str, Any] | None = None
```
TheJulianJES: nit: Could maybe be moved upwards a bit in `__init__`, but also fine like this.

PR #1754, opened 2026-01-28
Suppress ZCL read/update events more granularly

puddly:
To support use cases like this (until we find a better API to migrate to):

```python
class DoublingCluster(zcl.Cluster):
    def _update_attribute(self, attrid, value):
        if attrid == self.AttributeDefs.test_attr.id:
            # Double the value
            value = value * 2
            super()._update_attribute(attrid, value)

        # Also update a different attribute
        if attrid == self.AttributeDefs.test_attr.id:
            super()._update_attribute(self.AttributeDefs.other_attr.id, 123)
```

We need to filter with more granularity than just globally suppressing events in a context manager. Instead, we suppress events only for a given cluster ID and attribute ID and emit an attribute update event independently if the value changes.

### DIFF
--- a/tests/test_zcl.py
+++ b/tests/test_zcl.py
@@ -1823,16 +1823,19 @@ async def test_report_attributes_quirk_transforms_value(app_mock):
         {
             DoublingCluster.AttributeDefs.test_attr: t.uint8_t(25),
             DoublingCluster.AttributeDefs.passthrough_attr: t.uint8_t(77),
+            DoublingCluster.AttributeDefs.swallowed_attr: t.uint8_t(99),
         },
     ):
         await cluster.read_attributes(
             [
                 DoublingCluster.AttributeDefs.test_attr,
                 DoublingCluster.AttributeDefs.passthrough_attr,
+                DoublingCluster.AttributeDefs.swallowed_attr,
             ]
         )
 
     assert events == [
+        # No event for swallowed_attr since quirk swallows it entirely
         # No AttributeReadEvent for test_attr since the value was transformed
         # AttributeUpdatedEvent for other_attr (quirk side-effect)
         AttributeUpdatedEvent(
--- a/tests/test_zcl.py
+++ b/tests/test_zcl.py
@@ -1683,6 +1683,9 @@ async def test_report_attributes_quirk_transforms_value(app_mock):
             other_attr = foundation.ZCLAttributeDef(
                 id=0x0002, type=t.uint8_t, access="r"
             )
+            passthrough_attr = foundation.ZCLAttributeDef(
+                id=0x0003, type=t.uint8_t, access="r"
+            )
 
         def _update_attribute(self, attrid, value):
             if attrid == self.AttributeDefs.test_attr.id:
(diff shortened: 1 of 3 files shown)


zigpy/zcl/__init__.py
```diff
@@ -937,6 +949,8 @@ async def read_attributes(
                                 value=value,
                             ),
                         )
+
+                        self._update_attribute_with_event(attr_def, value)
```
TheJulianJES: From my understanding:

Before, we emitted `AttributeReadEvent` after the attribute cache was possibly modified by quirks.
So, in ZHA this triggers new sensor state (`_handle_attribute_updated_event`), which directly looks at the (modified) attribute cache.

Now, we first emit the `AttributeReadEvent`. This would cause a new ZHA sensor state already (with the non-quirk modified value), right? (though I guess that update won't really execute until we're finished here)
Only afterwards, we update the attribute cache with `_update_attribute_with_event`. Is this a non-issue? Or in general, the "double events" for `_handle_attribute_updated_event`?

puddly: Ah, I think you're right. Let me take a closer look.

puddly: Ok, I think I covered the quirk use case.

If an attribute is changed during a report/read, we replace the `AttributeReportedEvent`/`AttributeReadEvent` with `AttributeUpdatedEvent`, preventing the duplicate update in ZHA.

puddly: I think the current branch now works properly, I'm basically explicitly testing the Aqara case. Does it work for you?

zigpy/zcl/__init__.py
```diff
@@ -752,34 +800,71 @@ def handle_cluster_general_request(
...
+                            cluster_type=self._type,
+                            cluster_id=self.cluster_id,
+                            attribute_name=None,
+                            attribute_id=attr.attrid,
+                            manufacturer_code=hdr.manufacturer,
+                            raw_value=attr.value.value,
+                            value=value,
+                        ),
+                    )
+                    continue
```
TheJulianJES, line 827: note: If there's no attribute definition for an attribute report, the emitted event will always have the raw/original value, not the possibly modified one from the quirk.

And the old `attribute_updated` `listener_event` is no longer emitted for this case.

I think this should be fine for basically all quirks? So, just noting.

tests/test_zcl.py
```diff
@@ -1653,3 +1665,222 @@ class ServerCommandDefs(zcl.foundation.BaseCommandDefs):
         await cluster.command(0x00, manufacturer=0x9999)
 
     assert mock_request.mock_calls[0].kwargs["manufacturer"] == 0x9999
+
+
+async def test_report_attributes_quirk_transforms_value(app_mock):
+    """Test that quirks transforming values emit both reported and updated events."""
+    from zigpy.zcl.clusters.measurement import OccupancySensing
```
TheJulianJES, line 1672: nit: Is this non top-level import intentional/needed?

TheJulianJES (commented):
Had a thorough look at this now and it looks really good. Fixes my P1 motion sensor issue and also everything else I can think of with quirks. Nice work! 😄  

One minor behavioral change (compared to before the manufacturer rewrite) is that the old `attribute_updated` `listener_event` is no longer fired if there's no corresponding attribute definition for an attribute report. Previously, we did that anyway: [old/zigpy/zcl/__init__.py#L917](https://github.com/zigpy/zigpy/blob/f7550584b04d0df2aa8af3514ec0931fb6fee6d5/zigpy/zcl/__init__.py#L917) + [old/zigpy/zcl/__init__.py#L561-L573](https://github.com/zigpy/zigpy/blob/f7550584b04d0df2aa8af3514ec0931fb6fee6d5/zigpy/zcl/__init__.py#L561-L573)
The emitted `AttributeReportedEvent` also always contains the raw/original value in that case, not the possibly quirk-modified one.

I think we don't rely on that behavior anywhere though, so it should be fine to keep as is.

Another small thing (related to the entire refactor) is probably that setting values by (attempting to) directly modifying the `_attr_cache` now fires events, but I think that's fine? (could use `_suppress_attribute_update_event` in the setter there if not)

PR #1756, opened 2026-01-29
Correctly handle `manufacturer_code=None` for attributes and commands

puddly:
See https://github.com/zigpy/zha-device-handlers/issues/4694 for more context.

`manufacturer_code` in the ZCL attribute definitions should have `None` mean "do not provide a manufacturer code". A specific value should be used otherwise. `UNDEFINED`, the default, allows for auto-detection from the cluster ID and other context (e.g. just `is_manufacturer_specific=True`).

### DIFF
--- a/zigpy/zcl/foundation.py
+++ b/zigpy/zcl/foundation.py
@@ -9,7 +9,7 @@ import typing
 from typing import Final, Self
 
 import zigpy.types as t
-from zigpy.typing import UNDEFINED
+from zigpy.typing import UNDEFINED, UndefinedType
 
 _LOGGER = logging.getLogger(__name__)
 
@@ -1113,7 +1113,7 @@ class ZCLCommandDef(t.BaseDataclassMixin):
 
     # set later
     name: str = None
-    manufacturer_code: t.uint16_t | None = None
+    manufacturer_code: t.uint16_t | UndefinedType | None = UNDEFINED
 
     def __post_init__(self) -> None:
         # Backwards compatibility with positional syntax where the name was first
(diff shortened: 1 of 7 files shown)


zigpy/zcl/foundation.py
```diff
@@ -1281,8 +1281,14 @@ def __post_init__(self) -> None:
...
+        if (
+            self.manufacturer_code not in (None, UNDEFINED)
+            and not self.is_manufacturer_specific
+        ):
+            object.__setattr__(self, "is_manufacturer_specific", True)
+
         # Use UNDEFINED for manufacturer-specific attributes without explicit code
-        if self.is_manufacturer_specific and self.manufacturer_code is None:
+        if self.is_manufacturer_specific is True and self.manufacturer_code is None:
             object.__setattr__(self, "manufacturer_code", UNDEFINED)
```
TheJulianJES: nit: I think at least the comment here is outdated now (and condition?), since "manufacturer-specific attributes without explicit code"  have `UNDEFINED` as a default `manufacturer_code` now, not `None`.

puddly:
@TheJulianJES  We had a small hole in the database migration that prevented some manufacturer-specific attributes from being migrated. This has been fixed in this PR as well. You can probably re-trigger this by downgrading your database version with `PRAGMA user_version=13`, zigpy should delete and re-migrate the tables.

TheJulianJES:
Do you think it matters for those who already upgraded to the HA beta?

puddly:
It's all cached data so it probably has been updated by now, probably not.

TheJulianJES (approved):
Changes make sense to me

PR #1758, opened 2026-01-30
Fix interaction between `is_manufacturer_specific` and `manufacturer_id_override`

puddly:
This PR changes the `is_manufacturer_specific` default from `False` to `None`, allowing us to use overrides of the default as a hint when computing the manufacturer code for commands and attribute reads/writes.

- If disabled with `False`, we do not send a manufacturer code, even if the attribute is part of a cluster in a manufacturer-specific range.
- If enabled with `True`, we always send one, preferring the cluster-level `manufacturer_id_override` if it exists, falling back to the node descriptor otherwise.
- Both are overridden by `manufacturer_code`, the preferred way to set this value. It respects both `None` and specific values.


TheJulianJES (approved):
This looks perfect! (except for the failing test 😄)

PR #1760, opened 2026-02-05
Make `update_attribute` manufacturer code aware

TheJulianJES:
## Proposed change

This removes the legacy fallback in `_legacy_apply_quirk_attribute_update` to insert values into the legacy cache for attributes that have a shared ID. A `ContextVar` with the attribute ID and manufacturer code is used to preserve the original manufacturer code sent by the device and later update the typed cache with that.

## Additional information

An additional PR will be made later to fix an issue where `UnsupportedError` is raised too early, before the legacy cache is accessed.

Later, we can also merge something like https://github.com/TheJulianJES/zigpy/compare/tjj/manufacturer_update_fix...TheJulianJES:zigpy:tjj/manufacturer_update_fix_with_mfr_attr_fix_zcl_cache. See the comments for more details on this.

### DIFF
--- a/tests/test_zcl.py
+++ b/tests/test_zcl.py
@@ -2174,35 +2174,9 @@ async def test_quirk_manufacturer_code_context_isolation(app_mock) -> None:
     cluster.on_event(AttributeUpdatedEvent.event_type, events.append)
 
     # Report the manufacturer-specific attribute
-    frame_control = foundation.FrameControl(
-        frame_type=foundation.FrameType.GLOBAL_COMMAND,
-        is_manufacturer_specific=True,
-        direction=foundation.Direction.Server_to_Client,
-        disable_default_response=True,
-        reserved=0b000,
+    await mock_attribute_report(
+        cluster, {TestCluster.AttributeDefs.manuf_attr: t.uint8_t(42)}
     )
-
-    hdr = foundation.ZCLHeader(
-        frame_control=frame_control,
-        manufacturer=0x1234,
-        tsn=1,
-        command_id=foundation.GeneralCommand.Report_Attributes,
-    )
-
-    command = foundation.GENERAL_COMMANDS[
-        foundation.GeneralCommand.Report_Attributes
-    ].schema(
-        attribute_reports=[
-            foundation.Attribute(
-                attrid=0x0001,
-                value=foundation.TypeValue(
-                    type=foundation.DataType.uint8.type_id, value=42
-                ),
-            )
-        ]
-    )
-
-    cluster.handle_cluster_general_request(hdr, command)
 
     # The legacy cache should not contain the attribute, as the typed cache was used
     assert 0x0001 not in cluster._attr_cache._legacy_cache
--- a/tests/test_zcl.py
+++ b/tests/test_zcl.py
@@ -2173,6 +2173,9 @@ async def test_quirk_manufacturer_code_context_isolation(app_mock) -> None:
     cluster.on_event(AttributeReportedEvent.event_type, events.append)
     cluster.on_event(AttributeUpdatedEvent.event_type, events.append)
 
+    # The attribute is currently marked as unsupported
+    cluster.add_unsupported_attribute(TestCluster.AttributeDefs.manuf_attr)
+
     # Report the manufacturer-specific attribute
     await mock_attribute_report(
         cluster, {TestCluster.AttributeDefs.manuf_attr: t.uint8_t(42)}
--- a/tests/test_zcl.py
+++ b/tests/test_zcl.py
@@ -2123,3 +2123,124 @@ def test_manufacturer_id_override_extended_zcl_cluster(app_mock) -> None:
         (TestCluster.ServerCommandDefs.reset_fact_default, None),
     ]:
         assert cluster._get_effective_manufacturer_code(definition) is expected
+
+
+async def test_quirk_manufacturer_code_context_isolation(app_mock) -> None:
+    """Test that manufacturer code context is properly handled in _update_attribute.
+
+    When a manufacturer-specific attribute is reported and the cluster has multiple
+    attributes sharing the same ID (with different manufacturer codes), the
+    _update_attribute call must use the correct manufacturer code. This tests that:
+    1. The value is stored directly in the typed cache (not via legacy cache fallback)
+    2. Other attributes updated by quirks don't inherit the manufacturer code context
+    """
+
+    class TestCluster(zcl.Cluster):
+        cluster_id = 0xABCD
+        ep_attribute = "test_cluster"
+        _skip_registry = True
+
+        class AttributeDefs(zcl.foundation.BaseAttributeDefs):
+            # Two attributes sharing the same ID with different manufacturer codes
+            manuf_attr = foundation.ZCLAttributeDef(
+                id=0x0001,
+                type=t.uint8_t,
+                manufacturer_code=0x1234,
+            )
+            standard_attr = foundation.ZCLAttributeDef(
+                id=0x0001,
+                type=t.uint8_t,
+                manufacturer_code=None,
+            )
+            # A different attribute that the quirk will also update
+            other_attr = foundation.ZCLAttributeDef(
+                id=0x0002,
+                type=t.uint8_t,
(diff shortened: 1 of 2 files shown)


tests/test_zcl.py
```diff
@@ -2123,3 +2123,124 @@ class ServerCommandDefs(Basic.ServerCommandDefs):
...
+
+    dev = add_initialized_device(app_mock, nwk=0x1234, ieee=make_ieee(1))
+    cluster = TestCluster(dev.endpoints[1])
+    dev.endpoints[1].add_input_cluster(TestCluster.cluster_id, cluster)
+
+    events = []
+    cluster.on_event(AttributeReportedEvent.event_type, events.append)
+    cluster.on_event(AttributeUpdatedEvent.event_type, events.append)
+
+    # Report the manufacturer-specific attribute
```
puddly, line 2179: nit: we now have `mock_attribute_report`, which handles setting up the report with manufacturer codes and such: https://github.com/zigpy/zigpy/blob/dev/tests/test_zcl.py#L1758

TheJulianJES: I forgot to push two commits from a few days ago. Fixed now: https://github.com/zigpy/zigpy/pull/1760/changes/e0bf8c686938c668f693f21499bc6ed430ec78a7

zigpy/zcl/__init__.py
```diff
@@ -1113,8 +1136,17 @@ def _update_attribute(
...
+
         try:
-            attr_def = self.find_attribute(attrid)
+            if ctx is not None and ctx[0] == attrid:
+                attr_def = self.find_attribute(attrid, manufacturer_code=ctx[1])
+            else:
+                attr_def = self.find_attribute(attrid)
         except KeyError:
             if value is not None:
                 self._attr_cache.set_legacy_value(attrid, value)
```
TheJulianJES, line 1152: Both with and without this PR, we persist manufacturer manufacturer specific attribute reports into the mfr-keyed attribute cache if there's only a ZCL definition. IMO, we should adjust this behavior to only insert into the new typed cache if there's a proper attribute definition with the same manufacturer code.

If there isn't, we should only insert into legacy cache IMO. But currently, the legacy cache really isn't used, as `cluster.get(some_legacy_attr)` won't return the legacy cache value.

TheJulianJES: But right now, we're somewhat polluting the new attribute cache with `manufacturer_code=None` for manufacturer-specific attributes if there's no definition for them yet.

We can still go ahead with this PR now, as it doesn't change this behavior. But we should address that issue very soon.

zigpy/zcl/__init__.py
```diff
@@ -1113,8 +1136,17 @@ def _update_attribute(
...
 
+        # Get the manufacturer code from context if set (set by
+        # _legacy_apply_quirk_attribute_update to preserve manufacturer code).
+        # Only apply when the attribute ID matches the original to avoid affecting
+        # other attributes that quirks may update.
+        ctx = _attribute_update_context.get()
+
         try:
-            attr_def = self.find_attribute(attrid)
+            if ctx is not None and ctx[0] == attrid:
```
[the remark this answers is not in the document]
TheJulianJES: It'll never fail because everywhere where `_set_attribute_update_context` is called, `update_attribute` is called with the attribute ID (`int`). It's not optimal though, but we have the same "issue" for `_suppress_attribute_update_event`.

zigpy/zcl/__init__.py
```diff
@@ -1113,8 +1136,17 @@ def _update_attribute(
         # other clusters or attributes to emit their own events.
         suppressed = (self.cluster_id, attrid) in _suppressed_attribute_updates.get()
```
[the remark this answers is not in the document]
TheJulianJES: Ha, this is what I mentioned in the comment above. Everywhere `_suppress_attribute_update_event` is used, we're calling `update_attribute` with the attribute ID (`int`), so it's not really an issue.

zigpy/zcl/__init__.py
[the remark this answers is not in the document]
TheJulianJES: This isn't a new issue with this PR. It's still something we should address, but it can only happen if a quirk calls `update_attribute` with a `ZCLAttributeDef` from another cluster. Otherwise, `find_attribute` will always return the correct definition and we'll never hit this. IMO, it's not urgent to address.

TheJulianJES: A nice fix for this, built on top of this PR, would be the following: https://github.com/TheJulianJES/zigpy/compare/tjj/manufacturer_update_fix...TheJulianJES:zigpy:tjj/manufacturer_update_fix_with_mfr_attr_fix_zcl_cache

It's (probably?) not something we should do for a patch release, but this should work nicely.

puddly: > But currently, the legacy cache really isn't used, as `cluster.get(some_legacy_attr)` won't return the legacy cache value.

Do you mean that we don't allow persisting attributes without a corresponding definition? I believe the few quirks that rely on this behavior directly mess with `_attr_cache`.

TheJulianJES: > Do you mean that we don't allow persisting attributes without a corresponding definition?

They should be persisted into the database.

(But for attribute reports that do not have a corresponding definition, the unmodified value will land in the DB. With the modified one being stored in the runtime legacy cache. Though that seems to never be used really?)

See the last block for what I meant with my original comment.

---

So, I searched for `_attr_cache` in the quirks repo and except for [one usage in `LocalDataCluster`](https://github.com/zigpy/zha-device-handlers/blob/a49b8344b8c22821e87fde45aa868fbed2b8534d/zhaquirks/__init__.py#L104), all quirks I looked at had proper attribute definitions when they accessed `_attr_cache.get(attr_id)` directly.

I'm only aware of the Aqara P1 motion sensor quirk relying on the behavior of `_update_attribute` being called with an attribute ID with no corresponding definition [here](https://github.com/zigpy/zha-device-handlers/blob/a49b8344b8c22821e87fde45aa868fbed2b8534d/zhaquirks/xiaomi/__init__.py#L516). There may be a few other quirks that do this, but I imagine it's only very few. For that one, it's also not needed to persist the attribute into the DB.

Right now, it should also be saved to the DB with the manufacturer code of the attribute (which is correct): [zigpy/zcl/__init__.py#L861-L880](https://github.com/zigpy/zigpy/blob/cf3710843c60aa55c0f2745f8260b3fb672dde18/zigpy/zcl/__init__.py#L861-L880)
For attributes without definition, only the unmodified value is persisted into the DB. For ones with a definition, the modified value is persisted (depends on `value` in `AttributeReportedEvent`). This behavior is a bit weird..
`_update_attribute` will also (correctly) call `_attr_cache.set_legacy_value` with the modified value.

---

What I meant with my original comment was that ZHA sensors cannot access values only stored in the legacy cache at the moment. E.g. [zha/application/platforms/sensor/__init__.py#L321](https://github.com/zigpy/zha/blob/99ea585c3456b16e22fa395648f0dea79169e996/zha/application/platforms/sensor/__init__.py#L321) will call `Cluster.get`: [zigpy/zcl/__init__.py#L1594-L1600](https://github.com/zigpy/zigpy/blob/cf3710843c60aa55c0f2745f8260b3fb672dde18/zigpy/zcl/__init__.py#L1594-L1600), which calls `AttributeCache.get_value` directly [zigpy/zcl/helpers.py#L71-L73](https://github.com/zigpy/zigpy/blob/cf3710843c60aa55c0f2745f8260b3fb672dde18/zigpy/zcl/helpers.py#L71-L73), bypassing `AttributeCache.get` (which would have called `AttributeCache.__getitem__`). The latter two would have checked the legacy cache. ZHA entities do not, as they generally use `cluster.get`.

This does not seem to be an issue, since we shouldn't generally access the attribute cache without a definition. But it would be good to know what actually lands in the legacy cache at all now. Anything, except for the P1 sensor?

puddly: https://github.com/zigpy/zigpy/pull/1774 should address the `.get` regression.

puddly (approved):

PR #1762, opened 2026-02-05
Check for _db.join method before execution

HiDiHo01:
Check for _db.join method before execution

### DIFF
--- a/zigpy/appdb.py
+++ b/zigpy/appdb.py
@@ -225,11 +225,9 @@ class PersistingListener(zigpy.util.CatchingTaskMixin):
         await self._db.close()
 
         # FIXME: aiosqlite's thread won't always be closed immediately
+        # No join available, skip for compatibility with Windows/SQLite
         if hasattr(self._db, "join"):
             await asyncio.get_running_loop().run_in_executor(None, self._db.join)
-        else:
-            # No join available, skip for compatibility with Windows/SQLite
-            pass
 
     def register_cluster_events(self, cluster) -> None:
         cluster.on_event(AttributeReadEvent.event_type, self.on_attribute_read)
--- a/zigpy/appdb.py
+++ b/zigpy/appdb.py
@@ -225,7 +225,11 @@ class PersistingListener(zigpy.util.CatchingTaskMixin):
         await self._db.close()
 
         # FIXME: aiosqlite's thread won't always be closed immediately
-        await asyncio.get_running_loop().run_in_executor(None, self._db.join)
+        if hasattr(self._db, "join"):
+            await asyncio.get_running_loop().run_in_executor(None, self._db.join)
+        else:
+            # No join available, skip for compatibility with Windows/SQLite
+            pass
 
     def register_cluster_events(self, cluster) -> None:
         cluster.on_event(AttributeReadEvent.event_type, self.on_attribute_read)


puddly:
Can you provide some more context about what version of Python and SQLite you are using?

HiDiHo01:
I'm using latest Home Assistant 2026.2.0 on native Windows with python 3.14
It does not have "join" method

puddly:
How are you running it on Windows? WSL2?

HiDiHo01:
> How are you running it on Windows? WSL2?

No, just native Windows using native Python using core HA
(have modified HA for this to still work in 2026.2.0)
I don't want virtualization or extra layers. and want to use native deconz app for Windows along with ZHA
ps everything is working except backup that broke in 2025.12

puddly:
What version of aiosqlite are you using? We pin it to a specific range: https://github.com/zigpy/zigpy/blob/b2ee3c3b1a6b1f0c1414ff3f4ec68676d93fd536/pyproject.toml#L19

HiDiHo01:
C:\Program Files\HassWP2>pip show aiosqlite
Name: aiosqlite
Version: 0.22.1
Summary: asyncio bridge to the standard sqlite3 module
Home-page:
Author:
Author-email: Amethyst Reese <amethyst@n7.gg>
License:
Location: C:\Program Files\HassWP2\python\Lib\site-packages
Requires:
Required-by: zigpy

puddly:
Your `aiosqlite` version is incompatible with zigpy's constraints. Maybe the `pip` command that Home Assistant executes via `subprocess` does not properly abide by package constraints on Windows? Or it's overridden locally for you?

HiDiHo01:
I think it's overridden, it's not a big problem. Everything is working. It was only a warning that it could not use join.
I don't see it at this moment (oh that because of my fix/edit)
I believe it is on shutdown only

HiDiHo01:
This was the warning

Logger: zigpy.application
Bron: C:\Program Files\HassWP2\python\Lib\site-packages\zigpy\application.py:587
Eerst voorgekomen: 14:20:48 (1 gebeurtenis)
Laatst gelogd: 14:20:48

Failed to disconnect from database
  File "C:\Program Files\HassWP2\python\Lib\site-packages\zigpy\application.py", line 585, in shutdown
    await self._dblistener.shutdown()
  File "C:\Program Files\HassWP2\python\Lib\site-packages\zigpy\appdb.py", line 229, in shutdown
    await asyncio.get_running_loop().run_in_executor(None, self._db.join)
                                                           ^^^^^^^^^^^^^
AttributeError: 'Connection' object has no attribute 'join'

puddly:
You need to downgrade aiosqlite, we don't support this version.

HiDiHo01:
Downgraded to 0.21.0 and lost all connections to aiohue
pip install aiosqlite==0.21.0 --upgrade

back to 0.22.1 and it's working again

HiDiHo01 (commented):
Whats the problem not adding this guard?

TheJulianJES:
We will require the latest `aiosqlite` version in the future, but for now we explicitly don't allow this version. See [this](https://github.com/zigpy/zigpy/blob/7ba1551a950fb972f7844e0ef26c9ff1012c7239/pyproject.toml#L19).

It doesn't make sense to add a guard for something that can never be hit, unless the install is wildly broken. That's the case for you. You'll eventually run into other issues, regardless of us adding this one guard for your incorrect install.

PR #1766, opened 2026-02-09
v15 database schema migration to deprioritize unsupported attributes

puddly:
The v14 schema migration prioritized unsupported attribute table entries over those in the attribute cache. This breaks devices that had entries in *both* tables. We can perform a new migration to `v15` and peek at the `v13` tables (if they exist) to reprioritize them after the fact.


zigpy/appdb.py
```diff
@@ -1416,13 +1422,61 @@ async def _migrate_to_v14(self) -> None:
...
+                    c13.value, c13.last_updated
+                FROM attributes_cache_v15 a15
+                JOIN attributes_cache_v13 c13
+                    ON c13.ieee = a15.ieee
+                    AND c13.endpoint_id = a15.endpoint_id
+                    AND c13.cluster_type = a15.cluster_type
+                    AND c13.cluster_id = a15.cluster_id
+                    AND c13.attr_id = a15.attr_id
+                WHERE a15.status = :unsupported
+                    AND a15.manufacturer_code = :unmigrated
```
TheJulianJES: Is this not too strict? Wouldn't the runtime migration on previous versions already have migrated to the correct manufacturer code with the manufacturer code value just being `NULL` (or the actual code) now, instead of unmigrated?

tests/test_appdb_migration.py
```diff
@@ -718,3 +724,279 @@ class AttributeDefs(BaseAttributeDefs):
...
+        )
+        rows = cur.fetchall()
+
+    assert len(rows) == 3
+    assert rows[0] == (4, None, Status.SUCCESS)
+    assert rows[1] == (5, None, Status.SUCCESS)
+    assert rows[2] == (6, None, Status.SUCCESS)
+
+
+async def test_data_migration_ambiguous_attributes(tmp_path):
```
TheJulianJES, line 843: The `test_data_migration_ambiguous_attributes` function is duplicated now

zigpy/appdb.py
```diff
@@ -1539,3 +1545,72 @@ async def _migrate_to_v14(self) -> None:
...
+            WHERE status = :unsupported
+            AND EXISTS (
+                SELECT 1
+                FROM attributes_cache_v13 c13
+                WHERE c13.ieee = attributes_cache_v15.ieee
+                    AND c13.endpoint_id = attributes_cache_v15.endpoint_id
+                    AND c13.cluster_type = attributes_cache_v15.cluster_type
+                    AND c13.cluster_id = attributes_cache_v15.cluster_id
+                    AND c13.attr_id = attributes_cache_v15.attr_id
+            )
```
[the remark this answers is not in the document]
TheJulianJES: The suggestions mostly miss the point of this PR. However, I think the `last_updated = 0` check would actually work to only filter out potentially inaccurate unsupported attributes from the migration, but not ones correctly added "live on v14". Whether this is worth doing, I'm not sure about.

zigpy/appdb.py
```diff
@@ -1568,3 +1574,72 @@ async def _migrate_to_v14(self) -> None:
...
+                    AND a15.cluster_type = c13.cluster_type
+                    AND a15.cluster_id = c13.cluster_id
+                    AND a15.attr_id = c13.attr_id
+            )
+            """,
+            {
+                "status": Status.SUCCESS,
+                "manufacturer_code": UNMIGRATED_MANUFACTURER_CODE,
+            },
+        )
```
TheJulianJES, line 1645: I'm not sure if this is an actual issue, but if a user completely deleted their device on v14, the second statement would always re-insert old rows from v13, even if we did not actually delete any "unsupported" rows with the first statement, no?

Somewhat similar to the other PR, do we need to check if we deleted any rows in v15 and only try to re-insert those ones from v13?

TheJulianJES: (Especially if we narrow down the first statement to only delete rows that were migrated from v13 to v14, so if we add `last_updated = 0` to that, we'd only really re-insert rows from v13 that were actually (incorrectly) migrated from v13 as unsupported (related: [comment above](https://github.com/zigpy/zigpy/pull/1766#discussion_r2796085758)).)

puddly: That's a really good point and I think would break the migration. The second query would fail due to a foreign key constraint failure, since the underlying `device` row would be missing.

TheJulianJES: But what if a user deleted it on v14, then re-added it on v14 still? (e.g. to get rid of some old attribute cache)
Then he updates later. Wouldn't we restore the old values from v13 then if there's no row for them at all in v14/v15 (e.g. changed quirk or something).

Oh, or you meant "breaking" as in "not good breaking" already 😅

puddly: Yeah. There isn't a way to tell if a device has been removed and then re-added (I guess without peeking at the `rowid` or other SQLite internals). So I think these queries should be narrowed significantly.

puddly:
I think this may not be needed any more. Closing.

PR #1767, opened 2026-02-09
Correctly parse the ZCL default response in `write_attributes`

puddly:
https://github.com/zigpy/zigpy/pull/1764 used `status = rsp[1]` but `write_attributes` uses `status = response[0]`. I think both need to be aligned, since the ZCL default response stores the status as the second field.

### DIFF
--- a/tests/test_quirks.py
+++ b/tests/test_quirks.py
@@ -880,7 +880,7 @@ async def test_manuf_id_included(real_device):
     # The default behavior for a manufacturer-specific cluster, command, or attribute is
     # to include the manufacturer ID in the request
     with patch.object(ep, "request", AsyncMock()) as request_mock:
-        request_mock.return_value = (zcl.foundation.Status.SUCCESS, "done")
+        request_mock.return_value = (0x12, zcl.foundation.Status.SUCCESS)
         await ep.just_a_cluster.command(
             ep.just_a_cluster.commands_by_name["server_cmd0"].id,
         )
--- a/tests/test_zcl.py
+++ b/tests/test_zcl.py
@@ -528,15 +528,64 @@ async def test_write_attribute_types(
     "status", [foundation.Status.SUCCESS, foundation.Status.UNSUPPORTED_ATTRIBUTE]
 )
 async def test_write_attributes_cache_default_response(cluster, status):
-    write_mock = AsyncMock(
-        return_value=[foundation.GeneralCommand.Write_Attributes, status]
+    _, default_rsp = cluster._create_request(
+        general=True,
+        command_id=foundation.GeneralCommand.Default_Response,
+        schema=foundation.GENERAL_COMMANDS[
+            foundation.GeneralCommand.Default_Response
+        ].schema,
+        disable_default_response=True,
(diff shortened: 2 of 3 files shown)


TheJulianJES:
Oh, this is interesting! When do we even get a default response here? I only vaguely remember something about this.. Is it device specific?

puddly:
I just saw the inconsistency, I don't think it's a codepath that I've ever seen (it may even be impossible?) but we have it in zigpy so maybe it was seen at some point.

tests/test_quirks.py
```diff
@@ -880,7 +880,7 @@ class TestCluster(ManufacturerSpecificCluster):
     # The default behavior for a manufacturer-specific cluster, command, or attribute is
     # to include the manufacturer ID in the request
     with patch.object(ep, "request", AsyncMock()) as request_mock:
-        request_mock.return_value = (zcl.foundation.Status.SUCCESS, "done")
+        request_mock.return_value = (0x12, zcl.foundation.Status.SUCCESS)
```
TheJulianJES, line 883: I think we'll need to update quite a few quirks tests as well then 😅

tests/test_zcl.py
```diff
@@ -528,16 +528,65 @@ async def test_write_attribute_types(
...
+        },
+    )
+    cluster.endpoint.request.return_value = default_rsp
+
+    results = await cluster.write_attributes(
+        {Basic.AttributeDefs.manufacturer: "manufacturer"}
+    )
+
+    # It should correctly return FAILURE (0x01), not Write_Attributes (0x02)
+    assert results[0][0].status == foundation.Status.FAILURE
```
TheJulianJES, line 588: nit: While we're at it, should we also assert this didn't make it into the attribute cache?
On the other hand, the logic for the default response currently only adds the status to `records_group`, so this is basically already covered by the test above.

TheJulianJES (approved):
Nice catch!

tests/test_quirks.py
puddly: Looks like only three...

```python
FAILED tests/test_develco.py::test_frient_emi - ValueError: Failed to convert status='done' from type <class 'str'> to <enum 'Status'>
FAILED tests/test_tuya.py::test_fan_switch_writes_attributes[TS0501FanSwitch] - ValueError: Failed to convert status='done' from type <class 'str'> to <enum 'Status'>
FAILED tests/test_xiaomi.py::test_xiaomi_eu_plug_binding[PlugMAEU01] - ValueError: Failed to convert status='done' from type <class 'str'> to <enum 'Status'>
```

But I'm sure custom quirks might exist that rely on this behavior.

TheJulianJES: > Looks like only three...

Interesting, so I guess we don't verify the output a lot..? https://github.com/search?q=repo%3Azigpy%2Fzha-device-handlers+Status.SUCCESS%2C+%22done%22&type=code

> But I'm sure custom quirks might exist that rely on this behavior.

Really? For these, it's just the tests which do something stupid. I wouldn't expect any actual quirks to somehow return a default response with the fields in a different order.

PR #1768, opened 2026-02-09
Add writing and report config tests for manufacturer groups

TheJulianJES:
As suggested in https://github.com/zigpy/zigpy/pull/1764, this adds tests that verify the grouping of calls per manufacturer for `write_attributes` and `configure_reporting_multiple`.

### DIFF
--- a/tests/test_zcl.py
+++ b/tests/test_zcl.py
@@ -2164,14 +2164,20 @@ async def test_configure_reporting_multiple_manufacturer_groups(app_mock) -> Non
     # First call: standard attribute (no manufacturer code)
     std_call = mock_configure.call_args_list[0]
     assert std_call.kwargs["manufacturer"] is None
-    assert [c.attrid for c in std_call.args[0]] == [Basic.AttributeDefs.hw_version.id]
+    assert len(std_call.args[0]) == 1
+    assert std_call.args[0][0].attrid == Basic.AttributeDefs.hw_version.id
+    assert std_call.args[0][0].min_interval == 5
+    assert std_call.args[0][0].max_interval == 15
+    assert std_call.args[0][0].reportable_change == 20
 
     # Second call: manufacturer-specific attribute
     manuf_call = mock_configure.call_args_list[1]
     assert manuf_call.kwargs["manufacturer"] == 0x5678
-    assert [c.attrid for c in manuf_call.args[0]] == [
-        TestCluster.AttributeDefs.manuf_attr.id
-    ]
+    assert len(manuf_call.args[0]) == 1
+    assert manuf_call.args[0][0].attrid == TestCluster.AttributeDefs.manuf_attr.id
+    assert manuf_call.args[0][0].min_interval == 10
+    assert manuf_call.args[0][0].max_interval == 30
+    assert manuf_call.args[0][0].reportable_change == 5
 
 
 def test_manufacturer_id_override_manuf_specific_cluster(app_mock) -> None:
--- a/tests/test_zcl.py
+++ b/tests/test_zcl.py
@@ -840,7 +840,10 @@ async def test_handle_cluster_general_request_not_attr_report(cluster):
 
 
 async def test_configure_reporting_multiple(cluster):
-    cluster.endpoint.request.return_value = _mk_cfg_rsp()
+    cfg_response = zcl.foundation.ConfigureReportingResponse(
+        [zcl.foundation.ConfigureReportingResponseRecord(zcl.foundation.Status.SUCCESS)]
+    )
+    cluster.endpoint.request.return_value = [cfg_response]
 
     await cluster.configure_reporting(
         attribute=3,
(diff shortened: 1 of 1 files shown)


tests/test_zcl.py
```diff
@@ -2049,6 +2047,99 @@ async def test_zcl_write_attributes_update_cache(app_mock) -> None:
...
+
+    dev = add_initialized_device(app_mock, nwk=0x1234, ieee=make_ieee(1))
+    dev.node_desc.manufacturer_code = 0x1234
+
+    cluster = TestCluster(dev.endpoints[1])
+    dev.endpoints[1].add_input_cluster(TestCluster.cluster_id, cluster)
+
+    cfg_response = zcl.foundation.ConfigureReportingResponse(
+        [zcl.foundation.ConfigureReportingResponseRecord(zcl.foundation.Status.SUCCESS)]
+    )
```
TheJulianJES, line 2139: `_mk_cfg_rsp` doesn't support creating a global success response with `attrid=None`. We can either keep this as is everywhere or update it to create a global success response if `responses=None` is passed.

tests/test_zcl.py
```diff
@@ -840,9 +840,8 @@ async def test_handle_cluster_general_request_not_attr_report(cluster):
 
 
 async def test_configure_reporting_multiple(cluster):
-    cfg_response = zcl.foundation.ConfigureReportingResponse()
-    cfg_response.append(
-        zcl.foundation.ConfigureReportingResponseRecord(zcl.foundation.Status.SUCCESS)
+    cfg_response = zcl.foundation.ConfigureReportingResponse(
+        [zcl.foundation.ConfigureReportingResponseRecord(zcl.foundation.Status.SUCCESS)]
     )
```
TheJulianJES, line 845: If it's preferred to create `ConfigureReportingResponse()` and then append a `ConfigureReportingResponseRecord`, I can also switch this back everywhere.

tests/test_zcl.py
puddly: `_mk_cfg_rsp` is pretty simple, I think it's clearer to just directly create the objects like this.

tests/test_zcl.py
puddly: These ideally should be immutable objects anyways and it's an internal implementation detail that this object is a `list` subclass, I think this approach is better.

puddly (approved):

tests/test_zcl.py
```diff
@@ -2049,6 +2047,99 @@ async def test_zcl_write_attributes_update_cache(app_mock) -> None:
...
+        results = await cluster.configure_reporting_multiple(
+            {
+                Basic.AttributeDefs.hw_version: ReportingConfig(
+                    min_interval=5, max_interval=15, reportable_change=20
+                ),
+                TestCluster.AttributeDefs.manuf_attr: ReportingConfig(
+                    min_interval=10, max_interval=30, reportable_change=5
+                ),
+            }
+        )
```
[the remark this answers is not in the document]
TheJulianJES: I think it makes sense also asserting the individual attributes. Will add.

tests/test_zcl.py
```diff
@@ -2049,6 +2047,99 @@ async def test_zcl_write_attributes_update_cache(app_mock) -> None:
...
+
+    assert len(results) == 2
+    assert all(r.status == foundation.Status.SUCCESS for r in results)
+
+    # Two separate requests, one per manufacturer group
+    assert mock_write.call_count == 2
+    manufacturers = [c.kwargs["manufacturer"] for c in mock_write.call_args_list]
+    assert manufacturers == [None, 0x5678]
```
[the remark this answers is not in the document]
TheJulianJES: I think it's nice to know if the order changes, for whatever reason. If intentional, we can update the tests then.

tests/test_zcl.py
```diff
@@ -2049,6 +2047,99 @@ async def test_zcl_write_attributes_update_cache(app_mock) -> None:
...
+            }
+        )
+
+    assert len(results) == 2
+    assert all(r.status == zcl.foundation.Status.SUCCESS for r in results)
+
+    # Two separate requests should have been made (one per manufacturer group)
+    assert mock_configure.await_count == 2
+    manufacturers = [c.kwargs["manufacturer"] for c in mock_configure.call_args_list]
+    assert manufacturers == [None, 0x5678]
```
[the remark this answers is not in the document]
TheJulianJES: I think it's nice to know if the order changes, for whatever reason. If intentional, we can update the tests then.

tests/test_zcl.py
```diff
@@ -2049,6 +2057,117 @@ async def test_zcl_write_attributes_update_cache(app_mock) -> None:
...
+
+    # Two separate requests, one per manufacturer group
+    assert mock_write.call_count == 2
+
+    # First call: standard attribute (no manufacturer code)
+    std_call = mock_write.call_args_list[0]
+    assert std_call.kwargs["manufacturer"] is None
+    assert [a.attrid for a in std_call.args[0]] == [
+        Basic.AttributeDefs.location_desc.id
+    ]
```
TheJulianJES: Alternatively, we can also assert the entire call, but we need to construct some zigpy objects for that. E.g.:
```python
assert std_call == call(                                                                                                                                                                                                                                       
    [                                                                                                                                                                                                                                                          
        foundation.Attribute(                                                                                                                                                                                                                                  
            attrid=Basic.AttributeDefs.location_desc.id,                                                                                                                                                                                                     
            value=foundation.TypeValue(
                type=Basic.AttributeDefs.location_desc.zcl_type,
                value=t.CharacterString("Test"),
            ),
        )
    ],
    manufacturer=None,
)
```

tests/test_zcl.py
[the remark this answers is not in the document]
TheJulianJES: Fixed with [`571f0ec` (this PR)](https://github.com/zigpy/zigpy/pull/1768/commits/571f0ec6db5ee72b6c921729cb50f26a7fa68c40)

tests/test_zcl.py
puddly: That's also pretty clear and a bit clearer than the individual asserts, IMO

TheJulianJES: This works nicely for the write attributes test, but not for the reporting test, as `AttributeReportingConfig` is not comparable (not a `Struct`, unlike `Attribute`). 

Apparently, we also have no tests for reporting that actually assert if min/max/reportable aren't messed up.
I guess we can do something like this for the reporting test..? 

```python
# First call: standard attribute (no manufacturer code)
std_call = mock_configure.call_args_list[0]
assert std_call.kwargs["manufacturer"] is None

assert len(std_call.args[0]) == 1
assert std_call.args[0][0].attrid == Basic.AttributeDefs.hw_version.id
assert std_call.args[0][0].min_interval == 5
assert std_call.args[0][0].max_interval == 15
assert std_call.args[0][0].reportable_change == 20
```

PR #1769, opened 2026-02-10
Always prefer manufacturer specific attributes when running a data migration

puddly:
When a manufacturer-specific attribute is defined in a quirk, it most likely is in use. If the old ZCL cache without manufacturer codes contained a value but there is no unique attribute definition we can match it up against (due to two attributes sharing an ID), we should pick the manufacturer-specific one over the ZCL-defined one. Currently, we do nothing and ignore this, which breaks the cache.

### DIFF
--- a/tests/test_appdb_migration.py
+++ b/tests/test_appdb_migration.py
@@ -663,21 +663,21 @@ async def test_data_migration_ambiguous_attributes(tmp_path):
         conn.execute(insert_sql, (str(dev.ieee), 1, 0, 0xFC02, 0x0020, b"\x99", 0))
         conn.commit()
 
-    # Migration runs during load: attr cache is populated in the same open
     with patch("zigpy.quirks.DEVICE_REGISTRY", registry):
         app = await make_app_with_db(db_path)
-
         dev = app.get_device(ieee=t.EUI64.convert("aa:bb:cc:dd:11:22:33:44"))
-        disambiguated = dev.endpoints[1].in_clusters[0xFC01]
-        ambiguous = dev.endpoints[1].in_clusters[0xFC02]
 
-        # 2 candidates (1 manuf + 1 non-manuf): picked manuf-specific
-        assert disambiguated.get("manuf_attr") == b"\x42"
-        assert disambiguated.get("standard_attr") is None
+    # Migration runs during load
+    disambiguated = dev.endpoints[1].in_clusters[0xFC01]
+    ambiguous = dev.endpoints[1].in_clusters[0xFC02]
 
-        # 3 candidates: ambiguous, skipped
-        assert ambiguous.get("attr_a") is None
-        assert ambiguous.get("attr_b") is None
-        assert ambiguous.get("attr_c") is None
+    # 2 candidates (1 manuf + 1 non-manuf): picked manuf-specific
+    assert disambiguated.get("manuf_attr") == b"\x42"
+    assert disambiguated.get("standard_attr") is None
 
-        await app.shutdown()
+    # 3 candidates: ambiguous, skipped
+    assert ambiguous.get("attr_a") is None
+    assert ambiguous.get("attr_b") is None
+    assert ambiguous.get("attr_c") is None
+
+    await app.shutdown()
--- a/tests/test_appdb_migration.py
+++ b/tests/test_appdb_migration.py
@@ -19,7 +19,7 @@ from zigpy.quirks.registry import DeviceRegistry
 from zigpy.quirks.v2 import QuirkBuilder
 import zigpy.types as t
 from zigpy.zcl.clusters.general import Basic
-from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
+from zigpy.zcl.foundation import BaseAttributeDefs, Status, ZCLAttributeDef
 from zigpy.zdo import types as zdo_t
 
 
@@ -550,13 +550,20 @@ async def test_unknown_manufacturer_code_migration(test_db, caplog):
 async def test_manufacturer_code_migration_uses_device_manufacturer_id(test_db):
     """Test that attributes on manufacturer-specific clusters get the device's manufacturer_id."""
 
-    # Simple quirk for Third Reality night light with is_manufacturer_specific=True
+    # Simple quirk for Third Reality night light with is_manufacturer_specific=True.
+    # The real device (f4:42:50:c3:96:14:00:00) has cached attrs 2-5 on 0xFC00 and
+    # attr 4 is also in unsupported_attributes_v13.
     class TestCluster(CustomCluster):
         cluster_id = 0xFC00
 
         class AttributeDefs(BaseAttributeDefs):
             test_attr = ZCLAttributeDef(
                 id=0x0002,
+                type=t.uint8_t,
+                is_manufacturer_specific=True,
+            )
+            unsupported_attr = ZCLAttributeDef(
+                id=0x0004,
                 type=t.uint8_t,
                 is_manufacturer_specific=True,
             )
@@ -570,12 +577,12 @@ async def test_manufacturer_code_migration_uses_device_manufacturer_id(test_db):
     )
 
     test_db_path = test_db("zigbee_puddly2.db")
+    third_reality_ieee = "f4:42:50:c3:96:14:00:00"
 
     with patch("zigpy.quirks.DEVICE_REGISTRY", registry):
         app = await make_app_with_db(test_db_path)
-        await app.shutdown()
 
-    # Check that attributes on 0xFC00 got the device's manufacturer_id (0x130D = 4877)
+    # Check that cached attributes on 0xFC00 got the device's manufacturer_id
     with sqlite3.connect(test_db_path) as conn:
         cur = conn.cursor()
         cur.execute(
@@ -583,11 +590,33 @@ async def test_manufacturer_code_migration_uses_device_manufacturer_id(test_db):
             SELECT manufacturer_code
             FROM attributes_cache_v14
             WHERE cluster_id = 0xFC00 AND attr_id = 0x0002
-            """
+            """,
         )
         rows = cur.fetchall()
 
     assert rows == [(0x130D,), (0x130D,), (0x130D,)]
+
+    # The unsupported manufacturer-specific attr also got the correct manufacturer code
+    with sqlite3.connect(test_db_path) as conn:
+        cur = conn.cursor()
+        cur.execute(
+            """
+            SELECT manufacturer_code, status
+            FROM attributes_cache_v14
+            WHERE ieee = ? AND cluster_id = 0xFC00 AND attr_id = 0x0004
+            """,
+            (third_reality_ieee,),
+        )
+        row = cur.fetchone()
+
+    assert row == (0x130D, Status.UNSUPPORTED_ATTRIBUTE)
+
+    # Confirm it's loaded as unsupported in the device's attribute cache
+    dev = app.get_device(ieee=t.EUI64.convert(third_reality_ieee))
+    cluster = dev.endpoints[1].in_clusters[0xFC00]
+    assert cluster.is_attribute_unsupported(TestCluster.AttributeDefs.unsupported_attr)
+
+    await app.shutdown()
 
 
 async def test_data_migration_ambiguous_attributes(tmp_path):
--- a/tests/test_appdb_migration.py
+++ b/tests/test_appdb_migration.py
@@ -616,7 +616,7 @@ async def test_manufacturer_code_migration_uses_device_manufacturer_id(test_db):
     cluster = dev.endpoints[1].in_clusters[0xFC00]
     assert cluster.is_attribute_unsupported(TestCluster.AttributeDefs.unsupported_attr)
 
-    # Attr 0x0003 has no definition in our quirk but exist in the DB. It should be
+    # Attr 0x0003 has no definition in our quirk but exists in the DB. It should be
     # stored in the legacy cache regardless.
     with pytest.raises(KeyError):
         cluster.find_attribute(0x0003)
--- a/tests/test_appdb_migration.py
+++ b/tests/test_appdb_migration.py
@@ -616,6 +616,14 @@ async def test_manufacturer_code_migration_uses_device_manufacturer_id(test_db):
     cluster = dev.endpoints[1].in_clusters[0xFC00]
     assert cluster.is_attribute_unsupported(TestCluster.AttributeDefs.unsupported_attr)
 
+    # Attr 0x0003 has  no definition in our quirk but exist in the DB. It should be
+    # stored in the legacy cache regardless.
+    with pytest.raises(KeyError):
+        cluster.find_attribute(0x0003)
+
+    assert 0x0003 in cluster._attr_cache._legacy_cache
+    assert cluster._attr_cache.get(0x0003) == 20
+
     await app.shutdown()
 
 
--- a/tests/test_appdb_migration.py
+++ b/tests/test_appdb_migration.py
@@ -7,15 +7,18 @@ from aiosqlite.context import contextmanager
 import pytest
 
 from tests.async_mock import AsyncMock, MagicMock, patch
-from tests.conftest import app  # noqa: F401
+from tests.conftest import app, make_node_desc  # noqa: F401
 from tests.test_appdb import auto_kill_aiosqlite, make_app_with_db  # noqa: F401
 import zigpy.appdb
 from zigpy.appdb import sqlite3
 import zigpy.appdb_schemas
+import zigpy.endpoint
+from zigpy.profiles import zha as zha_profile
 from zigpy.quirks import CustomCluster
 from zigpy.quirks.registry import DeviceRegistry
 from zigpy.quirks.v2 import QuirkBuilder
 import zigpy.types as t
+from zigpy.zcl.clusters.general import Basic
 from zigpy.zcl.foundation import BaseAttributeDefs, ZCLAttributeDef
 from zigpy.zdo import types as zdo_t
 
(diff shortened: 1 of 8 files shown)


puddly:
This PR turned into a much larger one due to a bug I found while writing a unit test for the migration: the current data migrations are run _after_ we load attributes, so they are not available at runtime, only after ZHA is reloaded! I've reworked the migration logic.

I've rewritten the data migration to instead run implicitly within `load` and be made part of loading attributes. We also fully clear the attribute caches before re-populating them for quirks.

Currently depends on https://github.com/zigpy/zigpy/pull/1770.

zigpy/appdb.py
```diff
@@ -657,26 +658,51 @@ async def _network_backup_removed(self, backup_time: datetime) -> None:
...
         await self._load_endpoints()
         await self._load_clusters()
 
-        # Load as many attributes as we can in the first pass
-        await self._load_attributes()
+        # Read all attribute rows from the database once
+        all_attributes = await self._read_all_attributes()
+
+        # First pass: populate cache on bare clusters for quirks
+        await self._populate_attribute_cache(all_attributes)
```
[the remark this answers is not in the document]
puddly: Total memory usage is insignificant for a normal Zigbee network.

tests/test_zcl.py
```diff
@@ -1676,6 +1676,73 @@ class AttributeDefs(zcl.BaseAttributeDefs):
...
+    ]
+    assert TestCluster.find_attributes("manuf_1234") == [
+        TestCluster.AttributeDefs.manuf_1234,
+    ]
+    assert TestCluster.find_attributes(TestCluster.AttributeDefs.manuf_5678) == [
+        TestCluster.AttributeDefs.manuf_5678,
+    ]
+
+    with pytest.raises(KeyError):
+        TestCluster.find_attributes(0x9999)
```
TheJulianJES, line 1755: If there's only a manufacturer specific attribute for an ID, is it expected for `find_attributes` to return an empty list when checking for the ID with a different `manufacturer_code`, or should it also raise `KeyError`?
E.g., this passes currently:
```diff
diff --git a/tests/test_zcl.py b/tests/test_zcl.py
index 5b3883f8..edd3865a 100644
--- a/tests/test_zcl.py
+++ b/tests/test_zcl.py
@@ -1700,6 +1700,12 @@ def test_find_attributes_non_manuf_specific_undefined() -> None:
             manuf_5678 = foundation.ZCLAttributeDef(
                 id=0x0001, type=t.EUI64, manufacturer_code=0x5678
             )
+            only_explicit = foundation.ZCLAttributeDef(
+                id=0x0002, type=t.EUI64, manufacturer_code=0x1234
+            )
+
+    assert TestCluster.find_attributes(0x0002, manufacturer_code=0x5678) == []
+    assert TestCluster.find_attributes(0x0002, manufacturer_code=None) == []
 
     assert TestCluster.find_attributes(0x0001, manufacturer_code=None) == [
         TestCluster.AttributeDefs.explicit_none,
```
(Though `find_attribute` will raise a `KeyError(mf_code)` when it gets an empty list from `find_attributes`)

zigpy/appdb.py
```diff
@@ -61,6 +62,10 @@
 # if the device has "colliding" attributes.
 UNMIGRATED_MANUFACTURER_CODE = -1
 
+# ieee, endpoint_id, cluster_type, cluster_id, attr_id, manufacturer_code
+# manufacturer_code_idx, status, value, last_updated
+AttributeRow = tuple[t.EUI64, int, int, int, int, int | None, int, Any, float]
```
TheJulianJES: nit: Do you think this should be a `NamedTuple` at this point?

tests/test_zcl.py
puddly: I originally had `find_attributes` return a `KeyError` but I then moved it into `find_attribute`. In my mind, finding candidates "fails" by having no candidates (an empty list)? Otherwise, the two function the same but you need exception handling to deal with the empty list case.

zigpy/appdb.py
```diff
@@ -687,111 +719,193 @@ async def load(self) -> None:
...
+            return UNDEFINED
+
+        if len(attr_defs) == 1:
+            attr_def = attr_defs[0]
+        elif (
+            len(attr_defs) == 2
+            and attr_defs[0].is_manufacturer_specific
+            != attr_defs[1].is_manufacturer_specific
+        ):
+            attr_def = next(a for a in attr_defs if a.is_manufacturer_specific)
```
TheJulianJES: I'm not sure if this is actually an issue in practice, but imagine the following scenario:
On a manufacturer-specific cluster (`cluster_id=0xFC00`), there are two attributes defined:
1. The first one is defined both without `is_manufacturer_specific` (so default: `None`) and `manufacturer_code` (so default: `UNDEFINED`) .
2. The second is explicitly defined with `is_manufacturer_specific=False` (or I guess `manufacturer_code=None`).

The first attribute should always implicitly use the manufacturer code of that device, since it's on a manufacturer cluster. The second one, however, should not, as it's explicitly overridden not to.

The condition would evaluate `None` != `False` `== True`, but `next()` wouldn't have any attributes where `is_manufacturer_specific=True`, right?

I think something like this could solve it, but I'm not sure if it's even enough to just look at `is_manufacturer_specific`. (see text below code)

```python
manuf_attrs = [a for a in attr_defs if a.is_manufacturer_specific]

if len(attr_defs) == 1:
    attr_def = attr_defs[0]
elif len(attr_defs) == 2 and len(manuf_attrs) == 1:
    attr_def = manuf_attrs[0]
```

--- 

The attribute would always implicitly use the manufacturer code in practice, so should we do something along these lines?

```python
manuf_attrs = [
    a
    for a in attr_defs
    if cluster._get_effective_manufacturer_code(a) is not None
]

if len(attr_defs) == 1:
    attr_def = attr_defs[0]
elif len(attr_defs) == 2 and len(manuf_attrs) == 1:
    attr_def = manuf_attrs[0]
```

tests/test_zcl.py
TheJulianJES: Ok, fine with me. Should we test that behavior then? (of `find_attributes` potentially returning an empty list)

zigpy/appdb.py
puddly: You're right, I think your last code snippet is the correct approach.

tests/test_zcl.py
puddly: We actually already raise `KeyError` here in other circumstances so I take that back, an empty list doesn't make sense. I've moved it back as per your suggestion and explicitly test this scenario.

puddly:
https://github.com/zigpy/zigpy/pull/1766 depends on the data migration fixes in this PR

zigpy/appdb.py
```diff
@@ -657,26 +676,53 @@ async def _network_backup_removed(self, backup_time: datetime) -> None:
...
+    ) -> list[AttributeCacheRow]:
+        """Read all attribute rows from the database."""
+        async with self.execute(
+            f"""
+            SELECT ieee, endpoint_id, cluster_type, cluster_id, attr_id,
+                   manufacturer_code, status, value, last_updated
+            FROM attributes_cache{DB_V}
+            """
+        ) as cursor:
+            return await cursor.fetchall()
```
TheJulianJES: nit: With the return typing being a list of the named tuples (now), does this need to be changed slightly to actually use the `AttributeCacheRow`?
```python
return [AttributeCacheRow(*row) for row in await cursor.fetchall()]
```

puddly: Ah, that would be helpful. Thanks 😄

zigpy/appdb.py
TheJulianJES, line 1527: This is supposed to stay as is and be fixed with v15, right? (even with #1766)

puddly: Yeah, modifying already-deployed migrations wouldn't make much sense because the changes have already been applied to real databases. It would fragment things.

zigpy/appdb.py
```diff
@@ -687,111 +733,195 @@ async def load(self) -> None:
...
+            ieee,
+            endpoint_id,
+            cluster_type,
+            cluster_id,
+            attr_id,
+            manufacturer_code,
+            status,
+            value,
+            last_updated,
+        ) in rows:
```
TheJulianJES: nit: Now that the contents of `rows` are named tuples, we could just do `for row in rows:` and later access `row.ieee` and so on. I guess in the very unlikely scenario another field is added in the middle of it, we wouldn't run the risk of having a wrong order here? But I don't care, really..

zigpy/appdb.py
```diff
@@ -687,111 +733,195 @@ async def load(self) -> None:
...
+                                else manufacturer_code
+                            ),
+                        )
                 except KeyError:
                     LOGGER.debug("Unknown ZCL attribute, skipping")
-                    cluster._attr_cache.set_legacy_value(
-                        attr_id,
-                        value,
-                        last_updated=datetime.fromtimestamp(last_updated, UTC),
+                continue
```
TheJulianJES: So we don't migrate unsupported attributes to the correct manufacturer code at all.

We could theoretically use a similar logic to the normal runtime attribute migration (migrate when there's one attribute definition only OR if two, use the manufacturer specific one, else skip), right?
(But of course only if there's no value present already! They should take priority then.)

Is this something we want to do with https://github.com/zigpy/zigpy/pull/1766/, or should this just stay as is?

puddly: I've consolidated the two codepaths and added a test with a database file. Thankfully, one of the test databases contained a manufacturer-specific attribute that was marked as unsupported.

zigpy/appdb.py
```diff
@@ -687,111 +732,183 @@ async def load(self) -> None:
...
+                LOGGER.debug("Unknown ZCL cluster, skipping")
+                continue
 
-                cluster = clusters[cluster_id]
+            cluster = clusters[row.cluster_id]
 
-                # Handle unsupported attributes
-                if status != Status.SUCCESS:
-                    try:
-                        with suppress_events():
```
TheJulianJES, line 747: `suppress_events` seems to be completely unused now, but might still be nice to keep around?
(So either remove or add (now missing) test coverage?)

zigpy/appdb.py
```diff
@@ -687,111 +732,183 @@ async def load(self) -> None:
...
+
+                if row.status == Status.SUCCESS:
                     cluster._attr_cache.set_legacy_value(
-                        attr_id,
-                        value,
-                        last_updated=datetime.fromtimestamp(last_updated, UTC),
+                        row.attr_id,
+                        row.value,
+                        last_updated=datetime.fromtimestamp(row.last_updated, UTC),
                     )
```
TheJulianJES, line 839: Is there a test that checks if the value is correctly stored in the legacy cache then? Maybe there's the Aqara P1 sensor in your DB with the attribute with no definition at all?

TheJulianJES: To answer myself: If this is commented out, the following test fails:
```python
tests/test_appdb_migration.py::test_v4_to_v6_migration_missing_endpoints[True] - KeyError: 789
```

Should we have a bit more of an explicit test for this then?

puddly: I've added one for the same Hue device.

zigpy/appdb.py
puddly: Ah. That explains the CI failure. I'll add a quick test.

tests/test_appdb_migration.py
```diff
@@ -567,21 +577,144 @@ class AttributeDefs(BaseAttributeDefs):
...
+        row = cur.fetchone()
+
+    assert row == (0x130D, Status.UNSUPPORTED_ATTRIBUTE)
+
+    # Confirm it's loaded as unsupported in the device's attribute cache
+    dev = app.get_device(ieee=t.EUI64.convert(third_reality_ieee))
+    cluster = dev.endpoints[1].in_clusters[0xFC00]
+    assert cluster.is_attribute_unsupported(TestCluster.AttributeDefs.unsupported_attr)
+
+    # Attr 0x0003 has  no definition in our quirk but exist in the DB. It should be
```
TheJulianJES: nit: double space and *exists* 😄 
```suggestion
    # Attr 0x0003 has no definition in our quirk but exists in the DB. It should be
```

tests/test_zcl.py
```diff
@@ -1676,6 +1676,86 @@ class AttributeDefs(zcl.BaseAttributeDefs):
         TestCluster.find_attribute(0x0003, manufacturer_code=0x5678)
 
 
+def test_find_attributes() -> None:
+    """Test find_attributes across all attribute specificity combinations."""
+
+    class TestCluster(zcl.Cluster):
+        cluster_id = 0xABCD
+        ep_attribute = "test_cluster"
```
TheJulianJES, line 1685: nit: Should this also have `_skip_registry = True`?

puddly: Yeah, we had a lot of these. I ran into test determinism issues when running with xdist. I'll fix up the few that remain.

zigpy/appdb.py
```diff
@@ -687,111 +732,183 @@ async def load(self) -> None:
...
+                    manufacturer_code=(
+                        UNDEFINED
+                        if manufacturer_code == UNMIGRATED_MANUFACTURER_CODE
+                        else manufacturer_code
+                    ),
+                )
+            except KeyError:
+                LOGGER.debug("Unknown ZCL attribute, skipping")
+
+                if row.status == Status.SUCCESS:
```
TheJulianJES, line 834: It should be fine to completely ignore unsupported attributes for unknown attributes, right? (or for colliding ones which we can't resolve) (as done here)

zigpy/appdb.py
```diff
@@ -687,111 +732,183 @@ async def load(self) -> None:
...
+        manuf_attrs = [
+            a
+            for a in attr_defs
+            if cluster._get_effective_manufacturer_code(a) is not None
+        ]
+
+        if len(attr_defs) == 1:
+            attr_def = attr_defs[0]
+        elif len(attr_defs) == 2 and len(manuf_attrs) == 1:
+            attr_def = manuf_attrs[0]
```
TheJulianJES, line 905: nit: I guess this should be clear, but we could add a comment what we're doing here. E.g.:
```suggestion
        if len(attr_defs) == 1:
            attr_def = attr_defs[0]
        elif len(attr_defs) == 2 and len(manuf_attrs) == 1:
            # One standard + one manufacturer-specific: prefer manufacturer-specific
            attr_def = manuf_attrs[0]
```

zigpy/appdb.py
```diff
@@ -687,111 +732,183 @@ async def load(self) -> None:
...
+        cluster: Cluster,
+        attr_id: int,
+        value: Any,
+        dev: Device,
+    ) -> int | None | zigpy.typing.UndefinedType:
+        """Try to resolve the manufacturer code for an unmigrated attribute.
+
+        Returns the resolved manufacturer code (including `None`), or UNDEFINED if
+        unresolvable.
+        """
```
TheJulianJES, line 881: nit: Or alternatively, we can add it to the docstring. Or, just do neither of these if it's clear enough.
```suggestion
        """Try to resolve the manufacturer code for an unmigrated attribute.

        When multiple attribute definitions share an ID, the manufacturer-specific one
        is preferred if there are exactly two candidates (one standard, one
        manufacturer-specific). Otherwise the attribute is considered ambiguous and
        skipped (left unmigrated).

        Returns the resolved manufacturer code (including `None`), or UNDEFINED if
        unresolvable.
        """
```

zigpy/appdb.py
TheJulianJES: nit: I guess we could add a comment or something about that?
```suggestion
            except KeyError:
                LOGGER.debug("Unknown ZCL attribute, skipping")

                # Unsupported unknown attributes are dropped
                if row.status == Status.SUCCESS:
```
But I don't really see this being an issue at all.

TheJulianJES (commented):
nit: Could add a comment somewhere about the manufacturer-specific attribute taking priority over a standard one when migrating. But optional, of course.

TheJulianJES (approved):
This small fix looks good to me 😄 . Really nice work!
The last few comments are only about potentially adding some comments in the code.

zigpy/appdb.py
```diff
@@ -687,111 +732,190 @@ async def load(self) -> None:
...
+                        {
+                            "manufacturer_code": manufacturer_code,
+                            "ieee": row.ieee,
+                            "endpoint_id": row.endpoint_id,
+                            "cluster_type": row.cluster_type,
+                            "cluster_id": row.cluster_id,
+                            "attr_id": row.attr_id,
+                            "old_manufacturer_code": UNMIGRATED_MANUFACTURER_CODE,
+                        },
                     )
```
TheJulianJES, line 819: @puddly Tried to start HA with my v14 database, where I previously already read the Hue sensor's sensitivity attribute through the UI (with providing `manufacturer_code=0` in the UI, causing a read with the Hue manufacturer code to how that currently works). I got this in the logs:

```python
2026-02-12 01:54:11.685 DEBUG (MainThread) [zigpy.appdb] [0xefc7:2:0x0000] Loading attribute 0x0004='Signify Netherlands B.V.' status=0 mfg_code=None
2026-02-12 01:54:11.685 DEBUG (MainThread) [zigpy.appdb] [0xefc7:2:0x0000] Loading attribute 0x0005='SML003' status=0 mfg_code=None
2026-02-12 01:54:11.685 DEBUG (MainThread) [zigpy.appdb] [0xefc7:2:0x0406] Loading attribute 0x0030=4 status=0 mfg_code=-1

# mostly not very helpful trace
# with IntegrityError at the end:
sqlite3.IntegrityError: UNIQUE constraint failed: attributes_cache_v14.ieee, attributes_cache_v14.endpoint_id, attributes_cache_v14.cluster_type, attributes_cache_v14.cluster_id, attributes_cache_v14.attr_id, attributes_cache_v14.manufacturer_code_idx
```

v14 (before starting HA with this patch):
Image

v13:
Image

---

To fix this, could we just change this to `UPDATE OR IGNORE`? Afterwards, just always delete any rows for that `attrid` with `UNMIGRATED_MANUFACTURER_CODE` to clean up old/unmigrated entries?

TheJulianJES: E.g. something like this maybe?
```diff
diff --git a/zigpy/appdb.py b/zigpy/appdb.py
index 47450b8e..b34023ce 100644
--- a/zigpy/appdb.py
+++ b/zigpy/appdb.py
@@ -795,9 +795,11 @@ class PersistingListener(zigpy.util.CatchingTaskMixin):
                 if resolved_manufacturer_code is not UNDEFINED:
                     manufacturer_code = resolved_manufacturer_code
 
+                    # Update the manufacturer code. If a row with the resolved
+                    # code already exists, OR IGNORE silently skips the update.
                     await self.execute(
                         f"""
-                        UPDATE attributes_cache{DB_V}
+                        UPDATE OR IGNORE attributes_cache{DB_V}
                         SET manufacturer_code = :manufacturer_code
                         WHERE
                             ieee = :ieee
@@ -818,6 +820,30 @@ class PersistingListener(zigpy.util.CatchingTaskMixin):
                         },
                     )
 
+                    # Delete the stale unmigrated row if it still exists (i.e.
+                    # the UPDATE was skipped because the resolved row already
+                    # existed).
+                    await self.execute(
+                        f"""
+                        DELETE FROM attributes_cache{DB_V}
+                        WHERE
+                            ieee = :ieee
+                            AND endpoint_id = :endpoint_id
+                            AND cluster_type = :cluster_type
+                            AND cluster_id = :cluster_id
+                            AND attr_id = :attr_id
+                            AND manufacturer_code = :old_manufacturer_code
+                        """,
+                        {
+                            "ieee": row.ieee,
+                            "endpoint_id": row.endpoint_id,
+                            "cluster_type": row.cluster_type,
+                            "cluster_id": row.cluster_id,
+                            "attr_id": row.attr_id,
+                            "old_manufacturer_code": UNMIGRATED_MANUFACTURER_CODE,
+                        },
+                    )
+
             try:
                 attr_def = cluster.find_attribute(
                     row.attr_id,
```

puddly: Interesting, I'm surprised this wasn't caught by the tests. So your current v14 database contains _two_ rows for the manufacturer-specific attribute? Are the SQLite `UNIQUE` constraints too permissive somewhere and allowing a second row to sneak in?

TheJulianJES: > So your current v14 database contains two rows for the manufacturer-specific attribute?

Well, the one with `manufacturer_code = -1` in the v14 screenshot is the old v13 value that couldn't be migrated, as there were two attributes with colliding IDs – that would only migrate when now running this PR, which is where we fail.
The other row with the Hue manufacturer code was read manually using the Zigbee UI. That doesn't modify/delete the unmigrated row (apparently?).

TheJulianJES: Like, the new migration from this PR then tried to migrate the row with the unmigrated manufacturer code. So, it tries changing the `manufacturer_code` from `-1` to `4107`. Suddenly, it's the same as the other row, so we fail, as we'd have a duplicate unique index entry then.

PR #1773, opened 2026-02-12
Fix v14 startup when the database contains new rows with a manufacturer code

puddly:
Instead of catching the integrity error from the `UPDATE` statement, we can just delete the existing row and maybe insert a new one.

### DIFF
--- a/tests/test_appdb_migration.py
+++ b/tests/test_appdb_migration.py
@@ -699,7 +699,8 @@ async def test_data_migration_ambiguous_attributes(tmp_path):
                 # Unmigrated row from v13->v14 (ambiguous cluster)
                 (str(dev.ieee), 1, 0, 0xFC02, 0x0020, -1, b"\x99", 0),
                 # Manually read through the UI after v13->v14, duplicating the above
-                (str(dev.ieee), 1, 0, 0xFC01, 0x0010, 0xABCD, b"\x42", 1.0),
+                # but with a newer value
+                (str(dev.ieee), 1, 0, 0xFC01, 0x0010, 0xABCD, b"\x43", 1.0),
             ],
         )
         conn.commit()
@@ -712,8 +713,9 @@ async def test_data_migration_ambiguous_attributes(tmp_path):
     disambiguated = dev.endpoints[1].in_clusters[0xFC01]
     ambiguous = dev.endpoints[1].in_clusters[0xFC02]
 
-    # 2 candidates (1 manuf + 1 non-manuf): picked manuf-specific
-    assert disambiguated.get("manuf_attr") == b"\x42"
+    # 2 candidates (1 manuf + 1 non-manuf): picked manuf-specific.
+    # Uses the newer value from the already-resolved row, not stale/unmigrated value.
+    assert disambiguated.get("manuf_attr") == b"\x43"
     assert disambiguated.get("standard_attr") is None
 
     # 3 candidates: ambiguous, skipped
--- a/tests/test_appdb_migration.py
+++ b/tests/test_appdb_migration.py
@@ -687,17 +687,21 @@ async def test_data_migration_ambiguous_attributes(tmp_path):
     app.device_initialized(dev)
     await app.shutdown()
 
-    # Insert unmigrated attribute rows (manufacturer_code=-1)
     with sqlite3.connect(db_path) as conn:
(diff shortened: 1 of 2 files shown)


tests/test_appdb_migration.py
```diff
@@ -687,17 +687,21 @@ class AttributeDefs(BaseAttributeDefs):
...
-        insert_sql = (
+        conn.executemany(
             "INSERT INTO attributes_cache_v14"
             " (ieee, endpoint_id, cluster_type, cluster_id,"
             "  attr_id, manufacturer_code, status, value, last_updated)"
-            " VALUES (?, ?, ?, ?, ?, -1, 0, ?, ?)"
+            " VALUES (?, ?, ?, ?, ?, ?, 0, ?, ?)",
+            [
+                # Unmigrated row from v13->v14 (disambiguated cluster)
+                (str(dev.ieee), 1, 0, 0xFC01, 0x0010, -1, b"\x42", 0),
```
TheJulianJES, line 698: nit: Should the test also check that this unmigrated row is later deleted?

TheJulianJES (commented):
Nice, this is cleaner!

zigpy/appdb.py
```diff
@@ -808,16 +819,31 @@ async def _populate_attribute_cache(
...
+                             :last_updated)
+                        """,
+                        {
+                            **row_params,
+                            "manufacturer_code": manufacturer_code,
+                            "status": row.status,
+                            "value": row.value,
+                            "last_updated": row.last_updated,
+                        },
+                    )
```
TheJulianJES: Should there be a `continue` at the end of this block? Otherwise, the (deleted) row will still land in the runtime cache until zigpy is reloaded, right?

TheJulianJES: But just a `continue` is too strict, since that would not load the attribute IF we re-inserted it with the correct manufacturer code. I guess it kind of depends on the output of the `INSERT OR IGNORE` query whether we should skip this entry for runtime cache population or not?

EDIT: Ah, this is basically what Copilot already commented on 😄

TheJulianJES: I guess something along these lines should work: https://github.com/zigpy/zigpy/compare/puddly:puddly/fix-v14-startup...TheJulianJES:tjj/puddly/fix-v14-startup

puddly: Ah, that's exactly what I had written 😁

TheJulianJES (approved):
Looks good to me!

zigpy/appdb.py
```diff
@@ -808,16 +819,34 @@ async def _populate_attribute_cache(
...
+                            **row_params,
+                            "manufacturer_code": manufacturer_code,
+                            "status": row.status,
+                            "value": row.value,
+                            "last_updated": row.last_updated,
+                        },
+                    ) as cursor:
+                        # A resolved row already exists, it will populate the cache
+                        if cursor.rowcount == 0:
+                            continue
```
TheJulianJES, line 848: nit: Just wondering if we should also have debug logging here? Or more for this entire migration block in general, since it should only happen once? But I guess it should also be relatively predictable, so likely doesn't matter.

zigpy/appdb.py
TheJulianJES, line 724: nit: Since we already have some logging in `_populate_attribute_cache` for each attribute, it could be nice to have a separation for first pass vs second pass. Right now, you don't really see when the second attribute cache loading happens(?)

So, either log here in `load` or have a single message at the start of `_populate_attribute_cache` that also logs whether `migrate` is `True` or not.

EDIT: I guess we do have the quirk matching logging in between.

puddly: The logging right now is _super_ verbose already, since we basically dump the whole attribute cache twice in a row. I was thinking of even hiding it behind `log=True` so we would only log on the second pass?

TheJulianJES: Yeah, agreed. Should be fine unless something errors out on the first try then 😄 
It's not urgent to address now IMO

puddly: I was thinking of some ways to clean it up, in a follow-up PR. We could log per cluster instead of per attribute. This would hurt CLI filtering via `grep` but it would reduce the amount of log spam (though a lot comes from loading `zhaquirks`).

PR #1774, opened 2026-02-12
Allow accessing legacy cache attributes via `cluster.get`

puddly:
The recent attribute cache consolidation/rewrite introduced [a regression](https://github.com/zigpy/zigpy/pull/1760#discussion_r2800974172): undefined attributes no longer could be accessed via `cluster.get`.

This PR fixes the behavior and aligns `.get()` with other methods that accept `int | str | ZCLAttributeDef` and updates it to correctly interface with the underlying attribute cache for the cluster.

### DIFF
--- a/zigpy/zcl/helpers.py
+++ b/zigpy/zcl/helpers.py
@@ -98,20 +98,23 @@ class AttributeCache:
             last_updated=datetime.now(UTC) if last_updated is None else last_updated,
         )
 
-    def get(self, key: int, default: Any | None = None) -> Any:
+    def get(self, key: int | str | ZCLAttributeDef, default: Any | None = None) -> Any:
         try:
             return self[key]
         except (KeyError, UnsupportedAttribute):
             return default
 
-    def __getitem__(self, key: int) -> Any:
+    def __getitem__(self, key: int | str | ZCLAttributeDef) -> Any:
         try:
-            return self.get_value(self._cluster.find_attribute(key))
+            attr_def = self._cluster.find_attribute(key)
         except KeyError:
             # Fall back to legacy cache for unknown attributes
-            if key in self._legacy_cache:
-                return self._legacy_cache[key].value
-            raise
+            if not isinstance(key, int):
+                raise
+
+            return self._legacy_cache[key].value
+        else:
+            return self.get_value(attr_def)
 
     def __setitem__(self, key: int, value: Any) -> None:
         attr_def = self._cluster.find_attribute(key)
--- a/tests/test_quirks.py
+++ b/tests/test_quirks.py
@@ -405,10 +405,8 @@ async def test_read_attributes_uncached():
 
         return [[responses[attr_id] for attr_id in args]]
 
-    # Unknown attribute read passes through
-    with pytest.raises(KeyError):
-        cluster.get("unknown_attribute", 123)
-
+    # Unknown attribute name returns the default
+    assert cluster.get("unknown_attribute", 123) == 123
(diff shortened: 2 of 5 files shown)


zigpy/zcl/helpers.py
```diff
@@ -98,20 +98,23 @@ def set_legacy_value(
...
+    def __getitem__(self, key: int | str | ZCLAttributeDef) -> Any:
         try:
-            return self.get_value(self._cluster.find_attribute(key))
+            attr_def = self._cluster.find_attribute(key)
         except KeyError:
             # Fall back to legacy cache for unknown attributes
-            if key in self._legacy_cache:
-                return self._legacy_cache[key].value
-            raise
+            if not isinstance(key, int):
```
puddly, line 112: There is a bit of inconsistency with one existing unit test but I don't think we ever allowed persisting "legacy" attributes with string names, right?

TheJulianJES:
I do kind of wonder: What (useful) values actually land in the `_legacy_cache` at all now and where are they accessed?
Currently, ZHA entities all do not use the legacy cache because of this regression and I don't think I've seen a single issue that's related to this?

So, I wonder if we should just try to keep this..? Because if we re-introduce this, even with https://github.com/TheJulianJES/zigpy/compare/tjj/manufacturer_update_fix...TheJulianJES:zigpy:tjj/manufacturer_update_fix_with_mfr_attr_fix_zcl_cache (which fixes the issue of mf-specific reports with ZCL IDs landing in ZCL cache, instead of only in legacy cache), we'll then have ZHA entities use `cluster.get` and accessing mf-specific attribute values that overlap with ZCL attribute IDs if there's not an attribute definition for them, even though they (correctly) only landed in the legacy cache.
(Maybe we can also just block the legacy cache entirely for attribute reports and only allow it if quirks do updates?..)

But I guess we can rethink this at a later point. The involuntary beta testing by everyone was nice to see we may not rely on this at all though. I think this should be fine, for now. I'll give it a proper look in a minute.

puddly:
My original idea for the legacy cache was based on the (as it turns out, mostly false) assumption that quirks frequently used patterns like `cluster._update_attribute(0xABCD, "some value")` to write data to the database and then relied on it being readable via `cluster.get(0xABCD)` after restart, with there being no corresponding attribute definition. We could then migrate these to a generic "key/value" table separate from the ZCL attribute cache.

It turns out practically all quirks these days actually register their virtual attributes normally and thus "bypass" the legacy cache. They use the attribute system properly, just not with attributes that really exist.

I guess next release we can try to introduce some better logging for truly "legacy" attribute access patterns and then deprecate it entirely? This would really simplify the attribute cache.

puddly:
If nothing is really broken without this PR, let's leave things as is then.

TheJulianJES:
Yeah, I think the additional logging for everything that still ends up in the legacy cache now, and an eventual key/value migration sound sensible. But with https://github.com/zigpy/zigpy/pull/1769, I can't think of anything "really used" that would still end up in the legacy cache now. 

> If nothing is really broken without this PR, let's leave things as is then.

I mean, I'm not 100% sure, but all quirks I can think of do use "fake attribute definitions" then, most on `LocalDataCluster`s.
For the Aqara P1 sensor that's missing one attribute definition, we should eventually add one. But even there, it's only used to forward/update attributes on other clusters, so I guess that's why it was never added in the first place — just forgotten. Like, it's not directly used by any entity and everything with that works properly.

So I guess we can skip this PR for now?

PR #1776, opened 2026-02-19
Implement structured attribute reading and writing

puddly:
We've finally run across a device that requires this functionality!

I've extended `t.Struct` with a few new features:
- `StructField(length=lambda s: s.other_field)`, which allows array fields to have a length depend on earlier fields.
- `StructField(default=123)`, giving default values to fields.

This allows the complex operations encoded in the `Selector` object be relatively straightforward to use. These operations are rarely used so they are exposed as low-level `_raw` cluster methods:

- `read_attributes_structured_raw`
- `write_attributes_structured_raw`
- `write_attributes_raw` (this should have existed already)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -415,7 +415,7 @@ disable_error_code = [
 [[tool.mypy.overrides]]
 module = "zigpy.types.struct"
 disable_error_code = [
-    "arg-type",  # 2
+    "arg-type",  # 3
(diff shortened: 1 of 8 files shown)


puddly:
I've added some more explicit typing for the ZCL foundation responses, especially the default response. We can migrate to using these types more broadly in the future.

PR #1779, opened 2026-02-27
Persist OTA `query_next_image` fields in appdb

TheJulianJES:
## Note

An AI summary is below but it's adjusted to be accurate. 😄 

The main problem this tries to solve is the issue of all ZHA entities being "Unknown" after HA is restarted, with devices needing to send a `query_img_command` before zigpy fetches all possible OTA images, finally causing a proper state to be shown in HA. This can take hours, e.g. when zigpy sends its first OTA notify broadcast, where not even all devices check-in.

Do note that, if we like to change this behavior in the future, we can always drop the table and don't really need to care about migrating. The only issue is that it'll take some time before all the updates show again.
This is just a (very) nice to have feature. Nothing critical, like the attribute cache.

## Problem

To proactively fetch all possible OTA images for a device, zigpy needs the fields from the device's `QueryNextImageCommand`: `manufacturer_code`, `image_type`, `current_file_version`, and `hardware_version`. Currently, these values are only used transiently when a device checks in, and are not stored anywhere.

While the OTA cluster does define ZCL attributes for some of these fields (`manufacturer_id`, `image_type_id`, `current_file_version`), they **cannot** be reliably used:

- **Some devices return incorrect values** when these attributes are read directly, differing from what they report in the `query_next_image` command.
- **No ZCL attribute exists for `hardware_version`**, which is needed by `check_compatibility` to match images with hardware version constraints.

## Solution

Store the last received `QueryNextImageCommand` fields in a **new dedicated `ota_query_cache` table** in the application database, keyed by `(ieee, endpoint_id)`.

### How it works

1. When a device sends a `query_next_image` command (in response to `image_notify`), the `Ota` cluster now caches the command on `self.last_query_cmd` and emits an `OtaQueryCacheUpdatedEvent`.
2. The `PersistingListener` in `appdb.py` handles this event and upserts the fields into the `ota_query_cache` table.
3. On database load, stored query commands are reconstructed and placed back on the appropriate `Ota` cluster instances.
4. Callers can use `device.get_last_ota_query_cmd()` to retrieve the cached command for proactive OTA image lookups via `ota.get_ota_images(device, query_cmd)`.
5. After a successful firmware update, `last_query_cmd` is cleared so stale version info is not reused.

### Impact on ZHA

The existing `device_ota_image_query_result` listener event flow is unchanged. ZHA continues to work as-is.
(`OtaQueryCacheUpdatedEvent` is emitted immediately when the cache is updated. `get_ota_images` is called by zigpy afterwards, so we can't replace it with the listener event. But we could introduce another event or something, possibly in another PR.)

This PR enables ZHA to **proactively check for OTA images on startup** instead of waiting for each device to check in. Currently, after a ZHA restart, the update entity has no firmware availability information until the device sends a `query_next_image` (spontaneously or in response to `image_notify`). With the persisted query cache, ZHA could populate this immediately by reusing the existing `device_ota_image_query_result` handler:

```python
# In FirmwareUpdateEntity.on_add or similar:
query_cmd = self.device.device.get_last_ota_query_cmd()
if query_cmd is not None:
    images_result = await self.device.device.application.ota.get_ota_images(
        self.device.device, query_cmd
    )
    # Reuse the existing handler with cached data
    self.device_ota_image_query_result(images_result, query_cmd)
```

### Further thoughts for ZHA/HA

Above code would fetch info from providers on startup but not download any new images for trusted providers (zigpy-ota)  with pending PR https://github.com/zigpy/zigpy/pull/1749.
The Matter integration has a different implementation where the [update entity restores](https://github.com/home-assistant/core/blob/e8a35ea69de70f2d98f1996ac6eef0242314b635/homeassistant/components/matter/update.py#L179-L184) a [lot of data](https://github.com/matter-js/python-matter-server/blob/9637a985ed5946efc698abb5901f22d6436f91cf/matter_server/common/models.py#L228-L273). The Matter server itself must fetch DCL data later.

So alternatively, I'm wondering if we should also immediately restore old data in Home Assistant. I guess the update entities would have the correct state maybe a second or two sooner but that likely doesn't matter (no need to fetch the index).
Then, ZHA or zigpy could be modified to also call `get_ota_images` in `update_firmware` (which ZHA can call with the last known latest firmware version) to make sure the image can be installed.

We could delay any global `get_ota_images` checks by 10 to 60 minutes, so HA doesn't immediately re-fetch the latest index upon startup and just reuses previous update information until we re-fetch it.
Either way, this is something that's mostly done in ZHA/HA and nothing that would impact this first PR.

Related backlog issue:
- https://github.com/zigpy/backlog/issues/15

### DIFF
--- a/tests/test_appdb.py
+++ b/tests/test_appdb.py
@@ -1607,3 +1607,209 @@ async def test_device_signature_ignores_quirks(tmp_path) -> None:
     assert dev2.original_signature == expected_signature
 
     await app2.shutdown()
+
+
+@patch("zigpy.device.Device.schedule_initialize", new=mock_dev_init(True))
+async def test_ota_query_cache_persistence(tmp_path):
+    """Test that OTA query cache is persisted and restored from the database."""
+    db = tmp_path / "test.db"
+    app = await make_app_with_db(db)
+    ieee = make_ieee()
+    app.handle_join(99, ieee, 0)
+
+    dev = app.get_device(ieee)
+    ep = dev.add_endpoint(1)
+    ep.status = zigpy.endpoint.Status.ZDO_INIT
+    ep.profile_id = 260
+    ep.device_type = profiles.zha.DeviceType.PUMP
+    ep.add_input_cluster(0)  # Basic cluster, exercises non-OTA skip in load
+    ota_cluster = ep.add_output_cluster(Ota.cluster_id)
+    app.device_initialized(dev)
+
+    # Simulate a query_next_image command from the device
+    cmd = Ota.QueryNextImageCommand(
+        field_control=Ota.QueryNextImageCommand.FieldControl.HardwareVersion,
+        manufacturer_code=0x1234,
+        image_type=0x5678,
+        current_file_version=0x000A0001,
+    )
+    cmd.hardware_version = 3
+
+    ota_cluster.last_query_cmd = cmd
+
+    # Trigger a save
+    app.device_initialized(dev)
+    await app.shutdown()
+
+    # Reload and check the OTA query cache was restored
+    app2 = await make_app_with_db(db)
+    dev2 = app2.get_device(ieee)
+    ota2 = dev2.endpoints[1].out_clusters[Ota.cluster_id]
+
+    assert ota2.last_query_cmd is not None
+    assert ota2.last_query_cmd.manufacturer_code == 0x1234
+    assert ota2.last_query_cmd.image_type == 0x5678
+    assert ota2.last_query_cmd.current_file_version == 0x000A0001
+    assert ota2.last_query_cmd.hardware_version == 3
+
+    # Also test get_last_ota_query_cmd
+    assert dev2.get_last_ota_query_cmd() is not None
+    assert dev2.get_last_ota_query_cmd().manufacturer_code == 0x1234
+
+    await app2.shutdown()
+
+
+@patch("zigpy.device.Device.schedule_initialize", new=mock_dev_init(True))
+async def test_ota_query_cache_no_hardware_version(tmp_path):
+    """Test OTA query cache round-trip without hardware_version."""
+    db = tmp_path / "test.db"
+    app = await make_app_with_db(db)
+    ieee = make_ieee()
+    app.handle_join(99, ieee, 0)
+
+    dev = app.get_device(ieee)
+    ep = dev.add_endpoint(1)
+    ep.status = zigpy.endpoint.Status.ZDO_INIT
+    ep.profile_id = 260
+    ep.device_type = profiles.zha.DeviceType.PUMP
+    ota_cluster = ep.add_output_cluster(Ota.cluster_id)
+    app.device_initialized(dev)
+
+    cmd = Ota.QueryNextImageCommand(
+        field_control=Ota.QueryNextImageCommand.FieldControl(0),
+        manufacturer_code=0xAAAA,
+        image_type=0xBBBB,
+        current_file_version=0x00000042,
+    )
+    ota_cluster.last_query_cmd = cmd
+    app.device_initialized(dev)
+    await app.shutdown()
+
+    app2 = await make_app_with_db(db)
+    dev2 = app2.get_device(ieee)
+    ota2 = dev2.endpoints[1].out_clusters[Ota.cluster_id]
+
+    assert ota2.last_query_cmd is not None
+    assert ota2.last_query_cmd.manufacturer_code == 0xAAAA
+    assert ota2.last_query_cmd.image_type == 0xBBBB
+    assert ota2.last_query_cmd.current_file_version == 0x00000042
+    assert not hasattr(ota2.last_query_cmd, "hardware_version") or (
+        ota2.last_query_cmd.hardware_version is None
+    )
+
+    await app2.shutdown()
+
+
+@patch("zigpy.device.Device.schedule_initialize", new=mock_dev_init(True))
+async def test_ota_query_cache_event_save(tmp_path):
+    """Test that the OtaQueryCacheUpdatedEvent triggers a DB save."""
+    db = tmp_path / "test.db"
+    app = await make_app_with_db(db)
+    ieee = make_ieee()
+    app.handle_join(99, ieee, 0)
+
+    dev = app.get_device(ieee)
+    ep = dev.add_endpoint(1)
+    ep.status = zigpy.endpoint.Status.ZDO_INIT
+    ep.profile_id = 260
+    ep.device_type = profiles.zha.DeviceType.PUMP
+    ota_cluster = ep.add_output_cluster(Ota.cluster_id)
+    app.device_initialized(dev)
+
+    # Simulate the event that _handle_query_next_image would emit
+    from zigpy.zcl import OtaQueryCacheUpdatedEvent
+
+    event = OtaQueryCacheUpdatedEvent(
+        device_ieee=str(ieee),
+        endpoint_id=1,
+        cluster_type=ota_cluster.cluster_type,
+        cluster_id=ota_cluster.cluster_id,
+        manufacturer_code=0x1111,
+        image_type=0x2222,
+        current_file_version=0x00000099,
+        hardware_version=7,
+    )
+
+    ota_cluster.emit(OtaQueryCacheUpdatedEvent.event_type, event)
+    await asyncio.sleep(0.2)
+
+    await app.shutdown()
+
+    app2 = await make_app_with_db(db)
+    dev2 = app2.get_device(ieee)
+    ota2 = dev2.endpoints[1].out_clusters[Ota.cluster_id]
+
+    assert ota2.last_query_cmd is not None
+    assert ota2.last_query_cmd.manufacturer_code == 0x1111
+    assert ota2.last_query_cmd.image_type == 0x2222
+    assert ota2.last_query_cmd.current_file_version == 0x00000099
+    assert ota2.last_query_cmd.hardware_version == 7
+
+    await app2.shutdown()
+
+
+async def test_ota_query_cache_load_missing_device(tmp_path):
+    """Test that loading OTA cache skips entries for devices no longer in the app."""
+    db = tmp_path / "test.db"
+    app = await make_app_with_db(db)
+    ieee = make_ieee()
+    app.handle_join(99, ieee, 0)
+
+    dev = app.get_device(ieee)
+    ep = dev.add_endpoint(1)
+    ep.status = zigpy.endpoint.Status.ZDO_INIT
+    ep.profile_id = 260
+    ep.device_type = profiles.zha.DeviceType.PUMP
+    ota_cluster = ep.add_output_cluster(Ota.cluster_id)
+    app.device_initialized(dev)
+
+    # Set a query cmd and save
+    ota_cluster.last_query_cmd = Ota.QueryNextImageCommand(
+        field_control=Ota.QueryNextImageCommand.FieldControl(0),
+        manufacturer_code=0x1234,
+        image_type=0x5678,
+        current_file_version=0x00000001,
+    )
+    await app._dblistener._save_device(dev)
+    await app.shutdown()
+
+    # Insert a row referencing a non-existent device directly in the DB
+    import sqlite3
+
+    with sqlite3.connect(db) as conn:
+        conn.execute(
+            f"INSERT INTO ota_query_cache{zigpy.appdb.DB_V}"
+            " (ieee, endpoint_id, manufacturer_code, image_type,"
+            "  current_file_version, hardware_version, last_updated)"
+            " VALUES (?, ?, ?, ?, ?, ?, ?)",
+            ("ff:ff:ff:ff:ff:ff:ff:ff", 1, 0xAAAA, 0xBBBB, 0x00000002, None, 0),
+        )
+
+    # Reload — should not crash, just skip the unknown device row
+    app2 = await make_app_with_db(db)
+    dev2 = app2.get_device(ieee)
+    ota2 = dev2.endpoints[1].out_clusters[Ota.cluster_id]
+    assert ota2.last_query_cmd is not None
+    assert ota2.last_query_cmd.manufacturer_code == 0x1234
+
+    await app2.shutdown()
+
+
+async def test_get_last_ota_query_cmd_returns_none(tmp_path):
+    """Test that get_last_ota_query_cmd returns None when no query has been cached."""
+    db = tmp_path / "test.db"
+    app = await make_app_with_db(db)
+    ieee = make_ieee()
+    app.handle_join(99, ieee, 0)
+
+    dev = app.get_device(ieee)
+    ep = dev.add_endpoint(1)
+    ep.status = zigpy.endpoint.Status.ZDO_INIT
+    ep.profile_id = 260
+    ep.device_type = profiles.zha.DeviceType.PUMP
+    ep.add_output_cluster(Ota.cluster_id)
+    app.device_initialized(dev)
+
+    assert dev.get_last_ota_query_cmd() is None
+
+    await app.shutdown()
--- a/tests/test_appdb.py
+++ b/tests/test_appdb.py
@@ -1747,6 +1747,59 @@ async def test_ota_query_cache_event_save(tmp_path):
     await app2.shutdown()
 
 
+@patch("zigpy.device.Device.schedule_initialize", new=mock_dev_init(True))
+async def test_ota_query_cache_restore_correct_cluster_type(tmp_path):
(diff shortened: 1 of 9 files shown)


zigpy/appdb.py
```diff
@@ -1145,14 +1250,19 @@ async def _migrate_tables(
...
-                async for row in cursor:
-                    placeholders = ",".join("?" * len(row))
+            # Use explicit column names to skip generated columns automatically
+            async with self.execute(f"PRAGMA table_info({old_table})") as cursor:
+                columns = [row[1] async for row in cursor]
+
+            col_list = ", ".join(columns)
+            placeholders = ", ".join("?" * len(columns))
+            select_sql = f"SELECT {col_list} FROM {old_table}"
+            insert_sql = f"INSERT INTO {new_table} ({col_list}) VALUES ({placeholders})"
```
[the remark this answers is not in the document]
TheJulianJES: This was basically copied from https://github.com/zigpy/zigpy/pull/1766/.

So, the `hidden` column is only available in `table_xinfo`, not `table_info`.
It requires a newer SQLite version but it would still be available in Ubuntu 20.04 Focal.
But unlike what Copilot suggests, generated columns shouldn't be included in `table_info` in the first place.

With `table_xinfo`, where they are included, we could explicitly skip generated columns:
```python
# Use explicit column names and skip generated columns automatically.
# `table_xinfo` exposes generated columns with hidden >= 2.
async with self.execute(f"PRAGMA table_xinfo({old_table})") as cursor:
    columns = [
        row[1] async for row in cursor if len(row) <= 6 or row[6] < 2
    ]
```

But I'm pretty sure this already works fine as-is..

TheJulianJES (commented):
For the comments in Copilot's "Comments suppressed due to low confidence":
Right now, the `last_query_cmd` is restored to both server and client `OTA` clusters if a device has both.

I'm not sure if there are devices that have both. How does ZHA even handle that? Two update entities? But we could either include that extra info in the DB or prefer to restore to client clusters, then only restore to the server cluster if the client one doesn't exist.

I'm not sure if it's really an issue in the first place.

---

EDIT: There are a few devices which have both a client OTA and server OTA cluster. It seems like something we should quirk away maybe?
I think the behavior we have in this PR where `cluster.last_query_cmd` is restored to both the client and server clusters on restart is *fine*. Though it is a small difference between runtime and after the clusters are restored from the DB. ZHA will only create an update entity for the client cluster anyway if both exist.

We could also just restore it to the client clusters if found, and skip the server one then. Thoughts?

EDIT 2: Actually, it might be an issue if a device sends a new request after a restart and `get_last_ota_query_cmd` is used. It iterates through all clusters and return the first "last query cmd". So maybe we'll just persist `cluster_type` into the DB as well to be extra safe.

TheJulianJES:
Ok, apparently there are devices with OTA client + server clusters. From our diagnostics, at least:
  - `tests/data/devices/lumi-lumi-ctrl-neutral2.json` (endpoint 1)
  - `tests/data/devices/lumi-lumi-remote-cagl02.json` (endpoint 1)
  - `tests/data/devices/lumi-lumi-sensor-86sw2.json` (endpoint 1)
  - `tests/data/devices/lumi-lumi-switch-b1lacn02.json` (endpoint 1)
  - `tests/data/devices/lumi-lumi-vibration-aq1.json` (endpoint 1)
  - `tests/data/devices/somfy-sonesse-28-wf-li-ion-roller.json` (endpoint 232)
  - `tests/data/devices/uiot-uiot-windowcovring2-0402.json` (endpoint 1)

They only have one update entity though (based on `OtaClientClusterHandler`). Is this because there's a `unique_id` conflict?

TheJulianJES:
Ok, there is a hidden unique ID conflict in ZHA... We should probably use `feature_priority` to explicitly prefer the client OTA entity? Also, is there no logging for unique ID conflicts? 

EDIT: No, there wasn't logging. Related PRs for the above comments:
- https://github.com/zigpy/zha/pull/683
- https://github.com/zigpy/zha/pull/684

tests/test_appdb.py
```diff
@@ -2,6 +2,7 @@
 import contextlib
 from datetime import UTC, datetime, timedelta
 import pathlib
+import sqlite3
```
TheJulianJES: `test_appdb_migration.py` still does: `from zigpy.appdb import sqlite3`. That shouldn't be needed anymore after:
- https://github.com/zigpy/zigpy/pull/1772/

We can update the other test file in this PR or another PR.

TheJulianJES:
There seems to be a flaky test with a `0.05`s periodic topology scan doing one more scan than expected: https://github.com/zigpy/zigpy/actions/runs/23218647589/job/67485823909?pr=1779

zigpy/appdb_schemas/schema_v15.sql
```diff
@@ -0,0 +1,219 @@
...
+    backup_json TEXT NOT NULL
+);
+
+
+-- OTA query cache: stores fields from the last query_next_image command per device
+DROP TABLE IF EXISTS ota_query_cache_v15;
+CREATE TABLE ota_query_cache_v15 (
+    ieee ieee NOT NULL,
+    endpoint_id INTEGER NOT NULL,
+    cluster_type INTEGER NOT NULL,
```
TheJulianJES: This is regarding some comments above:
I added this to make sure we only restore the "last query cmd" to the server/client cluster type we got it from when restarting (for buggy devices which have it as server + client). Other options include
- to only restore it to client clusters (doesn't work for all devices),
- restore it to both types (this would cause stale commands at runtime if a query is sent),
- or prefer restoring to client OTA clusters if both server + client are present (could work).
  - There could still be a small inconsistency if we add an `OTA` client cluster at runtime, instead of re-loading the device from the DB. It might be fine to ignore...

`_handle_query_next_image`, which saves the query, is only fired when `hdr.direction == Client_to_Server`. However, if only a server OTA cluster is present, zigpy will swap the direction bit!
So you can get into a weird state doing the following:
1. Device has only one OTA cluster (server), receives a query with flipped direction, `last_query_cmd` gets set there.
2. Firmware update adds a client OTA cluster.
3. New query arrives correctly on the client cluster, updating only that one. Server cache (runtime + DB) will be stale.

This is a rare scenario but it's not completely trivial to handle correctly...

TheJulianJES: These are all edge cases... I've reverted storing `cluster_type` in the DB now and we always prefer client `OTA` clusters on the same endpoint when restoring from the DB, ignoring a server one if present as well. We only restore to a server `OTA` cluster if there's no client one.

This fixes the issue where a device with a server `OTA` cluster also gets a client `OTA` firmware update cluster, which zigpy would prefer for storing the "last query cmd". Do note this is only fixed when the `Device` is reloaded from the DB (i.e. zigpy restarted), not if a `OTA` client cluster is magically added at runtime. This can't happen at the moment, so I think we can ignore this, especially considering how unlikely this is to happen in general...

tests/test_appdb_migration.py
```diff
@@ -476,14 +481,36 @@ async def test_migration_missing_tables(app):
...
+    )
+    await app._dblistener.execute("INSERT INTO src_v1 VALUES (1, 'hello')")
+    await app._dblistener.execute("INSERT INTO dst_v2 VALUES (1, 'conflict')")
+
+    app._dblistener._get_table_versions = AsyncMock(return_value={"src_v1": "1"})
+
+    with pytest.raises(sqlite3.IntegrityError):
+        await app._dblistener._migrate_tables({"src_v1": "dst_v2"})
+
+    await app.shutdown()
```
TheJulianJES, line 513: Note: This test was apparently added to properly cover the the line below in `_migrate_tables`.
It's not actually new to this PR though, so I can remove it again if wanted or just quickly pull it out into another PR.
https://github.com/zigpy/zigpy/blob/5cb542998b6f3b2a0ac8f6dd7d3afa42ea57d7ae/zigpy/appdb.py#L1157-L1158

tests/test_appdb.py
```diff
@@ -1607,3 +1607,187 @@ async def test_device_signature_ignores_quirks(tmp_path) -> None:
...
+    ep2.profile_id = 260
+    ep2.device_type = profiles.zha.DeviceType.PUMP
+    ota_cluster = ep2.add_output_cluster(Ota.cluster_id)
+
+    ota_cluster.last_query_cmd = Ota.QueryNextImageCommand(
+        field_control=Ota.QueryNextImageCommand.FieldControl(0),
+        manufacturer_code=0x1234,
+        image_type=0x5678,
+        current_file_version=0x00000001,
+    )
```
TheJulianJES: Note for me: Should we have this be sent from a device? (instead of only assigning it)

tests/test_appdb.py
```diff
@@ -1607,3 +1607,187 @@ async def test_device_signature_ignores_quirks(tmp_path) -> None:
...
+
+    # With hardware_version
+    cmd = Ota.QueryNextImageCommand(
+        field_control=Ota.QueryNextImageCommand.FieldControl.HardwareVersion,
+        manufacturer_code=0x1234,
+        image_type=0x5678,
+        current_file_version=0x000A0001,
+    )
+    cmd.hardware_version = 3
+    ota_cluster.last_query_cmd = cmd
```
TheJulianJES: Note for me: Same as other comment. Do we actually have a test that makes sure `last_query_cmd` is assigned correctly on OTA requests?

zigpy/device.py
```diff
@@ -936,6 +937,17 @@ async def update_firmware(
...
+        """Return the last cached QueryNextImageCommand, preferring client clusters."""
+        for ep in self.non_zdo_endpoints:
+            # Prefer client (out) clusters since runtime routing places
+            # query_next_image there when both cluster types exist.
+            for clusters in (ep.out_clusters, ep.in_clusters):
+                cluster = clusters.get(Ota.cluster_id)
+                if isinstance(cluster, Ota) and cluster.last_query_cmd is not None:
+                    return cluster.last_query_cmd
+        return None
+
```
TheJulianJES, line 957: Note: This isn't actually used/needed anywhere now, but it should be useful.
Otherwise, we can also remove it if we only want to include it when it's actually needed.

TheJulianJES (commented):
Some notes I still had pending:

zigpy/device.py
```diff
@@ -928,6 +928,7 @@ async def update_firmware(
             cluster_id=Ota.cluster_id, cluster_type=ClusterType.Client
         )
         ota.update_attribute(Ota.AttributeDefs.current_file_version.id, None)
+        ota.last_query_cmd = None
```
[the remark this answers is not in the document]
TheJulianJES: Yeah.. This is the same issue as below. It shouldn't matter at all because devices send another query eventually. (We should also image notify after an upgrade and during pairing, but that's for another PR.)

But we also don't want to leave old data in the DB when HA is restarted before a new query was sent. It's also possible for a firmware upgrade to completely remove OTA handling, so we should likely add a "cleared event" here to trigger the DB save.

zigpy/appdb.py
```diff
@@ -526,6 +532,36 @@ async def _save_unsupported_attributes(self, ep: Endpoint) -> None:
...
+                         current_file_version, hardware_version, last_updated)
+                    VALUES (?, ?, ?, ?, ?, ?, ?)
+                    ON CONFLICT (ieee, endpoint_id) DO UPDATE SET
+                        manufacturer_code=excluded.manufacturer_code,
+                        image_type=excluded.image_type,
+                        current_file_version=excluded.current_file_version,
+                        hardware_version=excluded.hardware_version,
+                        last_updated=excluded.last_updated"""
+            await self._db.executemany(q, rows)
+
```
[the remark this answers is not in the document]
TheJulianJES: See comment above.

zigpy/appdb_schemas/schema_v15.sql
```diff
@@ -0,0 +1,218 @@
...
+    ieee ieee NOT NULL,
+    endpoint_id INTEGER NOT NULL,
+    manufacturer_code INTEGER NOT NULL,
+    image_type INTEGER NOT NULL,
+    current_file_version INTEGER NOT NULL,
+    hardware_version INTEGER,
+    last_updated REAL NOT NULL,
+
+    FOREIGN KEY(ieee)
+        REFERENCES devices_v15(ieee)
```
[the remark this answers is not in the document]
TheJulianJES: This is valid. I think the initial approach included saving one query cmd per device, but technically, I guess there can be multiple `OTA` clusters with different query commands, or non-functional ones to which we don't want to restore the last query cmd to, as that would be different to runtime behavior.

Fixed with https://github.com/zigpy/zigpy/pull/1779/commits/045b0d9cee6ccb8e0962260c1ffd5524359e3be1. This does bring up a point – do we (or should we) have test coverage for this in general?

zigpy/device.py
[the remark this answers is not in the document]
TheJulianJES: Added a cleared event now: https://github.com/zigpy/zigpy/pull/1779/commits/c8dd50b4e47c69f23f6f859b9ec10dd654d55b3a + https://github.com/zigpy/zigpy/pull/1779/changes/232239346048e8476405c55770cb7cfc99c16cc0.

zigpy/appdb.py
```diff
@@ -1145,14 +1272,19 @@ async def _migrate_tables(
...
+            select_sql = f"SELECT {col_list} FROM {old_table}"
+            insert_sql = f"INSERT INTO {new_table} ({col_list}) VALUES ({placeholders})"
+
+            async with self.execute(select_sql) as cursor:
+                async for row in cursor:
                     try:
-                        await self.execute(
-                            f"INSERT INTO {new_table} VALUES ({placeholders})", row
-                        )
+                        await self.execute(insert_sql, row)
```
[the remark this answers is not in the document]
TheJulianJES: The same comment I have further up applies: https://github.com/zigpy/zigpy/pull/1779#discussion_r2862839380

`table_info` does indeed NOT include generated columns. I'm not sure if this is completely reliable behavior for all (future) versions though? It doesn't seem particularly documented anywhere.

The migration is at least somewhat covered by the tests using `zigbee_puddly2.db` (v13 -> v14, now also: v14 -> v15). E.g. if I replace `table_info` with `table_xinfo` and leave out the `hidden` check, most tests in `test_appdb_migration.py` completely fail because of the issue Copilot mentioned. But with `table_info`, it's not an issue.

TheJulianJES:
I think this should be ready for a human look now. A couple of comments above go into some of the design decisions. The PR description should still be a good summary as well.

If this is good to go, there's going to be one or two more zigpy PRs that send an image notify after an OTA update is successfully finished – this PR only clears it before we wait for the device to restart. And that PR (or yet another one) could also send the image notify command directly during pairing (or after, also fine). That way, we have the OTA information from the first minute. We should also include this in ZHA diagnostics.

---

In the future, we may also be able to drop the read for `current_file_version` during pairing and have that attribute be updated by the OTA query instead. For now, I'd just keep the old logic for quirk matching based on the `current_file_version` attribute and OTA matching properly done on the OTA query command.
The `current_file_version` attribute luckily seems to be accurate on all devices I tested with, so it doesn't need to be urgently changed. (though `image_type` and `manufacturer_code` attributes are inaccurate, hence the reason for this PR to cache the command instead + we also need hw version).

And another PR (likely in HA Core's `FirmwareUpdateEntity.on_add` (or ZHA?)) would call `get_ota_images` + `self.device_ota_image_query_result` as mentioned in the PR description. This would finally(!) fix update entities to always show correctly (assuming internet is available on startup).
In even more PRs, we should also have ZHA query `get_ota_images` every 24 hours or so (per device, but index will be cached by zigpy already). And we should stop forwarding every OTA query to HA update entities. There's no need to then. Like, we'd drop fetching OTA images for every "image query command" in HA then.
You'll see in subsequent PRs. 😄 

But for an initial fix, I think mainly this and the change in HA Core to just call `get_ota_images` on startup would be fine. I still want to properly do the other stuff I mentioned, but it could wait for another release. We can likely also drop OTA `listener_event`s in favor of the new zigpy events then.

zigpy/device.py
TheJulianJES: Nevermind. This is needed for the initial ZHA implementation that would call `get_ota_images` when adding the entity, as mentioned in the PR description.

And as said in another comment, I want to make some other changes here soon, keeping in mind a WS migration.

TheJulianJES (commented):
Tested on my production database: Migration and new OTA functionality both seem to work fine

TheJulianJES:
To add some more reference on why we need to cache the command (details) and can't read the attribute(s):
Screenshot from ZCL showing that `image_type` is supposed to return a value of 0xFFFF when not downloading an image, meaning we can't use it.

PR #1780, opened 2026-02-27
Suppress expected warnings in tests

TheJulianJES:
This filters expected warnings triggered by various tests.
There are two approaches (commits) though the current one (filtering by test) makes more sense IMO.


puddly (approved):
LGTM. I wasn't aware of this fixture!

TheJulianJES:
Neither was I 😄 We can make good use of it for ZHA as well.

PR #1781, opened 2026-03-01
Increase OTA block size cap

TheJulianJES:
## Proposed change

This adjusts the default capped OTA block size from 40 to 50 to prevent an issue with Telink based devices requesting 48 bytes but only getting 40 bytes and then failing the upgrade.

This aligns behavior with Z2M [here](https://github.com/Koenkk/zigbee-herdsman/blob/8d96a5c4f4601f4753315a7db35627620f99a196/src/controller/helpers/ota.ts#L310-L321) and the default cap of 50 being set [here](https://github.com/Koenkk/zigbee-herdsman/blob/8d96a5c4f4601f4753315a7db35627620f99a196/src/controller/helpers/ota.ts#L432), so this should work reliably.

## Additional information

Noticed for a (Telink) Sonoff smart plug OTA: https://github.com/zigpy/zigpy-ota/issues/25#issuecomment-3979034782

Also note that Z2M has additional timeout settings depending on the manufacturer: [zigbee-herdsman/src/controller/helpers/ota.ts#L399-L424](https://github.com/Koenkk/zigbee-herdsman/blob/8d96a5c4f4601f4753315a7db35627620f99a196/src/controller/helpers/ota.ts#L399-L424).
These are not implemented here but especially the Sonoff one may be interesting to look at.

Also, newer SiLabs based Hue bulbs request 64 bytes blocks. Previously, they would have gotten 40 bytes, now they get 50 bytes at least.


puddly:
I think we originally used 40 bytes because older coordinators (I believe the HUSBZB-1) crashed with higher values.

Maybe we can have the radio library also provide the maximum supported transfer size? If it supports 64 bytes, we can use 64 bytes.

TheJulianJES:
I wasn't sure what the limit was for exactly. It was only introduced 2 years ago with https://github.com/zigpy/zigpy/pull/1340 and looked to be for some incompatible devices.

If it's mostly for coordinators (HUSBZB-1), would you just limit it based on the EZSP version?
Do you happen to have a HUSBZB-1 to test with and figure out its exact limit? (40 vs 50 vs 64, ...)

puddly (approved):
Tested with an ancient HUSBZB-1 and it supports block sizes up to _63_ bytes (with very sporadic `MESSAGE_TOO_LONG` errors). I think 40 -> 50 bytes is perfectly fine.

PR #1783, opened 2026-03-11
Match devices with missing models/manufacturers with quirks v2

puddly:
A few devices do not have a valid model or manufacturer name but we should still match them with quirks v2.

Should help with https://github.com/zigpy/zha/issues/705.

### DIFF
--- a/zigpy/quirks/v2/__init__.py
+++ b/zigpy/quirks/v2/__init__.py
@@ -682,7 +682,7 @@ class QuirkBuilder:
         """Register this quirks v2 entry for the specified manufacturer and model."""
         if manufacturer is None and model is None:
             raise ValueError(
-                "At least one manufacturer and model must be specified for a v2 quirk."
+                "A manufacturer and/or model must be specified for a v2 quirk."
             )
 
         self.manufacturer_model_metadata.append(
--- a/zigpy/quirks/v2/__init__.py
+++ b/zigpy/quirks/v2/__init__.py
@@ -11,7 +11,7 @@ import inspect
 import logging
 import pathlib
 from types import FrameType
-from typing import TYPE_CHECKING, Any, Self
+from typing import TYPE_CHECKING, Any, Self, overload
 
 import attrs
 from frozendict import frozendict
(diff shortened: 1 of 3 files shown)


TheJulianJES (commented):
Looks good. Both Copilot comments may be something to look at.

Also, should we prevent quirks using `(None, None)` for manufacturer/model if they're not also using a `filter`? (Or do we trust users enough to not make a custom quirk that accidentally applies to all devices, overriding built-in ones?)

puddly:
The quirks won't apply as wildcards, they're just for devices lacking model or manufacturer names, so omitting one won't impact matching unless a device is really missing it. I guess there is a possibility someone makes a custom quirk that matches devices lacking both?

TheJulianJES:
Ah, yeah. Wildcard matching with a `.filter()` or something like a dedicated `.signature_filter()` might be something to eventually look into for remaining devices we can't convert now (e.g. `TS011F`).

I think we don't really need to care about devices with no manufacturer and model then, though a guard for not allowing a `None`, `None` match would be fine to add IMO.

puddly:
I've added a check for this case and added typing overloads to help mypy catch this at edit time.

zigpy/quirks/v2/__init__.py
```diff
@@ -670,8 +669,22 @@ def _add_entity_metadata(self, entity_metadata: EntityMetadata) -> Self:
...
+    def applies_to(self, manufacturer: str, model: str | None) -> Self: ...
+
+    @overload
+    def applies_to(self, manufacturer: None, model: str) -> Self: ...
+
+    def applies_to(self, manufacturer: str | None, model: str | None) -> Self:
         """Register this quirks v2 entry for the specified manufacturer and model."""
+        if manufacturer is None and model is None:
+            raise ValueError(
+                "At least one manufacturer and model must be specified for a v2 quirk."
```
TheJulianJES: nit: This reads a bit like you have to provide one of each.

zigpy/quirks/v2/__init__.py
```diff
@@ -604,16 +605,11 @@ class QuirkBuilder:
 
     def __init__(
         self,
-        manufacturer: str | None = None,
-        model: str | None = None,
+        manufacturer: str | None | UndefinedType = UNDEFINED,
+        model: str | None | UndefinedType = UNDEFINED,
         registry: DeviceRegistry | None = None,
     ) -> None:
```
TheJulianJES, line 611: nit: The [Copilot comment](https://github.com/zigpy/zigpy/pull/1783#discussion_r2918589752) about `QuirkBuilder("IKEA")` registering `("IKEA", None)` does seem somewhat valid, though it's a very minor complaint. It's probably fine.

TheJulianJES (approved):
Nice!

zigpy/quirks/v2/__init__.py
puddly: I'm not a huge fan of `QuirkBuilder()` taking any args or kwargs for this reason, to be honest. I think it would be cleaner to deprecate and move to explicit `.applies_to()`?

TheJulianJES: Yeah, maybe. It does save one line, though I also think it's cleaner to have an explicit `applies_to`.
We can migrate the existing v2 quirks in the repo next (and adjust AI instructions).

TheJulianJES (approved):

PR #1784, opened 2026-03-12
Allow quirk loading to fail

puddly:
https://github.com/home-assistant/core/issues/165380 exposed a bug with quirk loading: if a quirk fails to load, it takes down zigpy and the rest of ZHA in the process. This should not happen.

Independently, should `.removes()` hard crash when the cluster isn't present? Or should it silently succeed?

### DIFF
--- a/zigpy/quirks/registry.py
+++ b/zigpy/quirks/registry.py
@@ -140,7 +140,7 @@ class DeviceRegistry:
             try:
                 return candidate(device._application, device.ieee, device.nwk, device)
             except Exception:
-                _LOGGER.exception("Error creating quirk for %r: %r", device, candidate)
+                _LOGGER.exception("Failed to load quirk for %r: %r", device, candidate)
                 return device
 
         # If none match, return the original device
--- a/zigpy/quirks/registry.py
+++ b/zigpy/quirks/registry.py
@@ -118,7 +118,13 @@ class DeviceRegistry:
         if key in self._registry_v2:
             for entry in self._registry_v2[key]:
                 if entry.matches_device(device):
-                    return entry.create_device(device)
+                    try:
+                        return entry.create_device(device)
+                    except Exception:
+                        _LOGGER.exception(
+                            "Failed to load quirk for %r: %r", device, entry
+                        )
+                        return device
 
         # Then, fall back to v1 quirks
         for candidate in itertools.chain(
--- a/zigpy/quirks/registry.py
+++ b/zigpy/quirks/registry.py
@@ -136,7 +136,12 @@ class DeviceRegistry:
             _LOGGER.debug(
                 "Found custom device replacement for %s: %s", device.ieee, candidate
             )
-            return candidate(device._application, device.ieee, device.nwk, device)
+
+            try:
+                return candidate(device._application, device.ieee, device.nwk, device)
+            except Exception:
+                _LOGGER.exception("Error creating quirk for %r: %r", device, candidate)
+                return device
 
         # If none match, return the original device
         return device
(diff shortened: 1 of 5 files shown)


zigpy/quirks/registry.py
```diff
@@ -136,7 +136,12 @@ def get_device(self, device: Device) -> CustomDevice | Device:
             _LOGGER.debug(
                 "Found custom device replacement for %s: %s", device.ieee, candidate
             )
-            return candidate(device._application, device.ieee, device.nwk, device)
+
+            try:
+                return candidate(device._application, device.ieee, device.nwk, device)
+            except Exception:
+                _LOGGER.exception("Error creating quirk for %r: %r", device, candidate)
+                return device
```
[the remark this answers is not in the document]
puddly: To maintain existing loading sematics, we probably shouldn't do this. We assume it's one quirk per device.

TheJulianJES:
This has happened a couple of times in the past with v2 quirks expecting certain endpoints that didn't exist on some device variants. The nice thing is that these issues are always reported if ZHA fails to load. 😄 

Though I do agree it's probably the right decision to implement this.

> Independently, should `.removes()` hard crash when the cluster isn't present? Or should it silently succeed?

It's `.replaces()` that later causes the processing of `RemovesMetadata` from `ReplacesMetadata`. If we let that fail silently, `AddsMetadata` processing would fail [in the same way](https://github.com/zigpy/zigpy/blob/43e2e13daa8ca14d98a562753933412acc25d799/zigpy/quirks/v2/__init__.py#L176), so I'd argue we should still hard crash loading that quirk – ideally with a message that this should be reported. Otherwise, what is the `.adds()` call going to do with the missing endpoint?

zigpy/quirks/registry.py
```diff
@@ -118,7 +118,19 @@ def get_device(self, device: Device) -> CustomDevice | Device:
...
+                    try:
+                        return entry.create_device(device)
+                    except Exception:  # noqa: BLE001
+                        _LOGGER.exception(
+                            (
+                                "Failed to load quirk for %r: %r\n"
+                                "This is a bug. Please report it here: %s"
+                            ),
+                            device,
+                            entry,
```
TheJulianJES: This will log quite a bit. Do we want that? There is some useful info in there, so I guess it's fine..?

``
tests/test_quirks_v2.py::test_quirk_v2_loading_failure ERROR    zigpy.quirks.registry:registry.py:124 Failed to load quirk for <Device model='model' manuf='manufacturer' nwk=0x2233 ieee=sentinel.ieee is_initialized=False>: QuirksV2RegistryEntry(quirk_file=PosixPath('/Users/julian/PythonProjects/zigpy/tests/test_quirks_v2.py'), quirk_file_line=1638, manufacturer_model_metadata=(ManufacturerModelMetadata(manufacturer='manufacturer', model='model'),), friendly_name=None, exposes_features=(), device_alerts=(), disabled_default_entities=(), changed_entity_metadata=(), filters=(), fw_version_filter=None, custom_device_class=<class 'tests.test_quirks_v2.test_quirk_v2_loading_failure.<locals>.BadCustomDevice'>, device_node_descriptor=None, skip_device_configuration=False, adds_metadata=(), removes_metadata=(), replaces_metadata=(), replaces_cluster_occurrences_metadata=(), adds_endpoint_metadata=(), removes_endpoint_metadata=(), replaces_endpoint_metadata=(), entity_metadata=(), device_automation_triggers_metadata=frozendict.frozendict({}))
This is a bug. Please report it here: https://github.com/zigpy/zha-device-handlers
``

puddly: Maybe we can drop the metadata for quirks v2 entries?

TheJulianJES: The `quirk_file` of the `QuirksV2RegistryEntry` might be nice to see in case it's a custom quirk that fails but we should be able to easily clarify that if someone creates an issue. `quirk_file` needs to be updated to just the module name ideally.

I think this is good for now.

TheJulianJES (approved):

PR #1785, opened 2026-03-14
Return `ZCLAttributeDef` and status for reporting configuration

TheJulianJES:
## Proposed change

This changes `configure_reporting_multiple` and `configure_reporting` to return `dict[ZCLAttributeDef, foundation.Status]`, instead of a list with `ConfigureReportingResponseRecord`, as those only contain the attribute ID and status, not the manufacturer code.

Without these changes, reporting results are unclear if you try to set up reporting for an attribute ID with a manufacturer code and without one in the same call.

## Additional information

Discovered whilst working on:
- https://github.com/zigpy/zha-device-handlers/pull/4595/

This will need ZHA changes:
- https://github.com/zigpy/zha/pull/710

### Future

A similar issue also exist for `write_attributes` results.

### DIFF
--- a/zigpy/zcl/__init__.py
+++ b/zigpy/zcl/__init__.py
@@ -1416,7 +1416,7 @@ class Cluster(util.ListenableMixin, util.CatchingTaskMixin, EventBase):
         min_interval: int,
         max_interval: int,
         reportable_change: int,
-    ) -> list[foundation.ConfigureReportingResponseRecord]:
+    ) -> dict[foundation.ZCLAttributeDef, foundation.Status]:
         """Configure attribute reporting for a single attribute."""
         attr_def = self.find_attribute(attribute)
         return await self.configure_reporting_multiple(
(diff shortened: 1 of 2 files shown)


zigpy/zcl/__init__.py
```diff
@@ -1431,7 +1431,7 @@ async def configure_reporting(
 
     async def configure_reporting_multiple(
         self, config: dict[foundation.ZCLAttributeDef, ReportingConfig]
-    ) -> list[foundation.ConfigureReportingResponseRecord]:
+    ) -> dict[foundation.ZCLAttributeDef, foundation.Status]:
         """Configure attribute reporting for multiple attributes in the same request."""
```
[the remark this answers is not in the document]
TheJulianJES: From what I can see, Zigbee for Domoticz doesn't use this. I don't think we should add a compatibility layer here.

zigpy/zcl/__init__.py
TheJulianJES, line 1430: I guess the single-attribute `configure_reporting` method could just return a `ConfigureReportingResponseRecord` with the `attrid` from `attr_def`, or just a `Status`? Or just keep as is? Thoughts?

puddly: I think it's fine to keep as-is. We should probably discourage use of `configure_reporting` anyways in favor of `configure_reporting_multiple`.

puddly (approved):
LGTM. We use this pattern already for reading attributes (you get back a dict keyed by the exact object you pass in) so I think it's good to transition to this pattern elsewhere to fully deprecate `attrid`.

Issue #1805, opened 2026-03-30
Time cluster: missing dst_start, dst_end, dst_shift attribute handlers


benmaximov:
The Time cluster server handles reads for time, time_status, time_zone, standard_time, and local_time, but returns UNSUPPORTED_ATTRIBUTE for dst_start (0x0003), dst_end (0x0004), and dst_shift (0x0005). This prevents Zigbee end devices from obtaining DST information from the coordinator.

PR #1786 fixed existing handlers and noted "A future PR may add additional attributes"

PR #1806, opened 2026-03-30
Add DST attribute handlers to Time cluster

benmaximov:
Add handle_read_attribute_dst_start, handle_read_attribute_dst_end, and handle_read_attribute_dst_shift methods to the Time cluster. These were previously missing, causing UNSUPPORTED_ATTRIBUTE responses when devices requested DST information from the coordinator.

The implementation scans the current year's timezone transitions to find DST start/end times, using binary search for second-level precision. For timezones without DST, returns 0xFFFFFFFF for start/end and 0 for shift.

Tests cover:
- Northern hemisphere winter (LA, PST)
- Northern hemisphere summer (LA, PDT)
- Eastern Europe summer (Sofia, EEST)
- Southern hemisphere winter (Sydney, AEST) - reversed DST schedule
- Timezone without DST (Tokyo, JST)

#1805

### DIFF
--- a/tests/test_zcl_clusters.py
+++ b/tests/test_zcl_clusters.py
@@ -176,38 +176,224 @@ def _make_patched_datetime(fake_now):
 
 
 @pytest.mark.parametrize(
-    ("fake_now", "expected_utc", "expected_tz", "expected_standard", "expected_local"),
+    (
+        "fake_now",
+        "expected_utc",
+        "expected_tz",
+        "expected_standard",
+        "expected_local",
+        "expected_dst_start",
+        "expected_dst_end",
+        "expected_dst_shift",
+    ),
     [
+        # ── America/Los_Angeles (UTC-8 standard / UTC-7 DST) ──
         pytest.param(
-            # January: PST (UTC-8), no DST
-            datetime(2000, 1, 2, 0, 0, 0, tzinfo=ZoneInfo("America/Los_Angeles")),
-            # UTC time: midnight Jan 2 PST = 08:00 Jan 2 UTC = 1 day + 8 hours
-            24 * 60 * 60 + 8 * 60 * 60,
-            # Standard timezone offset: UTC-8
-            -(8 * 60 * 60),
-            # Standard time: Time + TimeZone = midnight Jan 2 local (no DST)
-            24 * 60 * 60,
-            # Local time: midnight Jan 2 = 1 day from epoch
-            24 * 60 * 60,
-            id="winter_no_dst",
+            # January: PST (UTC-8), before DST
+            datetime(2025, 1, 15, 12, 0, 0, tzinfo=ZoneInfo("America/Los_Angeles")),
+            790286400,  # UTC: 2025-01-15 20:00
+            -28800,  # standard offset UTC-8
+            790257600,  # standard_time (no DST active)
+            790257600,  # local_time = standard_time (no DST)
+            794829600,  # DST start: 2025-03-09 10:00 UTC
+            815389200,  # DST end: 2025-11-02 09:00 UTC
+            3600,
+            id="la_jan_before_dst",
         ),
         pytest.param(
-            # July: PDT (UTC-7), DST active
-            datetime(2000, 7, 2, 0, 0, 0, tzinfo=ZoneInfo("America/Los_Angeles")),
-            # UTC time: midnight Jul 2 PDT = 07:00 Jul 2 UTC
-            183 * 24 * 60 * 60 + 7 * 60 * 60,
-            # Standard timezone offset: still UTC-8 (DST not included)
-            -(8 * 60 * 60),
-            # Standard time: Time + TimeZone = 11 PM Jul 1 standard (no DST adjustment)
-            183 * 24 * 60 * 60 - 1 * 60 * 60,
-            # Local time: midnight Jul 2 local
-            183 * 24 * 60 * 60,
-            id="summer_with_dst",
+            # April: PDT (UTC-7), during DST
+            datetime(2025, 4, 15, 12, 0, 0, tzinfo=ZoneInfo("America/Los_Angeles")),
+            798058800,  # UTC: 2025-04-15 19:00
+            -28800,  # standard offset UTC-8
+            798030000,  # standard_time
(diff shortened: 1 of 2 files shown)


tests/test_zcl_clusters.py
```diff
@@ -202,12 +219,63 @@ def now(cls, tzinfo=None):
...
+            # Standard timezone offset: UTC+10 (AEST)
+            10 * 60 * 60,
+            # Standard time: Time + TimeZone = midnight Jul 2 (no DST)
+            183 * 24 * 60 * 60,
+            # Local time: midnight Jul 2 local
+            183 * 24 * 60 * 60,
+            # DST start: Aug 26, 2000 16:00 UTC (entering summer)
+            238 * 24 * 60 * 60 + 16 * 60 * 60,
+            # DST end: Mar 25, 2000 16:00 UTC (entering winter)
+            84 * 24 * 60 * 60 + 16 * 60 * 60,
```
TheJulianJES: Shouldn't the "DST end" be in next year for Australia?

E.g. if a client wants to check if DST Is active and does something like `DstStart <= Time <= DstEnd`, it would return `False` in January or February 2000.

zigpy/zcl/clusters/general.py
```diff
@@ -1160,6 +1160,140 @@ def handle_read_attribute_local_time(self) -> t.LocalTime:
...
+        # Try TZ environment variable
+        tz_env = os.environ.get("TZ")
+        if tz_env:
+            try:
+                return zoneinfo.ZoneInfo(tz_env)
+            except KeyError:
+                pass
+
+        # Fallback: fixed-offset timezone (no DST support)
+        return datetime.now().astimezone().tzinfo
```
TheJulianJES: It's really not great to read "OS internals" like this.

zigpy/zcl/clusters/general.py
```diff
@@ -1160,6 +1160,140 @@ def handle_read_attribute_local_time(self) -> t.LocalTime:
...
+
+        Scans the year hourly and uses binary search to find exact transition
+        times to the second.
+
+        Returns a tuple of (dst_start_utc, dst_end_utc, dst_offset) where
+        dst_start_utc/dst_end_utc are timezone-aware UTC datetimes of the
+        transitions, or None if the timezone does not observe DST.
+        dst_offset is the DST offset applied during the DST period.
+        """
+        from datetime import datetime as _datetime
```
TheJulianJES: This is already imported as `datetime`, why import it here again as `_datetime`? If it's for tests, we should try to clean that up another way.

TheJulianJES (commented):
Thanks for the PR!
We recently improved time handling with these PRs (part of HA Core 2026.4.0b0):
- https://github.com/zigpy/zigpy/pull/1786
- https://github.com/zigpy/zigpy/pull/1793

Most devices can derive all necessary information from those attributes. What devices do you have that explicitly need the DST start/end/shift attributes?

Also, this code won't work for timezones where the daylight saving time crosses the new year, like Australia.

ICEMANno1:
Hi @TheJulianJES,

this is not my PR but may I add to your question:

> What devices do you have that explicitly need the DST start/end/shift attributes?

I did some trial and error with my Bosch Radiator Thermostat II [+M] and it only sets its internal time, if the the above mentioned attributes are returned as well.

Without those, the Set weekly command of the ZCL thermostat cluster does not work.

If you need and details, I can try to provide anything requested.

Kind regards

benmaximov:
Thanks for the review @TheJulianJES !

I've addressed all the feedback and force-pushed the updated code.
 
Southern hemisphere (Australia): Fixed. dst_end now always comes after dst_start — the code scans into the next/previous year as needed.

OS internals: Simplified `_get_local_tz()` to try `TZ` env var first (which HA sets), then `/etc/localtime` and `/etc/timezone` as guarded fallbacks. Falls back gracefully to a fixed-offset timezone (DST attributes return `0xFFFFFFFF`). The `/etc` reads are kept because `datetime.now().astimezone().tzinfo` only returns a fixed-offset timezone — it has no DST transition knowledge. Computing DST start/end requires an IANA timezone name to construct a `ZoneInfo`. The only alternative would be adding `tzlocal` as a dependency.

_datetime alias: Removed. Using the module-level datetime directly.

Tests: 11 parametrized cases covering northern hemisphere (LA, Sofia) and southern hemisphere (Sydney) — before, during, and after DST — plus a no-DST case (Tokyo).


Devices needing DST attributes: Our [Shelly H&T Display ZB](https://www.shelly.com/products/shelly-blu-h-t-display-zb) reads dst_start/dst_end/dst_shift from the coordinator to switch its clock at the right moment (it already does this via BLE, now adding Zigbee support). More broadly, any Zigbee device with an internal clock that displays time or runs schedules needs to know when DST transitions happen — standard_time/local_time only tell the current offset.

ICEMANno1:
@TheJulianJES would you mind looking at this again 😇 I would be very appreciative! 🙇

ICEMANno1:
@benmaximov are you planning on working on this? I don't want to do anything on someone else's PR. 😅

zigpy/zcl/clusters/general.py
```diff
@@ -1160,6 +1163,184 @@ def handle_read_attribute_local_time(self) -> t.LocalTime:
...
+        # Not in DST — return the next upcoming period
+        for start, end in periods:
+            if start > now:
+                return start, end, dst_offset
+
+        # All periods are in the past — return the last one
+        return periods[-1][0], periods[-1][1], dst_offset
+
+    @staticmethod
+    def _get_local_tz() -> Any:
```
puddly, line 1280: Home Assistant passes the current timezone as a `ZoneInfo` object to ZHA directly: https://github.com/zigpy/zha/blob/82a14cd8ac8d643c49f3558a6040885335e1522e/zha/application/helpers.py#L431. It also is updated by Home Assistant in realtime, so ZHA is always in sync.

I think this needs to be passed down to zigpy, we shouldn't have zigpy try to detect this from the environment/filesystem like this.

benmaximov:
> @benmaximov are you planning on working on this? I don't want to do anything on someone else's PR. 😅

I don't mind, if you can make it mergable - go ahead. Not enough time for me to work on this right now

PR #1812, opened 2026-04-12
Allow response matching on known custom-profile ZCL endpoints

### LINKED ISSUE #1810: Shelly Gen4 RPC replies on profile 0xC001 are dropped because zigpy only parses ZCL packets for ZHA/ZLL profiles
## Problem

Shelly Gen4 Zigbee devices expose their RPC transport on endpoint `239` / cluster `0xFC01` using profile `0xC001` and manufacturer code `0x1490`.

On a live Home Assistant `ZHA` setup, RPC requests can be written to the device, but replies appear to be dropped before they can be matched back to the caller.

## Environment

Observed on:

- Home Assistant `2026.4.1`
- `bellows` `0.49.0`
- `zigpy` `1.2.2`
- `zha` `1.1.1`
- coordinator path: EZSP over `socket://...`

Affected device in the live system:

- manufacturer: `Shelly`
- model: `Mini1PM`
- endpoint `239`
- endpoint `239` profile `0xC001`
- endpoint `239` input clusters: `0xFC01`, `0xFC02`

Relevant upstream references:

- Shelly RPC cluster docs: https://shelly-api-docs.shelly.cloud/gen2/Integrations/Zigbee/RPCCluster/
- Shelly input docs: https://shelly-api-docs.shelly.cloud/gen2/ComponentsAndServices/Input/
- Shelly device features: https://shelly-api-docs.shelly.cloud/gen2/Integrations/Zigbee/DeviceFeatures/
- Related ZHA discussion: https://github.com/zigpy/zha/discussions/470
- Related zha-device-handlers request: https://github.com/zigpy/zha-device-handlers/issues/4125

## What we verified

1. The device does **not** expose a standard Zigbee Binary Input / Multistate Input cluster for the physical input state.
2. The physical input is available via Shelly RPC (`Input.GetStatus`) according to Shelly docs.
3. A custom Home Assistant probe was able to send the RPC request by writing to the Shelly RPC cluster attributes:
   - attr `0` = `Data`
   - attr `1` = `TxCtl`
   - attr `2` = `RxCtl`
4. The request path did not immediately fail, but every attempt timed out waiting for the reply frame.
5. In `zigpy/device.py`, `_parse_packet_command()` only parses incoming ZCL packets when `packet.profile_id` is `zha.PROFILE_ID` or `zll.PROFILE_ID`; otherwise it returns `(None, None)`.

From the installed `zigpy 1.2.2` code:

```python
elif packet.profile_id in (zha.PROFILE_ID, zll.PROFILE_ID):
    hdr, _ = foundation.ZCLHeader.deserialize(data)
    ...
    return hdr, rsp_key
else:
    return None, None
```

## Why this looks like a zigpy bug / gap

Shelly's RPC replies for these devices are expected on endpoint `239` with profile `0xC001`. If zigpy refuses to parse incoming ZCL traffic for any non-ZHA/ZLL profile, the reply cannot be matched and higher layers only see a timeout.

That seems like the core reason Shelly's RPC transport is unusable through the current ZHA stack even when the cluster and manufacturer code are known.

## Expected behavior

`zigpy` should allow parsing and routing of valid ZCL traffic arriving on supported non-standard profiles when the endpoint and cluster are known, or provide an explicit mechanism for integrations / quirks to opt those profiles in.

## Actual behavior

Shelly RPC requests time out because the response path on profile `0xC001` appears to be dropped during packet parsing.

## Notes

I can provide the live device signature and the exact RPC probe code path if that would help, but the behavior above was enough to reproduce the timeout and identify the gating condition in `zigpy/device.py`.

dw86uk:
## Summary

This change lets `zigpy` parse and match ZCL replies on known non-ZHA/ZLL endpoints when the incoming packet profile matches the endpoint profile and the cluster is already known on that endpoint.

In practice, this unblocks devices like Shelly Gen4 units that expose their RPC transport on endpoint `239` using profile `0xC001` and ZCL cluster `0xFC01`.

Fixes #1810.

## Root cause

`Device._parse_packet_header()` only treated packets as ZCL when `packet.profile_id` was ZHA or ZLL. For valid ZCL traffic on known custom-profile endpoints, that caused the packet to be routed down the custom-profile ignore path instead of being parsed and matched against pending requests.

## What changed

- add a small helper to identify custom-profile packets that are still known ZCL traffic for the device
- allow `_parse_packet_header()` to deserialize those packets as ZCL
- add a regression test covering response matching on a known custom-profile endpoint

## Validation

- `python -m pytest -q tests/test_device.py -k "custom_profile_response_matched_on_known_endpoint_profile or handle_custom_profile"`
- `python -m pytest -q tests/test_device.py`

### DIFF
--- a/zigpy/device.py
+++ b/zigpy/device.py
@@ -700,6 +700,20 @@ class Device(zigpy.util.LocalLogMixin, zigpy.util.ListenableMixin):
             packet.profile_id,
         )
 
+    def _is_known_zcl_packet_profile(self, packet: t.ZigbeePacket) -> bool:
+        """Return whether a custom-profile packet can be handled as ZCL."""
+        if packet.src_ep not in self.endpoints:
+            return False
+
+        endpoint = self.endpoints[packet.src_ep]
+        if not isinstance(endpoint, zigpy.endpoint.Endpoint):
+            return False
+
+        return endpoint.profile_id == packet.profile_id and (
+            packet.cluster_id in endpoint.in_clusters
+            or packet.cluster_id in endpoint.out_clusters
+        )
+
     def _should_filter_packet(self, packet: t.ZigbeePacket) -> bool:
         """Check if packet should be filtered as duplicate."""
         return self._packet_debouncer.filter(
@@ -723,7 +737,9 @@ class Device(zigpy.util.LocalLogMixin, zigpy.util.ListenableMixin):
                 tsn=hdr.tsn,
             )
             return hdr, rsp_key
-        elif packet.profile_id in (zha.PROFILE_ID, zll.PROFILE_ID):
+        elif packet.profile_id in (zha.PROFILE_ID, zll.PROFILE_ID) or (
+            self._is_known_zcl_packet_profile(packet)
+        ):
             hdr, _ = foundation.ZCLHeader.deserialize(data)
             rsp_key = ResponseKey(
                 endpoint_id=packet.src_ep,
--- a/tests/test_device.py
+++ b/tests/test_device.py
@@ -475,6 +475,65 @@ async def test_handle_custom_profile(dev) -> None:
     assert mock_handler.mock_calls == [call(packet)]
 
 
+async def test_custom_profile_response_matched_on_known_endpoint_profile(dev) -> None:
+    """Test that ZCL responses can match on a known custom-profile endpoint."""
+    ep = dev.add_endpoint(239)
+    ep.profile_id = 0xC001
+    ep.add_input_cluster(OnOff.cluster_id)
+
+    tsn = 0x12
+    rsp_key = device.ResponseKey(
+        endpoint_id=239,
+        cluster_id=OnOff.cluster_id,
+        direction=foundation.Direction.Server_to_Client,
+        tsn=tsn,
+    )
+    future = asyncio.get_running_loop().create_future()
+    dev._requests[rsp_key] = future
+
+    default_rsp_hdr = foundation.ZCLHeader(
+        frame_control=foundation.FrameControl(
+            frame_type=foundation.FrameType.GLOBAL_COMMAND,
+            is_manufacturer_specific=False,
+            direction=foundation.Direction.Server_to_Client,
+            disable_default_response=True,
+            reserved=0,
+        ),
+        tsn=tsn,
+        command_id=foundation.GeneralCommand.Default_Response,
+    )
+
+    default_rsp_cmd = foundation.GENERAL_COMMANDS[
+        foundation.GeneralCommand.Default_Response
+    ].schema(
+        command_id=OnOff.ServerCommandDefs.on.id,
+        status=foundation.Status.SUCCESS,
+    )
+
+    with patch.object(dev, "custom_profile_packet_received") as custom_handler:
+        dev.packet_received(
+            t.ZigbeePacket(
+                src=t.AddrModeAddress(addr_mode=t.AddrMode.NWK, address=dev.nwk),
+                src_ep=239,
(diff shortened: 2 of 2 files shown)


zigpy/device.py
```diff
@@ -700,6 +700,20 @@ def custom_profile_packet_received(self, packet: t.ZigbeePacket) -> None:
             packet.profile_id,
         )
 
+    def _is_known_zcl_packet_profile(self, packet: t.ZigbeePacket) -> bool:
```
puddly, line 703: Won't this function always return `True`?

zigpy/device.py
puddly, line 696: I think it would be best for the Shelly device quirk to implement this function instead.

zigpy/device.py
dw86uk: Not for the existing ZHA/ZLL paths because those short-circuit before this helper is consulted, but I agree the current shape makes that too hard to see. I can tighten that up.

zigpy/device.py
dw86uk: That makes sense to me. Before I rework this, would you prefer a small `zigpy` hook that defaults to `False` and lets a quirk opt a custom-profile endpoint into normal ZCL response parsing, with the Shelly behavior in `zha-device-handlers`, or do you think this should stay entirely out of `zigpy` if there is a cleaner quirk-only route?

puddly: I think this should be kept in quirks. It's a little odd for ZCL to be used when the profile is explicitly _not_ ZCL. We had issues in the past with being lenient with parsing so I would prefer to keep zigpy unaware of any protocol extensions for now.

One way you can easily implement this is with a v2 quirk:

```python
class ShellyDevice(CustomDevice):
    def _parse_packet_header(
        self, packet: t.ZigbeePacket
    ) -> tuple[zdo_t.ZDOHeader | foundation.ZCLHeader, ResponseKey] | tuple[None, None]:
        """Parse packet header and create response key."""
        hdr, rsp_key = super()._parse_packet_header(packet)
        if hdr is not None and rsp_key is not None:
            return hdr, rsp_key

        data = packet.data.serialize()

        # Shelly devices use ZCL with a different profile ID
        if packet.profile_id == SHELLY_PROFILE_ID:
            hdr, _ = foundation.ZCLHeader.deserialize(data)
            rsp_key = ResponseKey(
                endpoint_id=packet.src_ep,
                cluster_id=packet.cluster_id,
                direction=hdr.frame_control.direction,
                tsn=hdr.tsn,
            )
            return hdr, rsp_key
        else:
            return None, None
```

Then, you would use `.device_class(ShellyDevice)` to override it.

---

Alternatively, we could modify `elif packet.profile_id in (zha.PROFILE_ID, zll.PROFILE_ID):` to reference something like `elif packet.profile_id in self._response_profile_ids:` that defaults to `(zha.PROFILE_ID, zll.PROFILE_ID)` within the device object itself.

The current request/response matching infrastructure assumes ZCL/ZDO, where responses are keyed by `ResponseKey` (allowing for `tsn` matching narrowed by endpoint and cluster ID).

dw86uk:
Closing this PR because the current implementation no longer needs a zigpy-core change.

The review feedback here was to keep Shelly custom-profile ZCL parsing in the quirk layer instead of making zigpy parse known non-ZHA/ZLL profiles globally. I have moved the working implementation to:

- https://github.com/zigpy/zha-device-handlers/pull/5106, which adds `ShellyCustomProfileDevice` and parses Shelly profile `0xc001` only for the Shelly quirk.
- https://github.com/zigpy/zha/pull/800, which binds the Shelly RPC cluster and configures `rx_ctl` reporting.

That path is now live-tested on HA 2026.6.3 with a Shelly Mini1PM on stable firmware `1.7.5`, and it avoids broadening zigpy packet parsing behavior.

PR #1815, opened 2026-04-20
Switch serial to serialx

puddly:
We initially added a runtime import "preference" for pyserial-asyncio-fast over pyserial-asyncio, later we did the same with serialx over pyserial and pyserial-asyncio. There's been no negative feedback from any ZHA users over the past few months so I think this experiment can be concluded.

This PR replaces the runtime imports with a normal dependency on serialx.

### DIFF
--- a/zigpy/serial.py
+++ b/zigpy/serial.py
@@ -74,9 +74,7 @@ async def create_serial_connection(
     flow_control: Literal["hardware", "software"] | None | UndefinedType = UNDEFINED,
     **kwargs: Any,
 ) -> tuple[asyncio.Transport, asyncio.Protocol]:
-    """Wrapper around pyserial-asyncio that transparently substitutes a normal TCP
-    transport and protocol when a `socket` connection URI is provided.
-    """
+    """Wrapper for serialx that provides simplified flow control kwargs."""
 
     if flow_control is not UNDEFINED:
         xonxoff = flow_control == "software"
--- a/zigpy/serial.py
+++ b/zigpy/serial.py
@@ -64,7 +64,7 @@ class SerialProtocol(asyncio.Protocol):
 
 
 async def create_serial_connection(
-    loop: asyncio.BaseEventLoop,
+    loop: asyncio.AbstractEventLoop,
     protocol_factory: Callable[[], asyncio.Protocol],
     url: pathlib.Path | str,
     *,
--- a/zigpy/serial.py
+++ b/zigpy/serial.py
@@ -1,26 +1,20 @@
 from __future__ import annotations
 
 import asyncio
-from asyncio import timeout as asyncio_timeout
+from collections.abc import Callable
+import errno
 import logging
 import pathlib
-import typing
-from typing import Literal
-import urllib.parse
+from typing import Any, Literal, cast
 
-try:
-    # serialx is API-compatible with pyserial
-    import serialx as pyserial
-    import serialx as pyserial_asyncio
-except ImportError:
-    import serial as pyserial
-    import serial_asyncio_fast as pyserial_asyncio
+from serialx import (
+    SerialTransport,
+    create_serial_connection as serialx_create_serial_connection,
+)
 
 from zigpy.typing import UNDEFINED, UndefinedType
 
 LOGGER = logging.getLogger(__name__)
-DEFAULT_SOCKET_PORT = 6638
-SOCKET_CONNECT_TIMEOUT = 5
 
 
 class SerialProtocol(asyncio.Protocol):
@@ -28,7 +22,7 @@ class SerialProtocol(asyncio.Protocol):
 
     def __init__(self) -> None:
         self._buffer = bytearray()
-        self._transport: pyserial_asyncio.SerialTransport | None = None
+        self._transport: SerialTransport | None = None
 
         self._connected_event = asyncio.Event()
         self._disconnected_event = asyncio.Event()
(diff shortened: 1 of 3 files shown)


zigpy/serial.py
```diff
@@ -71,15 +65,14 @@ async def disconnect(self) -> None:
...
+    protocol_factory: Callable[[], asyncio.Protocol],
     url: pathlib.Path | str,
     *,
     baudrate: int = 115200,  # We default to 115200 instead of 9600
-    exclusive: bool | None = True,
     xonxoff: bool | UndefinedType = UNDEFINED,
     rtscts: bool | UndefinedType = UNDEFINED,
     flow_control: Literal["hardware", "software"] | None | UndefinedType = UNDEFINED,
-    **kwargs: typing.Any,
+    **kwargs: Any,
```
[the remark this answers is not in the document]
puddly: This is the default in serialx.

zigpy/serial.py
```diff
@@ -104,38 +97,24 @@ async def create_serial_connection(
...
+            **kwargs,
+        )
+    except OSError as exc:
+        if exc.errno == errno.EBUSY:
             # Re-raise a more useful exception
             raise PermissionError(
                 "The serial port is locked by another application"
             ) from exc
 
+        raise
```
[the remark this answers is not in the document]
puddly: `BlockingIOError` was a pyserial internal. We remap this to OSError across platforms in serialx.

zigpy/serial.py
```diff
@@ -71,15 +65,14 @@ async def disconnect(self) -> None:
...
     *,
     baudrate: int = 115200,  # We default to 115200 instead of 9600
-    exclusive: bool | None = True,
     xonxoff: bool | UndefinedType = UNDEFINED,
     rtscts: bool | UndefinedType = UNDEFINED,
     flow_control: Literal["hardware", "software"] | None | UndefinedType = UNDEFINED,
-    **kwargs: typing.Any,
+    **kwargs: Any,
 ) -> tuple[asyncio.Transport, asyncio.Protocol]:
     """Wrapper around pyserial-asyncio that transparently substitutes a normal TCP
```
TheJulianJES: nit: stale docstring referencing `pyserial-asyncio`

zigpy/serial.py
```diff
@@ -104,38 +97,24 @@ async def create_serial_connection(
...
 
     url = str(url)
-    parsed_url = urllib.parse.urlparse(url)
-
-    if parsed_url.scheme in ("socket", "tcp"):
-        async with asyncio_timeout(SOCKET_CONNECT_TIMEOUT):
-            transport, protocol = await loop.create_connection(
-                protocol_factory=protocol_factory,
-                host=parsed_url.hostname,
-                port=parsed_url.port or DEFAULT_SOCKET_PORT,
```
TheJulianJES, line 114: Not sure if it's a big issue, but I don't think Serialx defaults to port `6638`? It always expects the port in the URL to be present?

puddly: Yeah, 6638 was used by I think one or two very early Tube devices long, long ago. If this causes issues we should probably perform a migration for this in Core to make the port explicit.

The only way a user would hit this is if they were using an old Tube device, skipped zeroconf discovery, and manually added the connection URI while omitting the port.

zigpy/serial.py
```diff
@@ -71,15 +65,14 @@ async def disconnect(self) -> None:
 
 async def create_serial_connection(
     loop: asyncio.BaseEventLoop,
```
TheJulianJES: pre-existing: Typing `loop` as `AbstractEventLoop`  might be slightly more accurate (same as Serialx), as that's the return typing for `asyncio.get_event_loop()` / `asyncio.get_running_loop()` (both of which are used by the radio libraries to call this wrapper).

zigpy/serial.py
```diff
@@ -104,38 +95,24 @@ async def create_serial_connection(
     )
 
     url = str(url)
-    parsed_url = urllib.parse.urlparse(url)
-
-    if parsed_url.scheme in ("socket", "tcp"):
-        async with asyncio_timeout(SOCKET_CONNECT_TIMEOUT):
```
TheJulianJES, line 110: Serialx doesn't have a timeout of its own for this, right? So this would just hang forever for invalid hosts? (or I guess until the OS times this out, which is 75 seconds for macOS and apparently about 2 minutes for Linux.)

puddly: Good point, we only have one for the sync class: https://github.com/puddly/serialx/blob/ad330c6e14866666c30dc3b9e86aff4c4dbcad6b/serialx/platforms/serial_socket.py#L235-L240

zigpy/serial.py
TheJulianJES: Yeah, ok. `6638` also seems to be the default port for SMLIGHT and ZigStar devices, though the documentation for manually adding the device I saw always has the port in the URI. Hopefully, it's not an issue then..

zigpy/serial.py
TheJulianJES: Added to Serialx with:
- https://github.com/puddly/serialx/pull/67

The 10 second timeout seems rather long for zigpy's use-case, but I think it should be fine.

TheJulianJES (approved):
Nice! Looks good. I hope the default port (`6638`) missing is not a big issue. The 10 second timeout also seems rather long for ZHA, but it should be fine.

(We can drop the [pyserial dependencies from ZHA](https://github.com/zigpy/zha/blob/8a0dac0767992dcdb55e025a139552b72be660e3/pyproject.toml#L24-L25) next then.)

zigpy/serial.py
tube0013: I have instructed customers to use :6638 since the beginning. I don't think I ever realized it would default to it before. So unless someone just didn't follow docs and had it luckily work I think it's fine.

PR #1824, opened 2026-05-26
Move application-level retries into zigpy

puddly:
Zigpy currently only performs retries for network-level issues: low-level delivery failures or enqueuing exceptions. It does not retry commands that fail a high-level flow, such as ZCL requests not receiving a ZCL response. ZHA currently performs three retries globally for all requests. To separate logic, this PR moves all retrying into zigpy itself and performs retries for both ZCL and ZDO requests within `device.py`.


zigpy/device.py
```diff
@@ -623,6 +623,8 @@ async def request(
         use_ieee=False,
         ask_for_ack: bool | None = None,
         priority: int | None = None,
+        retries: int = 5,
+        **kwargs,
```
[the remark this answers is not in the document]
puddly: Zigpy tried three times at the APS level and ZHA retried three times at the ZCL/ZDO level. This is hard to represent as a single number, since APS retries differ from ZCL/ZDO retries. We can stick to ZHA's default of 3 and see what happens.

zigpy/device.py
```diff
@@ -589,7 +586,6 @@ async def _discover(self) -> None:
 
         self.info("Discovered basic device information for %s", self)
 
-    @zigpy.util.retryable_request(tries=5, delay=0.5)
```
TheJulianJES, line 592: Hopefully just a note, but this does change behavior from restarting the entire discovery process when one failure happens to just resending the individual requests when a delivery failure happens.

For example, if we got a non-SUCCESS status when discovering endpoints, we previously restarted discovery, but now fail entirely, as we do not catch `InvalidResponse`, for example:
https://github.com/zigpy/zigpy/blob/13590e8f0dfc608f0b9ae373a8da6e2e2566c99a/zigpy/device.py#L503-L508

I don't know if there are any (Tuya or Aqara) devices that potentially have such flakiness – hopefully not. But I did want to point it out because the `Device._discover()` seems like it was intentionally written to be run multiple times potentially.

zigpy/device.py
```diff
@@ -334,9 +333,7 @@ async def reinterview(self) -> None:
                 async with self._application.request_priority(
                     t.PacketPriority.CRITICAL
                 ):
-                    await zigpy.util.retryable_request(tries=2, delay=0.5)(
-                        shadow._discover
-                    )()
```
TheJulianJES, line 339: I did chose a total of 2 tries here because the ZDO node descriptor is `None` for the fresh `Device` object at this point, so we default to the `APS_REPLY_TIMEOUT_EXTENDED` timeout of 28 seconds.

If we never got a response from the device at all when starting the reinterview, we had a worst case of 56 seconds. Now, that's 84 seconds (with 3 tries). Realistically, this should only happen for a fully sleeping end-device or a completely disconnected mains-powered device, but I intentionally did not want to go overboard here. This may still be fine.

Do note that we never even reach this step for post-OTA retrying, because we intentionally retry reading the `current_file_version` like 10 times with a long delay to make sure the device is awake. If it's not, we never even bother starting the re-interview if that attribute read fails completely, so we also never change routing from the active `Device` object to the new shadow one.

So, this can only really happen if the user manually starts a re-interview. I did like the 1 minute timeout more for the reconfiguration dialog (though we just fail with success on timeout..) and for the ~~entity unloading~~ (edit: I do think we may not even unload the entities at this stage, but just redirect routing to the new `Device`? Either way, doesn't really matter).

zigpy/device.py
```diff
@@ -1019,8 +1052,10 @@ async def update_firmware(
         )
 
         await asyncio.sleep(AFTER_OTA_ATTR_READ_DELAY)
-        await OTA_RETRY_DECORATOR(ota.read_attributes)(
-            [Ota.AttributeDefs.current_file_version.name]
+        await ota.read_attributes(
+            [Ota.AttributeDefs.current_file_version.name],
+            retries=10,
```
TheJulianJES, line 1057: nit: This is 11 tries now, instead of the 10 we had before. It's still fine – the more the better here.

(This is just to make sure the device is awake before we start a re-interview. If it never responds, we shouldn't even bother with what's below.)

zigpy/device.py
puddly: Yeah, the old behavior was very strange and I don't think was written in the pseudo-stateless way for any reason other than it being restarted a few time times at runtime.

puddly: Hmm, the layering comment is a good point. I think for the most part we assume that successfully-delivered responses aren't failures in the "retrying will fix it" sense. It would be a problem if `Active_EP_req`, like you mentioned, did have a different reply upon retry.

TheJulianJES: Yeah, I had this in my mind because some Aqara devices had some flakiness here, but it seems like those might just be timeout errors. In that case, we *should* be fine here.. There's another comment from the bot that landed in a new thread below: https://github.com/zigpy/zigpy/pull/1824#discussion_r3308252949

(I guess this brings us back to the entire non-SUCCESS topic..)

PR #1833, opened 2026-06-10
Cap aiohttp <3.14 in tests for aioresponses compatibility

TheJulianJES:
This caps `aiohttp <3.14` in the `testing` group. It's incompatible with `aioresponses` 3.14 and newer.

We can remove this once this PR is released:
- https://github.com/pnuckowski/aioresponses/pull/288

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,6 +36,9 @@ zigpy = ["py.typed"]
 [dependency-groups]
 testing = [
     "aioresponses",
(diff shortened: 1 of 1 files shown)


puddly (approved):

TheJulianJES:
Let's go with this for now. There's also another library we can look at that should be a drop-in-replacement:
- https://github.com/Polandia94/aiointercept
- from: https://github.com/home-assistant-libs/python-supervisor-client/pull/347

PR #1839, opened 2026-06-11
Skip registering duplicate OTA providers

TheJulianJES:
Follow-up promised in https://github.com/zigpy/zigpy/pull/1837#discussion_r3392829381 (raised by Copilot review).

## Problem

The same logical provider can be registered multiple times — e.g. listed in both `providers` and `extra_providers`, or once via a deprecated config option and once explicitly. Nothing deduplicates them: `_register_providers()` only filters by `override_previous` and disabled types.

Each duplicate instance keeps its own `_index_last_updated`, and the refresh calls in `get_ota_images()` are sequential awaits, so `combine_concurrent_calls` never merges them. The result is one redundant index download per duplicate instance per 24 h expiration cycle. (Correctness is unaffected: equal providers share a single image cache bucket.)

## Changes

`register_provider()` now ignores a provider that compares equal to an already-registered one, with a warning (a duplicate is always an actionable misconfiguration, registration happens once at startup, and the deprecated config options already warn). Equal providers are interchangeable — equality requires the exact type plus identical URL/manufacturer IDs since #1838 — so dropping the duplicate is safe for both explicit registrations and merged config intents.

## Testing

- Registering an equal provider twice keeps a single instance; a provider of a different type with identical fields is still registered.
- The existing `test_ota_matching_priority` exercises the dedup implicitly: it registers two equal `BrokenProvider` instances, and its assertions are unaffected by the second one now being skipped.

### DIFF
--- a/zigpy/ota/__init__.py
+++ b/zigpy/ota/__init__.py
@@ -462,7 +462,7 @@ class OTA:
     def register_provider(self, provider: zigpy.ota.providers.BaseOtaProvider) -> None:
         """Register a new OTA provider."""
         if provider in self._providers:
-            _LOGGER.debug("Ignoring duplicate OTA provider: %s", provider)
+            _LOGGER.warning("Ignoring duplicate OTA provider: %s", provider)
             return
 
         _LOGGER.debug("Registering new OTA provider: %s", provider)
--- a/zigpy/ota/__init__.py
+++ b/zigpy/ota/__init__.py
@@ -461,6 +461,10 @@ class OTA:
 
     def register_provider(self, provider: zigpy.ota.providers.BaseOtaProvider) -> None:
         """Register a new OTA provider."""
+        if provider in self._providers:
+            _LOGGER.debug("Ignoring duplicate OTA provider: %s", provider)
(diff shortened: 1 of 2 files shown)


zigpy/ota/__init__.py
```diff
@@ -461,6 +461,10 @@ def register_deprecated_provider(
 
     def register_provider(self, provider: zigpy.ota.providers.BaseOtaProvider) -> None:
         """Register a new OTA provider."""
+        if provider in self._providers:
+            _LOGGER.debug("Ignoring duplicate OTA provider: %s", provider)
```
TheJulianJES: I think we may want to bump this to `WARNING`. It should only ever happen because of misconfigurations, though I guess the deprecated providers also log a (different) warning already – new ones don't.

TheJulianJES: [`360897c` (this PR)](https://github.com/zigpy/zigpy/pull/1839/changes/360897c68bee374790bffa43d8af4acd14f4957a)

puddly (approved):

PR #1840, opened 2026-06-16
Add an `allow_downgrade` flag to OTA updates

TheJulianJES:
## Proposed change

This adds an `allow_downgrade` flag, allowing for a ZHA PR, so you can use the native `update.install` service in HA to just choose an older version to downgrade to. For that, we need to relax some checks in ZHA, so that if a manual version is provided (from ZHA/HA), the "newer version check" is ignored in zigpy.

There's going to be a (small) corresponding ZHA PR for this.

In the future, I also wanna look if we can trigger the `force` option from HA, as we need this for some devices (but other devices also do not support it at all and only need the `allow_downgrade from this PR). Maybe we can just have `force` at the start or end of the firmware version...? 😄 

Related ZHA PR:
- https://github.com/zigpy/zha/pull/793

## AI summary – Problem

A caller of `Device.update_firmware(image=...)` that hands in a specific image **older** than the device's running firmware cannot install it without also passing `force=True`.

This is because `OTAManager._image_query_req` replies with `NO_IMAGE_AVAILABLE` whenever `not self.force and not self.image.check_version(command.current_file_version)`, and `check_version` returns `False` for any image whose version is `<=` the device's current version.

`force` is the only escape hatch today, but it conflates two unrelated things: it skips the version check **and** the compatibility check, and it rewrites the advertised file version to `0xFFFFFFFE` so the device always accepts the image. That's far too blunt for "the user deliberately picked an older but otherwise compatible firmware."

Downstream (e.g. ZHA's update entity) the user picks a specific firmware from a list and has no way to know a downgrade needs `force=True`. The device just gets `NO_IMAGE_AVAILABLE`, surfaced as:

> Update was not successful: `<Status.NO_IMAGE_AVAILABLE: 152>`

Real example: a frient WISZB-131 at `current_file_version=131079`, user selected `131078`, log shows the silent `NO_IMAGE_AVAILABLE` reply.

## Change

- Add an explicit `allow_downgrade: bool = False` parameter to `Device.update_firmware`, `zigpy.ota.manager.update_firmware`, and `OTAManager`, separate from `force`. When set, `_image_query_req` skips **only** the version check; compatibility (`check_compatibility`) is still enforced, and the real file version is still advertised (no `force`-style version rewrite).
- When `_image_query_req` is about to reply `NO_IMAGE_AVAILABLE`, log at `WARNING` which guard failed — compatibility vs. version — including the device's current file version and a hint to pass `allow_downgrade=True`. The cause is now debuggable from logs alone.
- Add unit tests covering an explicitly selected older image: it installs when `allow_downgrade=True` (full block-transfer path to `SUCCESS`) and is refused with `NO_IMAGE_AVAILABLE` when `allow_downgrade=False`.

The default behavior is unchanged: implicit updates still refuse downgrades, and `force=True` keeps skipping both checks for backward compatibility.

## Downstream (ZHA)

ZHA's `async_install(version)` already carries the user-intent signal:

- `version is None` → auto-install the newest upgrade (can never be a downgrade).
- `version is not None` → the user explicitly named a version, resolved against both upgrades and downgrades.

So the ZHA call becomes:

```python
result = await self.device.device.update_firmware(
    image=firmware,
    progress_callback=self._update_progress,
    allow_downgrade=(version is not None),
)
```

This keeps the version guard on for the automatic "install latest" path (where it also protects against acting on a stale cached image) and lifts it only when the user explicitly chose a version. (Separate ZHA PR.)

## Note on device behavior

`allow_downgrade` lets zigpy *offer* a downgrade; whether the device *accepts* it is up to the device firmware. Many devices hard-refuse downgrades at the bootloader. For example, a Sonoff S60ZBTPF running `0x2003` offered `0x2002` replies `query_next_image_response(status=SUCCESS, file_version=8194)` (the offer goes out correctly) but immediately responds `upgrade_end(INVALID_IMAGE)` without requesting a single block. In that case the user now sees the device's honest `INVALID_IMAGE` instead of zigpy's silent `NO_IMAGE_AVAILABLE`.

### DIFF
--- a/tests/ota/test_ota_manager.py
+++ b/tests/ota/test_ota_manager.py
@@ -254,6 +254,129 @@ async def test_ota_manager_downgrade_rejected_without_flag(
     assert status == foundation.Status.NO_IMAGE_AVAILABLE
 
 
+@patch("zigpy.ota.manager.MAX_TIME_WITHOUT_PROGRESS", 0.1)
+async def test_ota_manager_incompatible_image_rejected(
+    image_with_metadata: OtaImageWithMetadata,
+) -> None:
+    """An incompatible image is refused even when the version check would pass.
+
+    allow_downgrade only lifts the version check; compatibility is still enforced.
+    """
+    img = image_with_metadata
+
+    app = make_app({})
+    dev = app.add_device(nwk=0x1234, ieee=t.EUI64.convert("00:11:22:33:44:55:66:77"))
+    dev.node_desc = make_node_desc(logical_type=zdo_t.LogicalType.Router)
+    dev.model = "model1"
+    dev.manufacturer = "manufacturer1"
+
+    ep = dev.add_endpoint(1)
+    ep.status = zigpy.endpoint.Status.ZDO_INIT
+    ep.profile_id = 260
+    ep.device_type = zigpy.profiles.zha.DeviceType.PUMP
+
+    ota = ep.add_output_cluster(Ota.cluster_id)
+
+    async def send_packet(packet: t.ZigbeePacket):
+        assert img.firmware is not None
+
+        if packet.cluster_id == Ota.cluster_id:
+            hdr, cmd = ota.deserialize(packet.data.serialize())
+            if isinstance(cmd, Ota.ImageNotifyCommand):
+                dev.application.packet_received(
+                    make_packet(
+                        dev,
+                        ota,
+                        "query_next_image",
+                        field_control=Ota.QueryNextImageCommand.FieldControl.HardwareVersion,
+                        manufacturer_code=img.firmware.header.manufacturer_id,
+                        # The image type does not match the image, so it is
+                        # incompatible regardless of the version
+                        image_type=img.firmware.header.image_type + 1,
+                        # A version the image's metadata would otherwise accept
+                        current_file_version=img.firmware.header.file_version - 5,
+                        hardware_version=1,
+                    )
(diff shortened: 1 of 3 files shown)


TheJulianJES:
Random idea: could we auto-detect when we need to set the `force` flag by checking if the device immediately responds with "invalid image" when we don't set it in the first try when downgrading firmware? Do all devices respond that before the update is transmitted? Or do some just not apply the update?

puddly:
I don't think I remember exactly but I think there was _some_ device that waited until the end to declare the image was invalid? Or maybe it was that one Aqara device that got bricked?

TheJulianJES:
So, I had that with one Frient sensor where I accidentally sent with `force=True` the entire time. It transmitted the entire image ("query next image response" contains a newer version, so it makes sense to continue), and then it must've failed some internal validation at the end because of the modified header.

But I was able to downgrade that device with this `allow_downgrade` flag, so by just sending the unmodified older image.

But some devices also immediately decline the image you're trying to send if you use an older one (so with `allow_downgrade` but not `force`). I guess they just decline upon seeing the old version in the query next image response (or the first block response)?



So I'm just thinking that if we first try the unmodified older image, we should fail fast. Then we can automatically try with `force=True` afterwards. It would avoid the user needing to specify whether to modify the header or not when downgrading via `update.install` (which I expect very few people to do anyway).

Issue #1822, opened 2026-05-19
Potential Memory leak: event_base.py on_event() listeners accumulate when ZCL requests time out or are cancelled


pipiche38:
PS/ This issue has been developed with the help of claude.ai, my apologize if that is wrong.
I'm currently investigating memory leak in the Z4D plugin which use zigpy radio libs

[EDIT] Last, I do not have that issue on a zigpy/bellows, so it could be only for zigpy/zigpy_znp

**Environment**

- zigpy: 1.2.1
- zigpy-znp (coordinator driver)
- Python: 3.12
- Platform: Linux (Raspberry Pi 4, Fedora)
- Integration: Domoticz-Zigbee plugin (https://github.com/zigbeefordomoticz/Domoticz-Zigbee)

**Description**

We observed a steady, unbounded memory growth in a long-running Zigbee coordinator process. Using tracemalloc snapshots taken every 5 minutes, the dominant allocator is consistently [zigpy/event/event_base.py](https://github.com/pipiche38/Domoticz-Zigbee/blob/claude/fix-timeserver-dst-cache/zigpy/event/event_base.py) at lines 55, 57 and 58 — the body of on_event().

Growth rates observed:

- ~306 new listener objects per 5 minutes during idle periods (no pending commands)
- ~693 new listener objects per 5 minutes during active device communication

These listeners are never freed. Over hours of operation this accumulates to thousands of unreleased EventListener objects held in _event_listeners dicts across cluster and endpoint instances.

**Root cause**

on_event() correctly returns an unsubscribe callable:

```
# zigpy/event/event_base.py
def on_event(self, event_name, callback, with_context=False):
    listener = EventListener(...)          # line 55
    listeners = self._event_listeners \
                    .setdefault(event_name, [])  # line 57
    listeners.append(listener)             # line 58

    def unsubscribe():
        if listener in listeners:
            listeners.remove(listener)

    return unsubscribe
```

[...]

PR #1846, opened 2026-06-22
Move quirks API out of zigpy

puddly:
This is the zigpy third of https://github.com/zigpy/zha/pull/762.

This PR removes the entire quirks API out of zigpy (it is now split between ZHA and zha-device-handlers) and instead replaces it with a `device_resolver` hook, which allows a controlling application (in this case ZHA) to provide a way to override device construction. This will eventually be phased out as we slowly migrate away from V1 quirks and device-modify V2 quirks.


zigpy/appdb.py
```diff
@@ -418,10 +417,6 @@ async def _save_device(self, device: Device) -> None:
         if device.node_desc is not None:
             await self._save_node_descriptor(device)
 
-        if isinstance(device, zigpy.quirks.BaseCustomDevice):
-            await self._db.commit()
-            return
```
TheJulianJES, line 423: Note to check later: This was always a (not so great) workaround to not persist custom clusters/endpoints into the appdb when shutting down. We need to make sure we don't accidentally break this.

puddly: [`e13bd61` (this PR)](https://github.com/zigpy/zigpy/pull/1846/commits/e13bd61364d235389311a1a037262ef3ea4ddf3f) adds a few test cases for this exact scenario in addition to cleaning up the logic a bit. We now rely on the presence of `original_signature` to know if the device has been finalized or not.

TheJulianJES: Ah, ok. I thought we ran `_save_device` on shutdown for all devices as well but we do not.

(I wonder if we should have a docstring noting that `_save_device` should only be called with a non-quirked `Device`?)

zigpy/appdb.py
```diff
@@ -395,9 +394,16 @@ async def _remove_device(self, device: Device) -> None:
         await self._db.commit()
 
     def raw_device_initialized(self, device: Device) -> None:
-        self.enqueue("_save_device", device)
+        # We work with a clone of a device so that quirks that run in the same event
+        # loop iteration do not manage to mutate it before the async database code runs
+        self.enqueue("_raw_device_initialized", device.clone())
```
TheJulianJES, line 399: I guess this is nice but a quirk shouldn't touch the original `Device` instance, right? It's an entirely new `Device` object later that's quirked? So the reference we get here should always be the unquirked/original `Device`?

puddly: We had a few unit tests that mutated the device right after firing off the completion event. This somewhat matches the pattern we use elsewhere: make a copy of the data being pushed to the DB instead of passing along a live object. For devices it's a little more unwieldy so the clone sort of "enforces" it.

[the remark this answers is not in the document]
TheJulianJES: This PollControl check-in is indeed somewhat annoying 😅 

We already have that issue when the device instance is initially replaced with a quirked one, and also when re-interviewing a device. It’s the same root cause for which we already have two PRs (https://github.com/zigpy/zigpy/pull/1816 + https://github.com/zigpy/zigpy/pull/1795) and the issue is also seen [here](https://github.com/zigpy/zigpy/issues/1822#issuecomment-4489991041) by Domoticz for Zigbee. I was never really fully happy with those PRs though..

I think `clone()` having some sort of "info_only" arg that doesn't register the callback might be nice, even if we properly address the other leaks.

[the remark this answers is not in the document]
puddly: ZHA moved all of this logic into `on_add`. I think for now we can just explicitly clean up within appdb?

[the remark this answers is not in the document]
puddly: We _might_ also be able to use weakrefs?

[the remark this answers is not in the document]
TheJulianJES: ZHA using `on_add` for this sort of logic is nice. 

> I think for now we can just explicitly clean up within appdb?

My edit was a bit late, but below could also work until we refactor `Device` to use `on_add` (or weak references)?

> I think `clone()` having some sort of "info_only" arg that doesn't register the callback might be nice, even if we properly address the other leaks.

But I'm fine with also just cleaning up in the appdb afterwards. (`.clone()` is only called very rarely anyway.)

zigpy/appdb.py
```diff
@@ -395,9 +394,24 @@ async def _remove_device(self, device: Device) -> None:
...
+        try:
+            await self._raw_device_initialized_internal(device)
+        finally:
+            device.on_remove()
+
+    async def _raw_device_initialized_internal(self, device: Device) -> None:
+        # `device` is a clone snapshotted in `raw_device_initialized`, taken before any
+        # quirk had a chance to mutate the live device
+        if device.original_signature is not None:
+            raise ValueError("A device with quirks cannot be saved to the database")
```
TheJulianJES, line 413: nit: new uncovered line

zigpy/application.py
```diff
@@ -336,11 +340,23 @@ async def startup(self, *, auto_form: bool = False) -> None:
...
+        uninitialized_packet_handler: Callable[..., None] | None = None,
     ) -> ControllerApplication:
         """Create new instance of application controller."""
         app = cls(config)
 
+        if device_resolver is not None:
+            app.register_device_resolver(device_resolver)
+
+        if uninitialized_packet_handler is not None:
+            app.register_uninitialized_packet_handler(uninitialized_packet_handler)
```
TheJulianJES, line 358: nit: new uncovered line

TheJulianJES (commented):
Small nit since the lines are rather trivial, so feel free to ignore: There's no test coverage for these two added lines below. Everything else is basically just the backwards-compatible imports.

---

However, we would fall below our 99% `MINIMUM_COVERAGE_PERCENTAGE` on `dev` (so Codecov check would constantly be "red"): https://github.com/zigpy/zigpy/blob/43dea44d371938db6d9c37d1dd4f66b062256e29/.github/workflows/ci.yml#L13

We could add explicit coverage ignore comments for the backwards-compatible imports, or lower the minimum coverage?

zigpy/appdb.py
puddly: Both should now be covered.

puddly:
I think this side is good to go. All known issues at this point are with ZHA and the new APIs, not zigpy.

PR #1814, opened 2026-04-15
Add Zigbee Green Power (ZGP) protocol support

nmingam:
## Summary

Add comprehensive Zigbee Green Power (ZGP) protocol support to zigpy, implementing the GP Sink role as defined in the ZGP specification. This is the most requested missing feature (#341, 175+ thumbs-up).

### New modules (`zigpy/zgp/`)

| Module | Purpose |
|--------|---------|
| `types.py` | GP constants (endpoint 242, cluster 0x0021, default link key) and enums (SecurityLevel, GPDCommandID, CommunicationMode, etc.) per ZGP spec tables |
| `crypto.py` | AES-128-CCM* encrypt/decrypt for GP security keys and frame payloads, with auth-only (SecurityLevel 0b10) and full encryption (0b11) modes per spec A.1.5.4 |
| `frame.py` | GP Commissioning payload parser (Tables 53/54/55) and Channel Request parser |
| `device.py` | GPDevice dataclass with sourceID-to-IEEE conversion, frame counter replay protection, and dict serialization |
| `proxy.py` | GP Proxy Table tracking which proxy devices forward for which GPDs |
| `manager.py` | Central GP protocol handler — commissioning, pairing, notification dispatch, security, dedup |

### Manager capabilities (equivalent to zigbee-herdsman's `greenPower.ts`)

- GP Notification and Commissioning Notification processing
- Commissioning flow with security key decryption (encrypted and plaintext keys)
- GP Commissioning Reply for RX-capable GPDs
- GP Channel Configuration response
- GP Pairing broadcast with encrypted key and dynamic communication mode (unicast/groupcast)
- Duplicate filtering per spec A.3.6.1.2 (bounded cache, 2s timeout)
- Commissioning window management via ProxyCommissioningMode broadcast
- Shutdown cleanup

### Integration (`application.py`)

- GP frames (endpoint 242, cluster 0x0021) intercepted in `packet_received()` **before** device lookup, so frames from unknown proxy NWK addresses don't trigger device discovery
- `permit_gp(time_s)` method separate from standard `permit()`
- `green_power.shutdown()` called during controller shutdown

### Known limitations

- Only `ApplicationID.SrcID` (0b000) supported — same as zigbee-herdsman; no known consumer GPD uses IEEE mode
- `CommissioningNotificationSchema` has a bit-field alignment issue (18 bits) in the existing GP cluster definition from #1659 — documented in tests
[...]

PR #1847, opened 2026-07-02
zgp: add the Green Power types, frame, and crypto parsing core

nmingam:
Standalone Green Power parsing core, split out of #1814 so it can land on its own — following @puddly's suggestion to get the types/command parsing in first and shrink the larger PR.

## Contents
- `zgp/types.py`: GPD command IDs, GP endpoint/cluster/group constants, the default GP link key, and fixes to `SecurityLevel` and the `ProxyCommissioningModeExitMode` bitmap
- `zgp/frame.py`: GP Commissioning (0xE0) and Channel Request (0xE3) payload parsing and serialization
- `zgp/crypto.py`: AES-128-CCM* nonce construction, key and payload (de)encryption per ZGP spec A.1.5.4 / A.3.7.1.2.3
- `zcl/clusters/greenpower.py`: GP Commissioning Notification / Response / Proxy Commissioning Mode ZCL schema fixes
- `tests/zgp/`: `test_types`, `test_frame`, `test_crypto` — 100% line coverage of the four modules above

The manager, device, and proxy layers that build on this core are intentionally left out and will follow in separate PRs.

## Credits
- @puddly — the CCM* associated-data (SrcID) fix for GPD key (de)encryption (A.3.7.1.2.3); credited as co-author on the commit.
- Thanks to @konistehrad (original draft ZGP implementation), @mineshaftgap (radio-side/EZSP fixes) and @dmatscheko (real-hardware testing) for the broader ZGP effort this series is part of.


tests/zgp/test_types.py
```diff
@@ -0,0 +1,183 @@
...
+    d = DeviceID(0x12345678)
+    assert int(d) == 0x12345678
+
+
+def test_device_id_hex_repr():
+    d = DeviceID(0x02)
+    assert "0x" in repr(d).lower()
+
+
+def test_no_security():
```
puddly: I think you can remove any unit tests that assert that value of an enum. They're not really useful tests.

zigpy/zgp/crypto.py
```diff
@@ -0,0 +1,179 @@
+"""Green Power security primitives: AES-128-CCM* per ZGP spec A.1.5.4.
+
+Nonce is 13 bytes built from the source ID and frame counter; MIC length
```
puddly: This module comment can be removed entirely and just the first line kept, it's very verbose.

zigpy/zgp/frame.py
```diff
@@ -0,0 +1,348 @@
...
+
+import dataclasses
+import struct
+from typing import ClassVar
+
+from zigpy.zgp.types import SecurityKeyType, SecurityLevel
+
+
+@dataclasses.dataclass
+class GPCommissioningOptions:
```
puddly: You should be able to turn this into a subclass of `t.Struct` and `t.uint8_t`, it will clean up all of the manual parsing code.

puddly (commented):
Thanks! This looks good, just a few comments.

tests/zgp/test_types.py
nmingam: Dropped the bare `assert Enum.Member == 0bNN` tests — agreed they weren't pulling weight. Kept the handful that check actual behavior: the protocol constants (endpoint/cluster/group IDs, the default link key), `DeviceID` uint32 coercion + hex repr, and the bitmap OR-composition test.

zigpy/zgp/crypto.py
nmingam: Trimmed the module docstring to its first line. The AAD/nonce specifics that were in it now live as inline comments right at the `associated_data=` call sites where they're actually relevant.

zigpy/zgp/frame.py
nmingam: Rewrote frame.py on top of `t.IntStruct`/`t.Struct` — the options bytes are now IntStructs with the bit-fields declared directly, and the commissioning payload is a `Struct` with `requires`-gated optional fields, dropping all the manual masks and offset math (349->139 lines). One deliberate behavior change: `GPChannelRequestPayload` now exposes the raw channel nibbles; the +11 IEEE-channel offset is applied by the consumer. Tests updated for the `deserialize()`/`serialize()` API and KeyData/LVList field types.

puddly:
I think the bot review makes sense. Some points are a little pedantic (table and figure number drift) but the AAD omission, the rejection criteria, etc. and the switch info flags are good to have. It may be referencing a newer version of the spec, which added the "switch info" flag. It's unlikely any devices use it.

puddly:
I've applied the few bot commits to this branch since they cumulatively were quite small and all made sense. This PR looks good to me. Thank you!

PR #1848, opened 2026-07-02
zgp: add the Green Power manager, device, and proxy runtime

nmingam:
Adds the Green Power (GP / GPEP) runtime on top of the types/frame/crypto parsing core: the `GreenPowerManager` (commissioning, pairing, packet dispatch), the `GPDevice` model, the proxy table, and a typed event bus, plus the `ControllerApplication` wiring — GP frame interception on endpoint 242 / cluster 0x0021, `permit_gp`, and manager shutdown.

Green Power devices are persisted in the zigpy SQLite database (schema v16, `gp_devices` table): commissioned GPDs and their frame counters survive a restart, with the frame counter updated per press (not only at join) to keep replay protection correct.

## Stacked on #1847

This PR is stacked on **#1847** (the GP types/frame/crypto parsing core). Review and merge #1847 first; the diff here is the manager-stage delta only. Until #1847 merges, its single commit shows in this diff and drops automatically on rebase.

## Scope

- Included: GP runtime (manager, device, proxy, events), `application.py` wiring, and SQLite persistence (schema v16 + v15→v16 migration).
- Excluded: GP quirks (`CustomGreenPowerDevice` / GP quirk registry) — a quirks refactor is in flight upstream, so that work is tracked separately.

## Credits

- **Kenn Herman (@mineshaftgap)** — Green Power SQLite persistence (schema v16, per-press frame-counter persistence, and its tests). His commits are preserved verbatim with original authorship.
- **konistehrad** — the original draft ZGP implementation that informed this work.
- **puddly (@puddly)** — the GP crypto (SrcID AAD) fix, which ships in #1847 (the stacked base), not in this PR.


nmingam:
Next on my list is your point 3, splitting `Device` into `BaseDevice` / `ZigbeeDevice` / `GreenPowerDevice`. Two questions before I write it, because both are cheap to answer now and expensive to undo later.

**1. Do you want to write the split yourself?** It touches the most-subclassed class in the ecosystem and you may already have a shape in mind. I am happy to do it or to stay out of it, but I would rather not hand you a competing version of something you were going to write.

If I do write it, my plan is a thin `BaseDevice` holding only identity and liveness - `application`, `ieee`, `manufacturer`, `model`, `last_seen`, `lqi`, `rssi`, `skip_configuration`, `create_task`, `on_remove` - with everything mesh-shaped staying exactly where it is. Two things I found while scoping it that are worth flagging either way:

- `Device.__init__` never calls `super().__init__()`; it hand-assigns `self._listeners = {}`. So `ListenableMixin.__init__` has never actually run for a `Device`. Harmless today, but reparenting has to fix it rather than inherit the bug.
- Keeping the concrete class named `Device` with `ZigbeeDevice` as an alias avoids churning `quirk_class` in diagnostics, which is built from `__class__.__module__ + __name__` and appears in 479 zha test fixtures. Renaming outright is a one-line change if you would rather have the name.

**2. Do Green Power devices belong in `app.devices`, or in a parallel registry?** This forks everything downstream, and I read your "rework ZHA to work with information from both" as pointing at two sources, so that is what I have been building toward - but I would rather confirm than assume.

The case for keeping them out: `app.green_power.devices` is already a public property, and zha's `get_or_create_device` keys on `zigpy_device.ieee` and never consults `app.devices`, so the second enumeration costs about ten lines in zha and no new zigpy API. Putting them in `app.devices` means either a `DB_VERSION` bump - and since `DB_V` is a global table-name suffix, v17 recreates all thirteen tables - or a device sitting in the registry while being deliberately excluded from that registry's persistence, which is the worse half of the two.

If you would rather have them in `app.devices`, say so and I will do that instead; the class hierarchy is the same either way, so it is a small change of direction now and a large one later.

zigpy/application.py
```diff
@@ -180,11 +183,14 @@ def _add_db_listeners(self):
         self.groups.add_listener(self._dblistener)
         self.backups.add_listener(self._dblistener)
         self.topology.add_listener(self._dblistener)
+        if hasattr(self, "green_power"):
```
puddly: Why check for the attribute? I think we can drop this check.

zigpy/appdb.py
```diff
@@ -1151,6 +1236,56 @@ async def _load_ota_query_cache(self) -> None:
                         cluster.last_query_cmd = cmd
                         break
 
+    async def _load_gp_devices(self) -> None:
+        if not hasattr(self._application, "green_power"):
```
puddly: Same `hasattr` check here.

zigpy/appdb.py
```diff
@@ -725,6 +726,89 @@ async def _delete_ota_query_cache_entry(
...
+        device = event.device
+        self.enqueue(
+            "_update_gp_frame_counter",
+            device.source_id,
+            device.frame_counter,
+            device.last_seen.isoformat() if device.last_seen else None,
+        )
+
+    async def _save_gp_device(self, device) -> None:
+        d = device.as_dict()
```
puddly: I think it would be best for `appdb` to handle the serialization/deserialization internally, as these are to an extent database operations, not something directly implemented by the device itself.

zigpy/zgp/manager.py
```diff
@@ -0,0 +1,829 @@
...
+        """
+        LOGGER.info(
+            "Sending GP Commissioning Reply to RX-capable GPD 0x%08X",
+            source_id,
+        )
+
+        # Commissioning Reply payload (cmd 0xF0):
+        # 1 byte options: 0x00 = no PAN ID, no key, no key encryption,
+        #                 no security level
+        commissioning_reply_payload = bytes([0x00])
```
puddly: To avoid hardcoding manually-serialized byte sequences, can `GPCommissioningReplyPayload` be added to `zgp/frame.py`?

zigpy/zgp/manager.py
```diff
@@ -0,0 +1,829 @@
+"""Green Power Manager: processes GP frames on cluster 0x0021, endpoint 242.
+
+Manages GP device lifecycle (commissioning, decommissioning) and dispatches
+GP commands to listeners. Equivalent to zigbee-herdsman's greenPower.ts.
```
puddly: Is this implementation a translation from zigbee-herdsman? We try to implement things from the specifications, as otherwise bugs get copy/pasted between libraries. If not, let's remove these LLM-isms: `matching zigbee-herdsman` appears six times within this PR.

zigpy/zgp/events.py
```diff
@@ -0,0 +1,44 @@
...
+class DeviceLeft:
+    """A commissioned Green Power device was decommissioned or removed."""
+
+    event_type: Final[str] = "gp_device_left"
+
+    device: GPDevice
+
+
+@dataclasses.dataclass(kw_only=True, frozen=True)
+class CommandReceived:
```
puddly, line 37: Events that pass through `payload: bytes` feel like they should be internal (e.g. `RawCommandReceived`), since we should be passing through structured objects wherever possible.

tests/databases/simple_v15.sql
```diff
@@ -0,0 +1,138 @@
...
+    manufacturer_code INTEGER NOT NULL,
+    image_type INTEGER NOT NULL,
+    current_file_version INTEGER NOT NULL,
+    hardware_version INTEGER,
+    last_updated REAL NOT NULL,
+    FOREIGN KEY(ieee, endpoint_id) REFERENCES endpoints_v15(ieee, endpoint_id) ON DELETE CASCADE
+);
+CREATE UNIQUE INDEX ota_query_cache_idx_v15 ON ota_query_cache_v15(ieee, endpoint_id);
+
+INSERT INTO devices_v15 VALUES('00:11:22:33:44:55:66:77', 1, 2, 0.0);
```
puddly: I don't think this database fixture is required, the new schema is additive and will be covered by existing tests that upgrade our older databases.

tests/zgp/test_real_frames.py
```diff
@@ -0,0 +1,152 @@
...
+    BJ6716U_OPERATIONAL_FRAMES,
+    BJ6716U_SOURCE_ID,
+)
+from zigpy.zgp.device import GPDevice
+from zigpy.zgp.events import CommandReceived, DeviceJoined
+from zigpy.zgp.manager import GreenPowerManager
+from zigpy.zgp.types import SecurityLevel
+
+
+async def _commission(manager: GreenPowerManager) -> None:
```
puddly: This helper is used in only two places. Let's merge the contents into the call sites, indirection like this makes tests hard to read.

tests/test_appdb.py
```diff
@@ -2083,3 +2083,130 @@ async def test_get_last_ota_query_cmd_returns_none(tmp_path):
     assert dev.get_last_ota_query_cmd() is None
 
     await app.shutdown()
+
+
+async def test_no_sidecar_file_created(tmp_path):
```
puddly: Where would this `zgp_devices.json` come from?

tests/test_appdb.py
```diff
@@ -2083,3 +2083,130 @@ async def test_get_last_ota_query_cmd_returns_none(tmp_path):
...
+    device = GPDevice(source_id=0x0040F4E4, device_id=2)
+    app.green_power.emit(DeviceJoined.event_type, DeviceJoined(device=device))
+    await app.shutdown()
+
+    assert not (tmp_path / "zgp_devices.json").exists()
+
+
+async def test_gp_device_round_trip(tmp_path):
+    """Emit DeviceJoined, shutdown, reopen - device restored with all fields intact."""
+    from zigpy.zgp.device import GPDevice
```
puddly: Please move all imports out of function bodies and to the top of the testing module.

tests/test_appdb_migration.py
```diff
@@ -784,3 +784,43 @@ def resolver(device):
             (str(dev.ieee), 0xFC02, 0x0020),
         ).fetchall()
         assert rows == [(-1,)]
+
+
+async def test_gp_device_tables_fresh_db(tmp_path):
```
puddly: These tests can be dropped as their functionality is already covered elsewhere.

zigpy/appdb_schemas/schema_v16.sql
```diff
@@ -0,0 +1,242 @@
...
+    frame_counter INTEGER NOT NULL,
+    manufacturer_id TEXT,
+    model_id TEXT,
+    gpd_commands TEXT NOT NULL,
+    server_clusters TEXT NOT NULL,
+    client_clusters TEXT NOT NULL,
+    mac_seq_num_capability INTEGER NOT NULL,
+    rx_on_capability INTEGER NOT NULL,
+    fixed_location INTEGER NOT NULL,
+    last_seen TEXT
```
puddly: Why is this a string? We use a numerical type for the "normal" Zigbee device table.

zigpy/zgp/events.py
```diff
@@ -0,0 +1,44 @@
...
+from zigpy.zgp.types import GPDCommandID
+
+
+@dataclasses.dataclass(kw_only=True, frozen=True)
+class DeviceJoined:
+    """A new Green Power device was commissioned."""
+
+    event_type: Final[str] = "gp_device_joined"
+
+    device: GPDevice
```
puddly: We should have these events emit IEEE addresses, as serializing a `GPDevice` object may not be possible in the future.

puddly (commented):
Thanks for the PR! I've left a few comments.

Regarding the `Device` base class: I'm going to spend a bit more time working on how this integrates with ZHA but I think the core of this PR (the manager, database schema, proxying, etc.) should be good.

One concern I have are the myriad references to `zigbee-herdsman`: is this a translation of the GP code from zigbee-herdsman? We generally perform direct spec implementations in zigpy and then validate with other repos after the fact, as otherwise structural design issues and bugs can migrate across projects.

nmingam:
Thanks for the review, and for #1872 - the command payloads landing upstream made the events change much smaller than it would have been. Rebased onto it and worked through the comments. Taking the substantive ones first.

**Is this a translation from zigbee-herdsman?** No. It was written against the specification, in Python, from the start. Where I used herdsman was as a cross-reference: when the spec left something ambiguous, or when a tester reported a device behaving in a way my reading did not predict, I checked what the working implementation did before deciding. Two of those checks turned out to be real bugs on my side, including one where the GPD key was going out in the clear in the GP Pairing.

That is still not the same as "spec first, validate after", and your point stands that leaving `matching zigbee-herdsman` in the source sends the next reader to the wrong document. Every reference is gone. Each one is replaced by what it was actually standing on: the key in the GP Pairing is protected per A.3.7.1.2.3, the same section `crypto.py` already cites, and the SrcID-only support is stated as the limitation it is.

One of them I could not replace with a citation, and that is worth flagging rather than papering over. The `send_pairing` docstring claimed the communication-mode choice matched herdsman. Going back to the spec, `gpsCommunicationMode` is a sink attribute and Table 27 only enumerates the values - nothing requires deriving the mode per pairing from whether a proxy forwarded the notification. So that logic is this sink's own policy, and the docstring now says exactly that instead of implying a normative basis. If you would rather it followed a fixed configured mode, that is a small change and I would rather hear it now.

**Structured events.** `CommandReceived.payload` is now the parsed struct from `zgp/commands`, and the events carry `device_ieee` instead of the `GPDevice`, matching the shape the ZCL events already use. Commands with no schema, and payloads that fail to deserialize, go out as `RawCommandReceived` rather than being dropped: the frame counter has already advanced by then, so the replay baseline still has to be persisted, and one device sending a malformed frame should not silence the others. Adding a schema to `zgp/commands` moves a command from one event to the other. Trailing bytes after a successful parse are logged and the parsed value is still emitted, following what `zcl/__init__.py` does for a ZCL frame.

**appdb owning serialization.** `as_dict`/`from_dict` are gone from `GPDevice`, along with `load_devices`/`get_devices_data` on the manager, which only existed to shuttle those dicts. `_save_gp_device` and `_load_gp_devices` now do the column mapping the way `_raw_device_initialized_internal` and `_load_devices` already do it, and a restored device goes back in through the manager's own `add_device()`. One consequence worth naming: loading no longer skips a row it cannot parse, matching `_load_devices`.

**The schema column.** Fixed, and I took two more in the same table while I was there: `manufacturer_id` and `model_id` were also TEXT, while `GPDevice` holds them as ints, so SQLite's affinity was converting them and a restored device came back with strings where a freshly commissioned one had numbers. `GPDevice.last_seen` now mirrors `zigpy.device.Device` - a private datetime behind a float property whose setter also accepts a timestamp.

**One thing to know before testing this branch.** Schema v16 has never shipped, so I changed the columns in place rather than adding a v17. Anyone who already ran an earlier build of this PR has a database at `user_version = 16` with the old text columns; migrations will skip it and the GP load will fail on startup. Deleting the `gp_devices_v16` rows, or the database, is enough. Say the word if you would rather have the v17 migration instead.

The rest are done as asked: both `hasattr` checks dropped (and a `getattr` default next to them that would have drawn the same comment), the v15 fixture and the two migration tests removed, `test_no_sidecar_file_created` deleted - you were right that nothing in the tree ever writes that file, so it was asserting against something that could never happen - the test imports hoisted, and `_commission()` inlined into its two call sites. `GPCommissioningReplyPayload` and `GPChannelConfigurationPayload` both come from your #1872 now, so both hand-packed byte sequences are gone; I checked the serialized bytes are unchanged for every channel in range.

Kenn Herman's persistence commits are still in here with their authorship intact.

bedge117:
Tested this branch end-to-end against real hardware: Home Assistant SkyConnect (EmberZNet 7.5.1, EZSP v13) via bellows, with an EnOcean PTM 215Z (Philips Hue Tap) as the GPD.

**Setup.** Two things were needed outside this PR to get frames in at all:

1. bellows' `gpepIncomingMessageHandler` schema fails to parse EmberZNet 7.x frames (fix submitted as zigpy/bellows#746).
2. bellows does not route parsed gpep frames anywhere. For the test I subclassed `ControllerApplication`, intercepted `gpepIncomingMessageHandler` in `ezsp_callback_handler`, and re-emitted it through `packet_received` as a `ZigbeePacket` to endpoint 242 / cluster 0x0021 carrying a ZCL `GreenPowerProxy` server command: `commissioning_notification` for `0xE0` frames, `notification` otherwise, with `NotificationOptions`/`CommissioningNotificationOptions` filled from the EZSP fields (application_id, security_level, security_key_type).

**Results.**

- Frames from the not-yet-commissioned device were dropped by `_dispatch_gp_command` as expected ("GP command from unknown device").
- Commissioning: holding the channel-11 button for 10 s produced the `0xE0` frame (security level 0, counter `0xFFFFFFFF`, 27-byte payload, no AppInfo). `_process_commissioning` unwrapped the key, registered the device (`device_id=0x02`, `security=FullFrameCounterAndMIC`, `IndividualKey`) and emitted `DeviceJoined`.
- Operational frames after that: parsed by `_handle_gp_notification` and dispatched as `CommandReceived` with the expected commands (`Toggle`, `RecallScene1`, `RecallScene2`), one per physical press, frame counter incrementing.
- The `schema_v16` GP device table round-trips: on restart the device is restored from the database.

**One finding.** The GPD re-broadcasts its commissioning frame several times while the button is held; each one was processed as a fresh commissioning and emitted another `DeviceJoined` for the same source ID. `_process_commissioning` creates and adds a new `GPDevice` unconditionally, so an already-known device probably wants idempotent handling there.

I have the captures as a fixture in the same shape as `tests/zgp/fixtures/busch_jaeger_6716u.py` (commissioning payload re-wrapped with a fixture key rather than the device's own, operational frames as captured) plus tests, including an `xfail(strict=True)` for the finding above. Happy to open that as a PR against this branch if useful.

PR #1859, opened 2026-07-15
Re-read `INSUFFICIENT_SPACE` attributes

### LINKED ISSUE #1855: `read_attributes`: follow up on `INSUFFICIENT_SPACE` records with smaller reads (ZCL R8 §2.5.2.3)
## Summary

`Cluster.read_attributes()` treats a per-record `INSUFFICIENT_SPACE` status in a Read Attributes Response as a terminal failure. Per ZCL R8 §2.5.1.3, that status means the attribute **is supported** but its value record didn't fit into the remaining response-frame space, and §2.5.2.3 gives analogous follow-up guidance for records that didn't fit (its literal case is trailing records omitted from the response entirely):

> If fragmentation is not supported, and some trailing attribute status records have not been returned, due to space limitations in the frame, the originator may issue an additional Read Attributes or Read Attributes Structured command to obtain their values.

zigpy never issues that follow-up read, so the attribute's value is lost for that call. The per-record status is exposed in the returned `failure` dict, so a consumer *could* special-case it — but each consumer would have to duplicate the retry policy, and ZHA's initialization path currently ignores read failures entirely. Since `read_attributes()` owns the chunking (`MAX_READ_ATTRIBUTES_PER_REQ = 5`, [`zigpy/zcl/__init__.py#L45`](https://github.com/zigpy/zigpy/blob/2.0.0/zigpy/zcl/__init__.py#L45)) and the per-record status handling ([`#L1172-L1186`](https://github.com/zigpy/zigpy/blob/2.0.0/zigpy/zcl/__init__.py#L1172-L1186), [`#L1256`](https://github.com/zigpy/zigpy/blob/2.0.0/zigpy/zcl/__init__.py#L1256)), it's the natural place to handle this once for all consumers.

## Motivating case

zigpy/zha#826 (full analysis in [this comment](https://github.com/zigpy/zha/issues/826#issuecomment-4959743123)): during the post-OTA re-interview of a ThirdReality Smart Filter Sensor (3RAP0149BZ, fw 1.00.24), ZHA's batched read of the `AnalogInput` config attributes came back with `description` marked `INSUFFICIENT_SPACE` while the other four records in the same frame succeeded:

```
12:06:06.170  [0x6D1D:1:0x000c] Sending request: Read_Attributes(attribute_ids=[28, 65, 69, 81, 103])
12:06:06.453  [0x6D1D:1:0x000c] Decoded ZCL frame: AnalogInput:ReadAttributesResponse(status_records=[
                ReadAttributeRecord(attrid=28, status=<Status.INSUFFICIENT_SPACE: 137>, value=None),
                ReadAttributeRecord(attrid=65, status=<Status.SUCCESS: 0>, value=TypeValue(type=Single, value=0.0)),
                ReadAttributeRecord(attrid=69, status=<Status.SUCCESS: 0>, value=TypeValue(type=Single, value=0.0)),
                ReadAttributeRecord(attrid=81, status=<Status.SUCCESS: 0>, value=TypeValue(type=Bool, value=<Bool.false: 0>)),
                ReadAttributeRecord(attrid=103, status=<Status.SUCCESS: 0>, value=TypeValue(type=enum8, value=<enum8.undefined_0x00: 0>))])
```

Because the failure is terminal, `description` stays unknown and ZHA drops the prospective `AnalogInput` sensor entity at its support check, which requires a description. A solo read of the attribute would plausibly succeed — the response above used only ~30 bytes, so the device is either carrying a long description string, applying a conservative internal buffer limit, or simply mis-reporting the status; a follow-up read with more room distinguishes these. (The device side is being checked in the zha issue; the client-side gap exists regardless.)

## Proposed behavior

In the chunked read loop of `read_attributes()`:

1. After processing a chunk's response, collect the attribute IDs whose records came back `INSUFFICIENT_SPACE`.
2. Re-issue each such attribute in its own solo read — one retry pass, giving the value the whole frame.
3. Bound it: an attribute that still returns `INSUFFICIENT_SPACE` when read **alone** is a terminal failure (broken device accounting or a value that genuinely can't fit unfragmented) — record it in `failure` as today.
4. The same pass should cover §2.5.2.3's literal case: requested attributes whose records were **omitted from the response entirely**. Today those end up in *neither* the `success` nor the `failure` dict ([`#L1181-L1188`](https://github.com/zigpy/zigpy/blob/2.0.0/zigpy/zcl/__init__.py#L1181-L1188) — the loop only walks returned records), so they're silently lost rather than even marked failed.

Notes:

- Manufacturer-code grouping is preserved (the retry happens within the group's loop).
- The extra reads only happen when a device actually returns `INSUFFICIENT_SPACE` (or omits records), so there's no added traffic in the common case — relevant for sleepy end devices like the motivating one.
- Side benefit: attributes that today come back as `INSUFFICIENT_SPACE` failures — or, for omitted trailing records, appear in *neither* returned dict — would resolve to normal successes where possible.

puddly:
Fixes #1855.

If a batch of attributes contains an `INSUFFICIENT_SPACE` failure, we should re-read the affected attributes _once_ in isolation to ensure they have as much room as possible in the frame.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -414,7 +414,10 @@ def mock_attribute_reads(
             if attr_def in mock_reads:
                 value = mock_reads[attr_def]()
 
-                if isinstance(value, foundation.Status):
+                if value is None:
+                    # Omit the record from the response entirely
+                    continue
+                elif isinstance(value, foundation.Status):
                     record.status = value
                 else:
                     record.status = foundation.Status.SUCCESS
--- a/tests/test_zcl.py
+++ b/tests/test_zcl.py
@@ -2639,6 +2639,132 @@ async def test_read_attributes_chunked_by_count(app_mock) -> None:
     }
 
 
+async def test_read_attributes_insufficient_space_retry_success(app_mock) -> None:
+    """An INSUFFICIENT_SPACE record is re-read individually and can then succeed."""
+
+    class TestCluster(Basic):
+        _skip_registry = True
+
+        class AttributeDefs(Basic.AttributeDefs):
+            attr_0 = foundation.ZCLAttributeDef(id=0xFF00, type=t.uint8_t)
+            attr_1 = foundation.ZCLAttributeDef(id=0xFF01, type=t.uint8_t)
+            attr_2 = foundation.ZCLAttributeDef(id=0xFF02, type=t.uint8_t)
+
+    dev = add_initialized_device(app_mock, nwk=0x1234, ieee=make_ieee(1))
+    cluster = TestCluster(dev.endpoints[1])
+    dev.endpoints[1].add_input_cluster(TestCluster.cluster_id, cluster)
+
+    attrs = [
+        TestCluster.AttributeDefs.attr_0,
+        TestCluster.AttributeDefs.attr_1,
+        TestCluster.AttributeDefs.attr_2,
+    ]
+
+    supported = {
+        TestCluster.AttributeDefs.attr_0: 10,
+        # Runs out of space in the batched read, but fits when read alone
+        TestCluster.AttributeDefs.attr_1: mock.Mock(
+            side_effect=[foundation.Status.INSUFFICIENT_SPACE, 20]
+        ),
+        TestCluster.AttributeDefs.attr_2: 30,
+    }
+
+    with mock_attribute_reads(cluster, supported) as (mock_read, _):
+        success, failure = await cluster.read_attributes(attrs)
+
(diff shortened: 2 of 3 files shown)


TheJulianJES:
Forgot to open this one 😅: https://github.com/zigpy/zigpy/tree/zigpy-bot/read-attrs-insufficient-space. The tests in this PR are much better though.

TheJulianJES (approved):
Nice! Looks good.

PR #1864, opened 2026-07-27
Add coordinator group subscription APIs

puddly:
Some coordinators (namely SiLabs without our firmware patches) require groups to be explicitly registered in order for packets addressed to that group to be forwarded by the firmware.

These new APIs will be used by quirks to better support the few devices that require explicit group registration.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -171,10 +171,10 @@ class App(BaseApp):
         else:
             return 10
 
-    async def _subscribe_to_multicast_group(self, group_id: t.GroupId) -> None:
+    async def _subscribe_to_multicast_group(self, group_id: t.Group) -> None:
         pass
 
-    async def _unsubscribe_from_multicast_group(self, group_id: t.GroupId) -> None:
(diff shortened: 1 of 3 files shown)


TheJulianJES:
@zigpy-review-bot Review. Also check how this is/was done in quirks/ZHA, e.g. for IKEA remotes or the LightLink group.

TheJulianJES:
@zigpy-review-bot Also check linked quirks PR (https://github.com/zigpy/zha-device-handlers/pull/5209) and possibly https://github.com/zigpy/bellows/issues/708.

TheJulianJES:
@zigpy-review-bot Check if you read my previous mentions. Also, make this (review) less spammy overall. Use /details and edit existing body/reviews if it makes sense.

puddly:
I don't think we need to worry about `endpoint_id` at the moment, no other coordinator firmwares rely on it and SiLabs should be no different. We can always add it as an optional kwarg in the future.

PR #1874, opened 2026-08-10
Split up `BaseDevice` into `GreenPowerDevice` and `ZigbeeDevice`

puddly:
Zigpy's `Device` is the only abstraction for communicating with a device object and currently assumes that a Zigbee device _must_ be a ZCL device. This is no longer true. In addition to normal, spec-compliant devices, we have:
1. Green Power devices
2. Tuya devices (a superclass of ZCL devices that have datapoints)
3. XBee devices (which use ZCL to tunnel an unrelated protocol)
4. ???

This PR creates a (very) generic `BaseDevice` class and creates `GreenPowerDevice`, which has derived NWK addresses (per the ZGP spec) and synthetic IEEE addresses (not in the spec). Green Power devices now have their own packet structure (`ZigbeeGpPacket`) and this PR converts ZCL-tunneled ZGP frames into this structure. I'm leaving coordinator GP sinks out of this PR, we can address it in another one.

The ZGP device manager, database schema, etc. are part of https://github.com/zigpy/zigpy/pull/1848.

---

Part of https://github.com/OpenHomeFoundation/roadmap/issues/195

PR #1877, opened 2026-08-18
Change struct bitfield endianness from big to little

puddly:
Caught in a review of https://github.com/zigpy/zigpy/pull/1874.

Zigpy's `t.Struct` deals with byte alignment a little inconsistently (it's effectively mixed little and big endian...). This affected GP's `CommissioningNotificationOptions` in a way that made if unfixable without intermediate structs. This PR switches zigpy (and all code relying on it) to little endianness.

### DIFF
--- a/tests/test_struct.py
+++ b/tests/test_struct.py
@@ -620,7 +620,9 @@ def test_bitstruct_misaligned():
         baz: t.uint7_t
 
     s = TestStruct(foo=0b1, bar=0b10101010, baz=0b1110111)
-    assert s.serialize() == bytes([0b1110111_1, 0b0101010_1])
+
+    # The segment spans two bytes, so the first byte holds the first-declared fields
+    assert s.serialize() == bytes([0b0101010_1, 0b1110111_1])
 
     s2, remaining = TestStruct.deserialize(s.serialize() + b"asd")
     assert s == s2
@@ -630,9 +632,9 @@ def test_bitstruct_misaligned():
 
 
 def test_bitstruct_multi_byte_segment():
-    """A bitfield segment wider than one byte can be laid out little-endian."""
+    """A bitfield segment wider than one byte is little-endian."""
 
-    class TestStruct(t.Struct, bitfield_endianness="little"):
+    class TestStruct(t.Struct):
         foo: t.uint3_t
         bar: t.uint1_t
         baz: t.uint2_t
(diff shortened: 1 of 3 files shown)


zigpy/zcl/clusters/greenpower.py
```diff
@@ -16,8 +16,9 @@
 import zigpy.zgp.types as zgptypes
 
 
-# Figure 31 — 16 bits total
-class CommissioningNotificationOptions(t.Struct):
+# Figure 31 — 16 bits total. `security_key_type` straddles the byte boundary, so all
+# 16 bits are a single segment, which ZCL lays out little-endian like any `bitmap16`
+class CommissioningNotificationOptions(t.Struct, bitfield_endianness="little"):
```
TheJulianJES: nit: As mentioned in the parked review, I think as long as we have specific behavior just for `CommissioningNotificationOptions` with the `bitfield_endianness` kwarg, a short test for this makes sense.

tests/test_struct.py
```diff
@@ -629,6 +631,108 @@ class TestStruct(t.Struct):
         TestStruct.deserialize(b"\xff")
 
 
+def test_bitstruct_multi_byte_segment():
```
TheJulianJES, line 634: Duplicate test name: `test_bitstruct_multi_byte_segment`. This really should have been caught by CI... 😅. Maybe we don't want to [disable `F811` for tests](https://github.com/zigpy/zigpy/blob/8fccc1a5c30804c46f2383027c1b9cc79d4c2f8f/pyproject.toml#L242)?

(Perhaps we should also enable the re-trigger Copilot option on changes, or request zigpy-review-bot but it takes a bit longer – always a nice dobule-check though 😄.)

tests/test_struct.py
```diff
@@ -629,6 +631,108 @@ class TestStruct(t.Struct):
...
+        qux: t.uint5_t
+
+    s = TestStruct(foo=0b1010101, bar=0b11110000, baz=0b1001, qux=0b11011)
+    value = 0b1010101 | (0b11110000 << 7) | (0b1001 << 15) | (0b11011 << 19)
+
+    assert s.serialize() == t.uint24_t(value).serialize()
+    assert TestStruct.deserialize(s.serialize() + b"asd") == (s, b"asd")
+
+
+def test_bitstruct_multi_byte_segment() -> None:
```
TheJulianJES, line 701: New test is named the same as an existing one, so one is never run: `test_bitstruct_multi_byte_segment`

zigpy/zcl/clusters/greenpower.py
TheJulianJES: Oops, this was already addressed (both by adding the  test and (correctly) changing the default).

PR #1887, opened 2026-09-14
Keep destination unsupported marks when cloning an attribute cache

### LINKED ISSUE #1875: Attribute cache restore from zigbee.db clears unsupported-attribute marks declared by quirks
**Describe the bug**

`AttributeCache.set_value()` unconditionally removes the attribute's unsupported marker before writing the value (`zigpy/zcl/helpers.py`):

```python
def set_value(self, attr_def, value, *, last_updated=None):
    self.remove_unsupported(attr_def)
    ...
```

The `zigbee.db` restore path writes every persisted attribute row back through this method at startup (`zigpy/appdb.py`, `cluster._attr_cache.set_value(...)`).

Together these mean a quirk that declares an attribute unsupported at cluster instantiation — the documented `add_unsupported_attribute()` pattern, used by many quirks to tell consumers "this device will never provide this attribute" — has that declaration silently erased on every startup whenever a stale cached value for the attribute exists in the database. Stale values are easy to acquire: any earlier quirk version (or the bare cluster) that once populated the attribute leaves rows behind, and quirk swaps don't purge them.

Clearing the marker on `set_value()` is sound semantics for **live traffic** — a successful read or report is genuine evidence the device supports the attribute. The database restore is not new evidence, though: it just replays history, possibly from a different quirk's era, and it happens *after* the current quirk's clusters have been instantiated and have made their declarations.

**Impact (downstream, zha)**

zha's entity discovery gates sensor creation on `Cluster.is_attribute_unsupported()`. With the mark wiped by the restore, zha creates entities for attributes the active quirk explicitly declared unavailable. Nothing ever feeds them, so users get permanently-"Unknown" entities that cannot be deleted (the integration keeps providing them) and reappear after every restart.

Concrete case: the Tuya PJ-1203A quirk in zigpy/zha-device-handlers#4498 marks `Metering.instantaneous_demand` and several `ElectricalMeasurement` attributes unsupported. Users migrating from an earlier custom quirk that populated those attributes get zombie "Instantaneous demand" / "Power factor" sensors on every start (HA 2026.8.1: zigpy 2.1.0 / zha 2.1.0 / zha-quirks 2.2.0). Reproduced end-to-end with the zha test harness: a fresh join is clean; the same join with pre-seeded cache values recreates every zombie.

**Minimal reproduction (pure zigpy 2.1.0)**

```python
from unittest.mock import MagicMock
from zigpy.zcl.clusters.smartenergy import Metering

cluster = Metering(MagicMock())
attr = Metering.AttributeDefs.instantaneous_demand

cluster.add_unsupported_attribute(attr.id)
assert cluster.is_attribute_unsupported(attr.name)  # True — quirk's declaration

# What the zigbee.db attribute restore does on startup (zigpy/appdb.py):
cluster._attr_cache.set_value(attr, 1234)
assert not cluster.is_attribute_unsupported(attr.name)  # mark silently erased
```

**Expected behavior**

An unsupported mark set programmatically on the current cluster instance (i.e. by the active quirk) should survive the persistence restore. Values written by live reads/reports clearing the mark is fine.

**Possible directions**

- Give the restore path a way to write values without clearing marks (e.g. `set_value(..., clear_unsupported=False)` used by appdb), or
- Skip restoring cached values for attributes currently marked unsupported on the instantiated (quirked) cluster, or
- Re-assert quirk-declared marks after the restore completes.

Any of these would also stop the stale rows from being re-persisted indefinitely.

**Workaround**

Quirk-side, on zha-quirks 2.x: suppress the affected default entities with `prevent_default_entity_creation(...)` (proposed for the PJ-1203A quirk in zigpy/zha-device-handlers#4498). This hides the symptom but doesn't restore the `add_unsupported_attribute()` contract for other quirks.

**Environment**

- zigpy 2.1.0 (Home Assistant 2026.8.1, zha 2.1.0, zha-quirks 2.2.0)
- Radio: n/a (persistence-layer behavior, reproducible without a radio)

cpruijsen:
Restoring the attribute cache from `zigbee.db` clears unsupported-attribute marks that a quirk
declared. The quirk knows the device does not support an attribute, and a stale cached value from
before the quirk applied overwrites that knowledge, so zigpy resumes polling an attribute the device
will never answer.

The database is a record of what was read previously. A quirk is a statement about the device itself,
so where the two disagree the quirk is the better authority: the cached value is evidence the
attribute was readable under a different set of assumptions.

`zigpy/appdb.py` now leaves an attribute alone during restore when the cluster already marks it
unsupported, and `zigpy/zcl/helpers.py` carries the supporting change. Attributes with no such mark
restore exactly as before, so the cache still does its job for everything the quirk says nothing
about.

The test builds a quirked cluster declaring an attribute unsupported, seeds the DB with a stale value
for it, and asserts the mark survives the restore.

Fixes #1875


puddly:
This is a bit of a tricky case. I think calling `self.add_unsupported_attribute` within `__init__` to prevent an entity from being created accomplishes its goal but indirectly: it pollutes the (already fake) ZCL database state with runtime modifications to mimic real device communication.

This isn't a common pattern. I think the better approach would be to use `prevent_default_entity_creation`, which is the dedicated API to accomplish exactly this.

cpruijsen:
Pushed 70e5357 for the ruff-format failure.

On `prevent_default_entity_creation`: it lives in zha-device-handlers (`zhaquirks/builder/builder.py`), not zigpy, so the test here cannot use it. Agreed it is the right tool for suppressing an entity, and the PJ-1203A quirk in the issue would be better served by it.

The zigpy-level bug looks separate from entity creation. `is_attribute_unsupported()` is zigpy's own public API, and its answer flips across a restart whenever a stale row exists for that attribute. The quirk's `__init__` declaration is erased by the restore regardless of what any consumer does with the answer, and that is what this PR changes.

So the question is whether zigpy wants `add_unsupported_attribute()` at instantiation to survive a database restore. If not, and quirks should express this only through `prevent_default_entity_creation`, I am happy to close this and the issue. If yes, the test can set the mark on the cluster directly rather than overriding `__init__`.

PR #1888, opened 2026-09-17
Answer requests sent to clusters not advertised by devices

puddly:
Might address https://github.com/home-assistant/core/issues/181626#issuecomment-5719258517.

---

Zigbee devices have endpoints and every endpoint advertises its server and client clusters. For 99.99% of devices, this is accurate and messages do not arrive on endpoints or clusters that are unexpected. Zigpy assumes this is how devices work.

Unfortunately, it looks like the above Sonoff device has hit a bad combination of bugs:
1. It has a `KeepAlive` cluster, so it tries to determine connectivity to the coordinator on its own.
2. It does not _advertise_ having support for the `KeepAlive` cluster, so zigpy does not expect this packet to come in and ignores it.
3. Zigpy gained Zigbee 4.0 support back in 2026.8.0 so we advertise supporting the `KeepAlive` cluster.

The end result is the device thinks it is not connected to the coordinator and tries to re-connect every few hours.

This PR gives zigpy the ability to respond even if the device is not _supposed_ to have a cluster. Ideally, this should be fixed in the device firmware. We could also work around it in a quirk. This is a more general workaround.


Wackymax:
One additional interoperability point for this PR: ZCL revision 8 §3.18.4 requires the Keep-Alive Read Attributes **response** to be APS-encrypted. In zigpy 2.2.0, `ControllerApplication.request()` adds `TransmitOptions.APS_Encryption` for `ZigbeeDirectConfiguration`, but not for `KeepAlive`. A narrowly scoped regression test can assert the APS flag for both `expect_reply=True` and `False` on cluster `0x0025`, while retaining the existing behavior for other clusters.

This is separate from the unadvertised-client-cluster dispatch fix in this PR and from the device-specific quirk in [zha-device-handlers#5349](https://github.com/zigpy/zha-device-handlers/pull/5349). Specification: [ZCL revision 8, §3.18.4](https://zigbeealliance.org/wp-content/uploads/2021/10/07-5123-08-Zigbee-Cluster-Library.pdf).

PR #1890, opened 2026-09-19
Add public endpoint group membership readback

liudger:
## Why

A proposed Home Assistant/ZHA workflow creates or reuses a Zigbee group from selected lights. Before offering that group as an automation target, it needs to confirm membership directly with each device. Cached membership alone cannot confirm that the devices actually joined.

The existing `group_membership_scan()` updates the local cache and handles failures internally. Callers need a read-only result that distinguishes a successful read from a timeout or unsupported operation.

## Changes

- Add public `Endpoint.get_group_membership()` to request membership from the device without reading or updating the local group cache.
- Return group IDs as a `frozenset`; raise `UnsupportedCluster` for missing/rejected Groups support and propagate transport errors.
- Reuse this method in `group_membership_scan()` and add regression tests.

Kept as a draft for feedback on the API and error handling. 

## Testing

- `pytest tests/test_endpoint.py -q --tb=short`: 75 passed.
- Ruff check and format check on the three changed files.
- Unit tests use mocked device responses; no hardware validation yet.


puddly:
Thanks for the PR! Could you elaborate a bit more about this proposed workflow for ZHA?

liudger:
For example. When a user creates an automation for lights and select a few or area. The option is in the interface to create a group directly from the selection. For this we need to check if the group is created correctly or if the group already exists and re-use the group.

zigpy/endpoint.py
```diff
@@ -167,19 +167,37 @@ async def remove_from_group(self, grp_id: int) -> ZCLStatus:
...
+            res = await self.groups.get_membership(groups=[])
+        except AttributeError as err:
+            raise zigpy.exceptions.UnsupportedCluster(
+                "Endpoint does not have a Groups cluster"
+            ) from err
+
         if isinstance(res, GENERAL_COMMANDS[GeneralCommand.Default_Response].schema):
-            self.debug("Device does not support group commands: %s", res)
-            return
+            raise zigpy.exceptions.UnsupportedCluster(
```
liudger: A Default Response does not necessarily mean the cluster/command is unsupported. I reproduced `FAILURE`, `NOT_AUTHORIZED`, and even `SUCCESS` all raising `UnsupportedCluster`. This exposes an incorrect capability signal to callers: a request failure or missing membership payload is not evidence that the device lacks support. Please reserve this exception for explicit unsupported statuses, and use `InvalidResponse` (preserving the status) for other default responses. A SUCCESS default response must still fail because it contains no membership list. Add parameterized coverage for these distinctions.

zigpy/endpoint.py
```diff
@@ -167,19 +167,37 @@ async def remove_from_group(self, grp_id: int) -> ZCLStatus:
...
+        """Read this endpoint's physical Groups membership from the device.
+
+        The request is always sent to the endpoint and never uses the local
+        group cache. A successful response returns the decoded group IDs. A
+        missing or unsupported Groups cluster raises ``UnsupportedCluster``;
+        transport timeouts and other Zigbee errors are propagated unchanged.
+        """
+        try:
+            res = await self.groups.get_membership(groups=[])
+        except AttributeError as err:
```
liudger, line 191: The AttributeError handler covers the entire awaited request, not just the lookup of `self.groups`. With a Groups cluster present, an AttributeError from the request implementation is converted to `UnsupportedCluster("Endpoint does not have a Groups cluster")`. This masks a software/quirk error as a device capability limitation. Resolve the cluster inside the narrow try/except, then await `get_membership()` outside it, and test that a request-side AttributeError propagates unchanged.

liudger (commented):
Self-review of the current draft at f760e172 (not an independent maintainer approval). The small read-only API is a reasonable fit, but the error classification should be tightened before marking this ready. See the two inline findings.

Validation: endpoint + group suites passed (96 tests), scoped Ruff check/format and diff check passed. Additional external probes produced 4 failures covering the two findings and 3 passes confirming fresh reads including empty membership, unchanged local membership, transport-error propagation, and cancellation. Those read-only/freshness assertions should also be added to the committed regression tests. No hardware validation was performed, and GitHub currently reports no checks for this PR.

